PR #103, opened 2023-03-02
fix: Formatting and linting for Keras Reference docs

ayulockin:
## Description

This PR fixes the code formatting and linting. Making these changes because Keras is a top priority for growth integration push this quarter and would like to have clean documentation for whoever click on it.

Covers:

- `WandbEvalCallback`
- `WandbMetricsLogger`
- `WandbModelCheckpoint`

## Ticket

## Checklist

### DIFF
--- a/docs/ref/python/integrations/keras/wandbevalcallback.md
+++ b/docs/ref/python/integrations/keras/wandbevalcallback.md
@@ -10,10 +10,10 @@ Abstract base class to build Keras callbacks for model prediction visualization.
 
 ```python
 WandbEvalCallback(
- data_table_columns: List[str],
- pred_table_columns: List[str],
- \*args,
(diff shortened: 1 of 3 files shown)


ngrayluna:
Bad news: The API Reference guide is autogenerated from the SDK release to update the API Reference guide. To update it,   find the source code, where the docstring is defined and create a PR there (e.g. wandb/wandb). Else, this whatever we update here will get overwritten on the next SDK release.

That being said, the autogen script doesn't play well with code example spacing within a docstring, so I don't think we can fix that it easily. We shouldn't have code examples within doc strings to begin with so this is not a high priority at the moment.

What I did do:
I have a fix that I'll push to get rid of the `\*` -> `*`.

PR #124, opened 2023-03-21
Tool to fix broken docs links in the wild as they are discovered

laxels:


cvphelps:
[Related slack thread](https://weightsandbiases.slack.com/archives/C01DX8LRZEJ/p1679082906509069)

PR #137, opened 2023-04-03
adding details on setting a base image for tensorflow

bcsherma:
## Description

Adds details on setting a cuda base image in case it is needed to work with tensorflow.

## Ticket

## Checklist

### DIFF
--- a/docs/guides/launch/getting-started.md
+++ b/docs/guides/launch/getting-started.md
@@ -194,7 +194,7 @@ Add the following resource configuration in order to use GPUs in jobs submitted 
 
 The `gpus` key of the resource configuration is used to pass values to the `--gpus` argument of `docker run`. This argument can be used to control which GPUs will be used for by a launch agent when it picks up runs from this queue. For more information, see the relevant [NVIDIA documentation](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/user-guide.html#gpu-enumeration).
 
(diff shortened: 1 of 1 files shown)


docs/guides/launch/getting-started.md
```diff
@@ -193,6 +193,19 @@ Add the following resource configuration in order to use GPUs in jobs submitted
 ```
 
 The `gpus` key of the resource configuration is used to pass values to the `--gpus` argument of `docker run`. This argument can be used to control which GPUs will be used for by a launch agent when it picks up runs from this queue. For more information, see the relevant [NVIDIA documentation](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/user-guide.html#gpu-enumeration).
+
+For jobs that use tensorflow on GPU, you may also need to specify a custom base image for the container build that the agent will perform in order for your runs to properly utilize GPUs. This can be done by adding an image tag under the `builder.cuda.base_image` key to the resource configuration. For example:
```
KyleGoyette, line 197: This is an okay temporary fix maybe, but there's some longer term work here that's needed. Indicating that we have this `builder` key thats beside the other resource args keys...

KyleGoyette (approved):
This is okay as an incremental improvement. But we need to break these things out more and define the builder key and other keys.

docs/guides/launch/getting-started.md
bcsherma: I agree 100%, my intent is to get this into the docs as fast as possible in case people run into the same problem as Ayush

PR #151, opened 2023-04-05
changing prerequisites to agent docs

bcsherma:
## Description

The prerequisites section of the docs has a lot of material that's also in the quickstart, but crucially it also has the agent docs in it. I'm removing some of the redundant stuff and adding more robust documentation of how the agent works. Docs on how to deploy the agent in kubernetes will still live in the forthcoming guide to launch on k8s, but this is intended to outline supported agent behaviors and how to use them. It's rough ATM but I want to get the content out there and iterate from there.

## Ticket

## Checklist

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -151,7 +151,7 @@ const sidebars = {
       },
       items: [
         'guides/launch/getting-started',
-        'guides/launch/prerequisites',
+        'guides/launch/agent',
         'guides/launch/create-job',
         'guides/launch/add-jobs-to-queue',
         'guides/launch/launch-jobs',
--- a/docs/guides/launch/add-jobs-to-queue.md
+++ b/docs/guides/launch/add-jobs-to-queue.md
@@ -123,5 +123,3 @@ Select the **Agents** tab to view active and inactive agents assigned to the que
 - **Host:** The machine the agent is polling on.
 
 ![](/images/launch/queues_all_agents.png)
-
-<!-- For information on how to add an agent see [LINK to Prerequisites]. -->
--- a/docs/guides/launch/agent.md
+++ b/docs/guides/launch/agent.md
@@ -33,117 +33,135 @@ The launch agent will execute the run in the container image that was built in t
 For example, if the job was in a Docker queue, the agent will execute the run locally with the `docker run` command. If the job was in a Kubernetes queue, the agent will execute the run on a k8s cluster as a [k8s Job](https://kubernetes.io/docs/concepts/workloads/controllers/job/) via the k8s API.
 
 
+## Agent configuration
 
-<!-- 
-### Activate a launch agent
-A launch agent listens the queue and executes jobs added to the queue. If certain criteria are met, the launch agent builds an image for the environment and sends the job to the desired compute runner. 
-
-The following tabs provide information on how to activate a launch agent with basic, standard control settings and with more control options with a config file.
-
-
-<Tabs
-  defaultValue="basic"
-  values={[
-    {label: 'Basic', value: 'basic'},
-    {label: 'EKS', value: 'eks'},
-  ]}>
-  <TabItem value="basic">
-
-Activate a launch agent in your terminal with the `wandb launch-agent` command:
-
-```bash
-wandb launch-agent -q <queue-name>
-```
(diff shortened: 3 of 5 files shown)


sidebars.js
```diff
@@ -151,7 +151,7 @@ const sidebars = {
       },
       items: [
         'guides/launch/getting-started',
-        'guides/launch/prerequisites',
+        'guides/launch/agent',
```
KyleGoyette, line 154: Oh nice, we should update the doc call outs in the launch agent code to this. Not a task for you, but a good page IMO

ngrayluna (changes requested):
Not sure we should remove the prerequisits. We can't assume folks will look at the quickstart before trying it out. I agree that some info is redundant, but this is one of those cases were it makes sense to repeat.

But people are more likely to look at the prereqs before moving on.

 Dig the new agent page.

bcsherma:
I think "getting started" is as reliable as "prerequisites" as something that people will look at before assuming they are set up to use a product. I would prefer that we eliminate the prequisites page and replace it with the agent docs page

KyleGoyette:
Our current prerequisites page feels weird since it is actually explaining the requirements for each step of using launch:
- make a job
- make a queue
- start an agent

In reality our prerequisites depend on the way users intend on using launch. Generally they are:
- activate the feature flag
- make a queue (can be made while launching a job in the UI)
- for docker container install and start docker on your machine
- start an agent

It seems to me we could have a separate page for:
- Quickstart (minimal flow of make a job, and use the quickstart queue maker)
- Creating a job
- Creating a queue (docker and k8s and other types)
- Starting an agent (K8s and CLI)
- Enqueueing a job (CLI and app)
- Sweeps on Launch

Within the quickstart we could have links to each of these other pages indicating how to do more advanced features, rather than having the advanced features interspersed throughout the prereqs. The prereqs for the quickstart would specifically be the things required to do the quickstart (feature flag, docker).

So once we add the agent docs page we can add a "Launch Queues" page and then each of the concepts is split functionally across the pages. Then as users go through the quick start we can talk about advanced set up on each of the pages. I think this will end up with a much shorter quickstart, and will follow more closely to our other docs structures. What do y'all think

tristan-spaulding2:
I agree @KyleGoyette 👍

ngrayluna:
@KyleGoyette Makes sense. QQ: In the v0 docs we avoided having a 
* "Creating a queue (docker and k8s and other types)" and 
* "Starting an agent (K8s and CLI)"  pages 
because we wanted to keep this info "hidden". 

Re: It was suggested by the PM at the time that we wanted to hide those pages because:
1. We wanted MLOps teams chats w/ us, make sure they paid for the service.  Hence, that info was oddly, and incompletely, bunched up into the "Prereqs" page. 
2.  We wanted to target the ML Practitioner. So, in that sense, they "shouldn't" see that info.  TL;DR the v0 docs are aimed for the ML Practitioner. 

That's all to say that, it's completely fine to change this. But we should decided WHO we are writing for, and from there decided their knowledge path.

bcsherma:
For now I'm not that concerned about targeting specific personas, I just want to make sure we have accurate and robust documentation. @KyleGoyette I agree with your proposal for a restructure. I think we also should have end to end guides for launch on k8s, launch on docker, launch on sagemaker, etc. which can of course refer to the more specific pages you propose, some of which we already have. It will be something like

- quickstart
- create a job
- create a queue
- run an agent
- launch a job
- launch on docker
- launch on k8s
- launch on sagemaker
- launch on vertex
- sweeps on launch

I think the "getting-started" we have now becomes the "launch on docker" page and a pruned down version that refers to the "Starter queue" can be our quickstart.

In the meantime I think we should review and merge this PR. The prerequisites page should go and we need an agent documentation page. Any objections?

KyleGoyette:
Makes sense @ngrayluna, I think we're adjusting and adding more info publicly. I think so long as we all agree with the new strategy, I'm on board with this change

KyleGoyette (approved):

bcsherma:
I need to fix some broken links and then will murg

ngrayluna:
> Makes sense @ngrayluna, I think we're adjusting and adding more info publicly. I think so long as we all agree with the new strategy, I'm on board with this change

Right on. So long everyone is on board and it doesn't come back that "Oh, hey. We said we didn't want this info public". And if it does, ya'll take ownership of that.

PR #179, opened 2023-04-19
Add JS docs

scottire:
Add JS docs


docs/ref/js/README.md
```diff
@@ -0,0 +1,90 @@
...
+    wandb.log({acc: 0.9, loss: 0.1});
+    wandb.log({acc: 0.91, loss: 0.09});
+    await wandb.finish();
+}
+
+await track()
+```
+
+:::caution
+We spawn a seperate MessageChannel to process all api calls async.  This will cause your script to hang if you don't call `await wandb.finish()`.
```
jamie-rasmussen: seperate => separate

docs/ref/js/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+```javascript
+const wandb = require('@wandb/sdk').default;
+```
+
+We're currently missing a lot of the functionality found in our Python SDK, but basic logging functionality is available. We'll be adding additional features like [Tables](https://docs.wandb.ai/guides/data-vis?utm_source=github&utm_medium=code&utm_campaign=wandb&utm_content=readme) soon.
+
+### Authentication and Settings
+
+In node environments we look for `process.env.WANDB_API_KEY` and prompt for it's input if we have a TTY.  In non-node environments we look for `sessionStorage.getItem("WANDB_API_KEY")`.  Addtional settings can be [found here](https://github.com/wandb/wandb-js/blob/main/src/sdk/lib/config.ts).
```
jamie-rasmussen: it's => its, Addtional => Additional

PR #185, opened 2023-04-20
[For London] New openai integration docs

morganmcg1:
Adds docs for the new `autolog` integrations for the OpenAI API 

No colab for now, the code examples here will work if the user copies and runs them

### Screenshots:

Screenshot 2023-04-20 at 01 42 20

Screenshot 2023-04-20 at 01 42 28

Screenshot 2023-04-20 at 01 42 34


docs/guides/integrations/other/openai.md
```diff
@@ -12,7 +12,62 @@ import TabItem from '@theme/TabItem';
...
+
+### 1. Import autolog and initialise it
+First, import `autolog` from `wandb.integration.openai` and initialise it.  
+
+```python
+import os
+import openai
+from wandb.integration.openai import autolog
+
+autolog({"project":"gpt5"})
```
kptkin, line 32:
```suggestion
autolog(init={"project":"gpt5"})
```

kptkin: we modified this api a bit to allow for future extensions, so we namespace the init args with this argument. 
I know it is a bit different from your original design, but i hope it's not too much of a difference

PR #198, opened 2023-04-27
Expand on bare-metal docs relating to infrastructure requirements

zacharyblasczyk:
## Description

## Ticket

## Checklist

- [x] Build (`yarn docusaurus build`) was run locally and successfully without errors or warnings.

### DIFF
--- a/docs/guides/hosting/hosting-options/self-managed.md
+++ b/docs/guides/hosting/hosting-options/self-managed.md
@@ -73,7 +73,7 @@ Additionally, we recommend the following [parameter overrides](../how-to-guides/
 
 ### Object Storage
 
-W&B has no hardware requirements for the on premises storage backend as long as it supports an S3 API interface, Signed URL’s, and a CORS policy. We recommend specing the storage array to the current needs of your ML practitioners and to capacity plan on a regular cadence.
+W&B is compatible with an object storage that supports S3 API interface, Signed URLs and CORS. We recommend specing the storage array to the current needs of your practitioners and to capacity plan on a regular cadence.
 
 The [following script](https://gist.github.com/vanpelt/2e018f7313dabf7cca15ad66c2dd9c5b) can be used to validate that the Object Store supports Signed URL’s.
 
--- a/docs/guides/hosting/hosting-options/self-managed.md
+++ b/docs/guides/hosting/hosting-options/self-managed.md
@@ -54,7 +54,7 @@ It is important to keep in mind that these specifications are minimum requiremen
 
 ### Database Server
 
-W&B recommends a [MySQL 8](../how-to-guides/bare-metal.md#mysql-80) database as a metadata store of all project, report, artifact, and run data. The shape of the ML practitioners parameters and metadata will greatly affect the performance of the database. The database is typically incrementally written to as SMLE’s track their training runs and is more read heavy when queries are executed in reports and dashboard.
+W&B recommends a [MySQL 8](../how-to-guides/bare-metal.md#mysql-80) database as a metadata store. The shape of the ML practitioners parameters and metadata will greatly affect the performance of the database. The database is typically incrementally written to as practitioners track their training runs and is more read heavy when queries are executed in reports and dashboard.
 
 To ensure optimal performance we recommend deploying the W&B database on to a server with the following starting specs:
 
--- a/docs/guides/hosting/hosting-options/self-managed.md
+++ b/docs/guides/hosting/hosting-options/self-managed.md
@@ -31,4 +31,4 @@ W&B can provide recommendations for compatible database engines, object stores a
 
 ### Contact
 
-If you have questions about planning an onpremises installation of W&B please take a look at our [On Prem / Baremetal](../how-to-guides/bare-metal.md) docs and reach out to us at support@wandb.com.
+See the [On Prem / Baremetal](../how-to-guides/bare-metal.md) documentation if you have questions about planning an on premises installation of W&B and reach out to W&B Supported at support@wandb.com.
--- a/docs/guides/hosting/hosting-options/self-managed.md
+++ b/docs/guides/hosting/hosting-options/self-managed.md
@@ -29,64 +29,6 @@ This is a fully self-hosted solution where W&B server is running in a scalable d
 
 W&B can provide recommendations for compatible database engines, object stores and have an experienced team to help with the installation process. The complexity of administrating a database, creating and maintaining a distributed object storage system adds additional overhead to the customer's IT/DevOps/MLOps teams. When possible, W&B recommends using W&B managed cloud solutions for better user experience.
 
-### Infrastructure Guidelines
-
-It is highly recommended to deploy W&B into a kubernetes cluster to fully take advantage of all the features of the product and to utilize the standardized [helm interface](https://github.com/wandb/helm-charts) that we provide.
-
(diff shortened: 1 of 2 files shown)


docs/guides/hosting/hosting-options/self-managed.md
```diff
@@ -28,3 +28,79 @@ This is a fully self-hosted solution where W&B server is running in a scalable d
 - a redis cache (optional)
 
 W&B can provide recommendations for compatible database engines, object stores and have an experienced team to help with the installation process. The complexity of administrating a database, creating and maintaining a distributed object storage system adds additional overhead to the customer's IT/DevOps/MLOps teams. When possible, W&B recommends using W&B managed cloud solutions for better user experience.
+
+### Infrastructure Guidelines
+
+It is highly recommended to deploy W&B into a kubernetes cluster to fully take advantage of all the features of the product and to utilize the standardized [helm interface](https://github.com/wandb/helm-charts) that we provide.
+
+If kubernetes is unavailable, W&B can be installed onto a bare-metal server and configured manually, however, future features may be broken out into k8s native or custom resource definitions and may not be backported into the standalone Docker Container.
```
zacharyblasczyk: future features may be broken out into k8s native or custom resource definitions and may not be backported into the standalone Docker Container.  

I was thinking about Prometheus as an example.

I don't think there has been a shift to a docker composed based deployment.

docs/guides/hosting/hosting-options/self-managed.md
```diff
@@ -28,3 +28,80 @@ This is a fully self-hosted solution where W&B server is running in a scalable d
...
+| Memory (GiB)               | 8                                 |
+
+This ensures that W&B has sufficient disk space to process W&B server application data and store temporary logs before they are externalized. It also ensures fast and reliable data transfer, the necessary processing power and memory for smooth operation, and that W&B will not be affected by any noisy neighbors. 
+
+It is important to keep in mind that these specifications are minimum requirements, and actual resource needs may vary depending on the specific usage and workload of the W&B application. Monitoring the resource usage and performance of the application is critical to ensure that it operates optimally and to make adjustments as necessary.
+
+
+### Database Server
+
+W&B recommends a [MySQL 8](../how-to-guides/bare-metal.md#mysql-80) database as a metadata store of all project, report, artifact, and run data. The shape of the ML practitioners parameters and metadata will greatly affect the performance of the database. The database is typically incrementally written to as SMLE’s track their training runs and is more read heavy when queries are executed in reports and dashboard.
```
venky-wandb: - Just saying `database as a metadata store.` would probably suffice here. Can remove this part `of all project, report, artifact, and run data`
- SMLE is an internal name we have for our success MLEs. We should change this to practitioners in all places for consistency.

venky-wandb:
@wandb-zacharyblasczyk This entire page should be under `how-to guides > bare-metal` section. The current location is just an intro page for what bare-metal. Actual configuration details should be part of the how-to guide.

docs/guides/hosting/hosting-options/self-managed.md
```diff
@@ -28,3 +28,80 @@ This is a fully self-hosted solution where W&B server is running in a scalable d
...
+| Memory (GiB)               | 32                                |
+
+
+Again, we recommend monitoring the resource usage and performance of the database to ensure that it operates optimally and to make adjustments as necessary.
+
+Additionally, we recommend the following [parameter overrides](../how-to-guides/bare-metal.md#mysql-80) to tune the DB for MySQL 8.
+
+### Object Storage
+
+W&B has no hardware requirements for the on premises storage backend as long as it supports an S3 API interface, Signed URL’s, and a CORS policy. We recommend specing the storage array to the current needs of your ML practitioners and to capacity plan on a regular cadence.
```
venky-wandb: I think we could rephrase this to say, `W&B is compatible with an object storage that supports S3 API interface, Signed URLs and CORS.` 

No hardware requirements part might not necessarily be true in some cases so just tweaking it to say what we need.

zacharyblasczyk:
@venky-wandb, I kind of agree and disagree.

I started with that page but found that it really doesn't fit. 

This page gives a high level narrative of _what_ is needed for bare metal and brief details about server requirements and object storage options/reqs, the other page gives a detailed explanation of _how_ to configure those pieces for bare metal regardless of what they are.

Let me know what you think.

zacharyblasczyk:
@venky-wandb, I moved the CORS policy out as it was more of a detailed config item.

docs/guides/hosting/hosting-options/self-managed.md
```diff
@@ -28,3 +28,65 @@ This is a fully self-hosted solution where W&B server is running in a scalable d
 - a redis cache (optional)
 
 W&B can provide recommendations for compatible database engines, object stores and have an experienced team to help with the installation process. The complexity of administrating a database, creating and maintaining a distributed object storage system adds additional overhead to the customer's IT/DevOps/MLOps teams. When possible, W&B recommends using W&B managed cloud solutions for better user experience.
+
+### Infrastructure Guidelines
+
+It is highly recommended to deploy W&B into a kubernetes cluster to fully take advantage of all the features of the product and to utilize the standardized [helm interface](https://github.com/wandb/helm-charts) that we provide.
```
ngrayluna: QQ: would customers otherwise not have access to all W&B features and/or the helm interface?

docs/guides/hosting/hosting-options/self-managed.md
ngrayluna: How certain are we that future features might not be backported into the standalone Docker container?

If fairly positive, maybe we can say something like "W&B features will no longer be backported into standalone Docker Container"... (or something like that).

venky-wandb:
> @venky-wandb, I kind of agree and disagree.
> 
> I started with that page but found that it really doesn't fit.
> 
> This page gives a high level narrative of _what_ is needed for bare metal and brief details about server requirements and object storage options/reqs, the other page gives a detailed explanation of _how_ to configure those pieces for bare metal regardless of what they are.
> 
> Let me know what you think.

I would like to keep the Hosting Options view clean and just have a 30,000 foot view of the deployment type. Specifics of the deployment type should go into the configuration section. `Infrastructure Guidelines` sounds like it'd be more closer to the Guides section rather than the hosting options section. Also, hosting options is the page we expect a lot of user personas to visit including decision makers, ml engineers etc. - that's one of the main reasons to keep it generic and not go into too much depth and overwhelm with lots of info. Guides is a section which we expect MLOps Admins would visit who'd be able to consume the infra info meaningfully. Hope that makes sense! :)

docs/guides/hosting/hosting-options/self-managed.md
venky-wandb: To avoid any confusion, I'd suggest to make the following changes:
We will split this into two sentences. 

First,
```
If kubernetes is unavailable, W&B can be installed onto a bare-metal server and configured manually.
```

Second,
```
Note that, future W&B features may require leveraging k8s native capabilities and hence would require a K8s first environment.
```

ngrayluna:
Great response @venky-wandb and @wandb-zacharyblasczyk. Funny enough,  I agree with both of ya'll. 

It makes sense to keep the hosting options clean. 
But it also makes sense that the infra guidelines don't really belong in the *how* portion of our docs.

It feels like these infra guidelines should be their own thing page? Do we have infra guidelines for on-prem private cloud as well?

ngrayluna (approved):

venky-wandb (approved):
Looks solid, thank you Zach!! 🔥

PR #207, opened 2023-05-02
docs(artifacts): document Azure storage handler

szymon-piechowicz-wandb:
## Description

Documented new functionality added in https://github.com/wandb/wandb/pull/5317.

## Ticket

WB-13498

## Checklist

### DIFF
--- a/docs/guides/artifacts/track-external-files.md
+++ b/docs/guides/artifacts/track-external-files.md
@@ -66,13 +66,13 @@ By default, W&B imposes a 10,000 object limit when adding an object prefix. You 
 
 Our new reference artifact `mnist:latest` looks and behaves similarly to a regular artifact. The only difference is that the artifact only consists of metadata about the S3/GCS/Azure object such as its ETag, size, and version ID (if object versioning is enabled on the bucket).
 
-Weights & Biases will use the default mechanism of looking for credentials for a given cloud provider. Consult the documentation from your cloud provider:
+W&B will use the default mechanism to look for credentials based on the cloud provider you use. Read the documentation from your cloud provider to learn more about the credentials used:
 
 | Cloud provider | Credentials Documentation |
 | -------------- | ------------------------- |
-| AWS            | [Link](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials) |
-| GCP            | [Link](https://cloud.google.com/docs/authentication/provide-credentials-adc) |
-| Azure          | [Link](https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python) |
+| AWS            | [Boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials) |
+| GCP            | [Google Cloud documentation](https://cloud.google.com/docs/authentication/provide-credentials-adc) |
+| Azure          | [Azure documentation](https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python) |
 
 Interact with this artifact similarly to a normal artifact. In the App UI, you can look through the contents of the reference artifact using the file browser, explore the full dependency graph, and scan through the versioned history of your artifact.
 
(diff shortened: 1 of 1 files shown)


docs/guides/artifacts/track-external-files.md
```diff
@@ -74,6 +74,8 @@ Weights & Biases will attempt to use the corresponding credential files or envir
 | 2 - Shared credentials file | `~/.aws/credentials`                                                                                                | `application_default_credentials.json` in `~/.config/gcloud/` |
 | 3 - Config file             | `~/.aws.config`                                                                                                     | N/A                                                           |
 
+For Azure Blob Storage, [default Azure credential](https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python) will be used.
```
ngrayluna: QQ: Does the schema for Azure differ from Amazon S3 or GCP?
e.g. in our docs we have this table listing env var, etc. 
Screen Shot 2023-05-02 at 1 25 29 PM

szymon-piechowicz-wandb: Yes, it's a little more complicated for Azure. If you click through the link, you can see there are 6 options, each of which has sub-options. For someone familiar with Azure, I think saying that we use `DefaultAzureCredential` is exactly what they need to know.

noaleetz: @szymon-piechowicz-wandb - Does it make sense to add a col for Azure and define that sentence for #2- shared credential files? Would #1 an #3 be N/A?

noaleetz: @szymon-piechowicz-wandb - it looks like you linked the screenshot to the existing table for GCP/S3 (not an Azure link).

docs/guides/artifacts/track-external-files.md
```diff
@@ -30,11 +30,11 @@ For an example of tracking reference files in GCP, see the [Guide to Tracking Ar
 
 The following describes how to construct reference artifacts and how to best incorporate them into your workflows.
```
noaleetz, line 31: @ngrayluna Hey Noah! One thing I was thinking - would it make sense to add a box that specifies that Azure has been newly added? Similar to how we have 'warning' blocks, would there be a similar one to add for a new feature to call attention that Reference Artifacts in Azure are now available? We will surface this elsewhere as well but was wondering if there's a place in this doc to call this out as well..

ngrayluna: Sadly, it is considered a 'bad practice' to include text, blocks etc. that is 'dated' in tech docs .  :/

docs/guides/artifacts/track-external-files.md
ngrayluna: We can also create a new table and summarize what customers need to know?

docs/guides/artifacts/track-external-files.md
noaleetz: Gotcha makes sense, I had been thinking of something like this but it sounds like we are working to remove similar blocks / deprecate going forward. 

Screenshot 2023-05-02 at 3 50 23 PM

docs/guides/artifacts/track-external-files.md
noaleetz, line 1: @szymon-piechowicz-wandb - should we specify Azure in [this](https://github.com/wandb/docodile/blob/643c5f7d073059de9d060a87d5b5d10359d67d2b/docs/guides/artifacts/track-external-files.md?plain=1#L98) line where we encourage users to enable obj versioning in s3/gcs:

> W&B recommends that you enable 'Object Versioning' on your Amazon S3 or GCS buckets if you overwrite files as part of your workflow. With versioning enabled on your buckets, artifacts with references to files that have been overwritten will still be intact because the older object versions are retained.

i,e "enable 'Object Versioning' on your Amazon S3 / GCS buckets or Azure blob" ([here](https://learn.microsoft.com/en-us/azure/storage/blobs/versioning-enable?tabs=portal) are Azure instructions)

docs/guides/artifacts/track-external-files.md
szymon-piechowicz-wandb: Btw, the instructions for AWS and GCP are oversimplified as well. See the official docs for AWS [here](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials). For example, my former company used #8 which we don't mention here.

szymon-piechowicz-wandb: Updated based on discussion on Zoom.

docs/guides/artifacts/track-external-files.md
szymon-piechowicz-wandb: done

ngrayluna (approved):

PR #282, opened 2023-06-12
Added new env var

ngrayluna:
## Description

## Ticket

Does this PR fix an existing issue? If yes, provide a link to the ticket here: https://wandb.atlassian.net/browse/DOCS-719

## Checklist


tristan-spaulding2:
Sounds like this ended up getting named GORILLA_DATA_RETENTION_PERIOD and it is in hours.  Default is 168.  Sound right @ramit-wandb ?

ramit-wandb:
> Sounds like this ended up getting named GORILLA_DATA_RETENTION_PERIOD and it is in hours. Default is 168. Sound right @ramit-wandb ?

Yes, that's correct.

ngrayluna:
Duplicate: https://github.com/wandb/docodile/pull/317

docs/guides/hosting/env-vars.md
```diff
@@ -27,7 +27,7 @@ In addition to configuring instance level settings via the System Settings admin
 | REDIS                            | Can be used to setup an external REDIS instance with W&B.                                                                                                                                |
 | LOGGING_ENABLED                  | When set to true, access logs are streamed to stdout. You can also mount a sidecar container and tail `/var/log/gorilla.log` without setting this variable.                              |
 | GORILLA_ALLOW_USER_TEAM_CREATION | When set to true, allows non-admin users to create a new team. False by default.                                                                                                         |
-
+| GORILLA_DATA_RETENTION_PERIOD   | The number of hours before the data from deleted runs will be permanently deleted and unrecoverable. |
```
ramit-wandb: We would want to mention the format in which this variable would need to be set for users to know how to invoke this method. We need to specify the number of hours with an `h` suffixed to the end of it. So, something like:

```
export GORILLA_DATA_RETENTION_PERIOD="168h"
```

is valid, but

```
export GORILLA_DATA_RETENTION_PERIOD="168"
```
is not.

docs/guides/hosting/env-vars.md
```diff
@@ -27,7 +27,7 @@ In addition to configuring instance level settings via the System Settings admin
...
```
ramit-wandb, line 37: This feature starts up a cron job that goes ahead and starts clearing up deleted and unused data, and checks for it every minute. So we should also mention that users should be very cautious when they use this environment variable since their data will start getting cleaned up effective immediately. 

We probably also want to recommend taking a bucket and database backup before enabling this flag.

ramit-wandb (changes requested):

docs/guides/hosting/env-vars.md
ngrayluna: @ramit-wandb QQ: This one minute cadence of clearing and deleting data happens after the time specified? e.g. If I specify 2hours, it'll start clearing out data AFTER two hours?

PR #325, opened 2023-06-29
Add LlamaIndex integration

morganmcg1:
Adds a page for LlamaIndex for improved discoverability. Will likely move the LlamaIndex example from the Prompts quickstart to this page in the future.

**Testing**
- Viewed locally, sidebar showed as expected in en & ja
- `yarn docusaurus build`
  - en built successfully
  - ja failed


morganmcg1:
@ngrayluna not sure why the tests are failing here, as they were passing last night on a separate PR...

Maybe its because I'm adding a new page to the sidebar? The link it's complaining about (../prompts/quickstart.md) was copy-pasted from the LangChain integrations page, so I'm not sure why it would be problematic now...

ngrayluna:
We have this weird bug where we need to manualy copy and paste any new files to the translated docs IF the page we add refers to a doc outside of that directory.  In this case, this 

```
(../prompts/quickstart.md)
```

On that note. There's no content here. Customers don't like seeing pages that exist only to redirect them. Proposal:
Can we move the current LlamndaIndex content in the prompts/quickstart.md  to this page instead in integrations?

morganmcg1:
Thanks Noah, its the same as the LangChain placeholder we have at the moment. The goal is to increase visibility & discoverability of all our Prompts offersing if users are browsing the docs. I think breaking some integrations-like pages out of Prompts makes sense in the long-run, but right now I think we need 1 place where we can send users that will point to all the wonderful LLM offerings and integrations that we have.

ngrayluna:
Closing this PR. There's no content yet.

PR #333, opened 2023-07-06
Auto translate Action

tcapelle:
This PR adds a workflow that will call GPT4 to auto-translate to Japanese modified `.md` files.

## How does it work?
It's a manual action that you can call on branch, and it will check what files have been changed inside /docs and translate them and automatically commit them. Once the action is run, do a `git pull` to recover the corresponding Japanese files.

## Don't run this action on Huge PRs...

This action requires GITHUB_TOKEN to have write access and have a secret the OPENAI_API_KEY


johndmulhausen:
We have a new strategy where we update once per quarter; closing this

PR #351, opened 2023-07-20
feat: Add guide on Dagster partitions

chrishiste:
## Description

Adds a guide on the new features included in the Dagster integration.

## Checklist


docs/guides/integrations/dagster.md
```diff
@@ -426,6 +426,42 @@ from dagster_wandb import WandbArtifactConfiguration
...
+            "type": "dataset",
+        }
+    },
+)
+def create_my_daily_partitioned_asset(context):
+    partition_key = context.asset_partition_key_for_output()
+    context.log.info(f"Creating partitioned asset for {partition_key}")
+    return random.randint(0, 100)
+```
+This code will produce one W&B Artifact for each partition. They can be located under the Artifact name, which will be appended with the partition key, for example `my_daily_partitioned_asset.2023-01-01`. Assets that are partitioned across multiple dimensions will have each partition divided by a dot e.g. `my_asset.car.blue`.
```
ngrayluna: QQs: 
1. By "stored" do we mean w/in a W&B project?
2. We say "may" is it not always the case? Or can we rephrase this to: 
" This code will produce one W&B Artifact for each partition. It is located under the Artifact name, which will is appended with the partition key, for example..."

chrishiste: I can't seem to find any use of "stored" nor "may", which line were you looking at?

ngrayluna: Lol. Oops. I changed the wording and pushed it. I rewrote it as: 

"This code will produce one W&B Artifact for each partition. They can be located under the Artifact name, which will be appended with the partition key, for example `my_daily_partitioned_asset.2023-01-01`."

In this new version, I am unsure what this means "They can be located under the Artifact name..."

chrishiste: What I mean by "They can be located under the Artifact name..." is that you will find multiple Artifact with the same name in the Artifacts UI. 

Let's see we have an artifact called `robot` with 3 partitions: `blue`, `red` and `green` you will find 3 Artifacts:
`robot.blue`
`robot.red`
`robot.green`

chrishiste: Is that less confusing? Maybe we should say Artifact UI?

ngrayluna: Ahh That makes sense. Thank you. 

Yes, let's add "W&B App UI" or something like that.

chrishiste: Should be all good now

ngrayluna (approved):

PR #387, opened 2023-08-07
Rewrote resume chapter

ngrayluna:
## Description

## Ticket

https://wandb.atlassian.net/browse/DOCS-487?atlOrigin=eyJpIjoiMGU1NjJkN2UwYmZmNDRjNGE2YTFiNWY2NWQxMDc4YTciLCJwIjoiaiJ9

## Checklist


docs/guides/runs/resuming.md
```diff
@@ -11,116 +11,43 @@ import TabItem from '@theme/TabItem';
   <title>Resume W&B Runs</title>
 </head>
 
-You can have W&B automatically resume runs by passing `resume=True` to `wandb.init()`. If your process doesn't exit successfully, the next time you run it W&B will start logging from the last step.
+Use the `resume` parameter to resume runs that have crashed or or stopped. If the `resume` parameter is left unspecified W&B will, by default, create a new run and overwrite the data of the crashed run if you start a new run that has the same run ID as the run that crashed.
```
kptkin:
```suggestion
Use the `resume` parameter to resume runs that have crashed or stopped. If the `resume` parameter is left unspecified W&B will, by default, create a new run and overwrite the data of the crashed run if you start a new run that has the same run ID as the run that crashed.
```

kptkin: What does `stopped` means in this context? we usually have crashed and finished.

kptkin:
```
If the `resume` parameter is left unspecified W&B will, by default, create a new run and overwrite the data of the crashed run if you start a new run that has the same run ID as the run that crashed.
```
I think this sentence is a bit confusing. I think we should start by addressing the case where the user provided an ID, but not the `resume` argument and then describe the consequences (also we probably should discourage this kinda of usage as it will lead to un-expected behavior).

kptkin: Also I think it might be confusing jumping into the above details without saying that these parameters are specified in `wandb.init`, also it assumes that the users know what `ID` field means in `wandb.init`

docs/guides/runs/resuming.md
```diff
@@ -11,116 +11,43 @@ import TabItem from '@theme/TabItem';
...
-    wandb.save(CHECKPOINT_PATH) # saves checkpoint to wandb
-    epoch += 1
+run = wandb.init(resume=True)
 ```
-  </TabItem>
-</Tabs>
 
-### Resume Guidance
+:::note
+This only works if you are running your script in the same directory as the one that failed as the file is stored at: `wandb/wandb-resume.json`.
```
kptkin: this note is skipping a lot of context, what does this file mean? probably worth describing what we do, or on the hand if we don't want to specify all the details of the implementation we should remove reference to a file name that is not described where it came from.

kptkin: Also ideally we want to get away from `True` and instead use `resume='auto'` as we want to use literals, i.e `resume` should be one of `'auto'`, `'allow'`, `'must'`, `'never'`

kptkin: Also worth mentioning all the corner cases of this mode... we can walk through them if you need more context

docs/guides/runs/resuming.md
```diff
@@ -11,116 +11,43 @@ import TabItem from '@theme/TabItem';
...
+:::
+
+### Resume an identical run with the same run ID
+Determine how W&B resumes identical runs based on the run ID. Define the run ID either with the `id` parameter when you initialize a run with `wandb.intit` or set the`WANDB_RUN_ID` environment variable. Ensure that you provide the ID of the run, the project the ID belongs to, along with the project the run belongs to.
```
kptkin:
```suggestion
Determine how W&B resumes identical runs based on the run ID. Define the run ID either with the `id` parameter when you initialize a run with `wandb.init` or set the`WANDB_RUN_ID` environment variable. Ensure that you provide the ID of the run, the project the ID belongs to, along with the project the run belongs to.
```

kptkin:
```
the project the ID belongs to, along with the project the run belongs to.
``` 
this seems duplicated

kptkin: Also maybe we shouldn't encourage using `WANDB_RUN_ID` as there are a lot of gotcha when using env variables, these are only valid before we call the first init after that we ignore any changes to env variables.

docs/guides/runs/resuming.md
```diff
@@ -11,116 +11,43 @@ import TabItem from '@theme/TabItem';
...
 
-There are different ways in which W&B can be used to resume runs as detailed below:
+Based on your use case, specify one of the following:
+- `"allow"`:  W&B automatically resumes the run with the run ID specified. Otherwise, W&B will start a new run.
```
kptkin: This mode means that we check if the run already exists, if it does we use that run otherwise we initialize a new run with the specified id. i.e. the run can exist for us to resume

docs/guides/runs/resuming.md
```diff
@@ -11,116 +11,43 @@ import TabItem from '@theme/TabItem';
...
+- `"never"`:  W&B will crash.
+- `"must"`:   W&B automatically resumes the run with the ID specified. Otherwise, W&B will crash.
```
kptkin: This mode is similar to allow as we check if the run exists, if it does we use it otherwise we fail the run. I.e. the run must exists for us to resume.

docs/guides/runs/resuming.md
```diff
@@ -11,116 +11,43 @@ import TabItem from '@theme/TabItem';
...
+This only works if you are running your script in the same directory as the one that failed as the file is stored at: `wandb/wandb-resume.json`.
+:::
+
+### Resume an identical run with the same run ID
+Determine how W&B resumes identical runs based on the run ID. Define the run ID either with the `id` parameter when you initialize a run with `wandb.intit` or set the`WANDB_RUN_ID` environment variable. Ensure that you provide the ID of the run, the project the ID belongs to, along with the project the run belongs to.
 
-There are different ways in which W&B can be used to resume runs as detailed below:
+Based on your use case, specify one of the following:
+- `"allow"`:  W&B automatically resumes the run with the run ID specified. Otherwise, W&B will start a new run.
+- `"never"`:  W&B will crash.
```
kptkin: This mode means that we check if the run exists and if it does we fail, otherwise we start a new run with the specified run id.

docs/guides/runs/resuming.md
```diff
@@ -142,20 +69,23 @@ If multiple processes use the same `run_id` concurrently unexpected results will
...
+<!-- END -->
 
-### Preemptible Sweeps
 
-If you are running a sweep agent in a compute environment that is subject to preemption (e.g., a SLURM job in a preemptible queue, an EC2 spot instance, or a Google Cloud preemptible VM), you can automatically requeue your interrupted sweep runs, ensuring they will be retried until they run to completion.
 
-When you learn your current run is about to be preempted, call
 
-```
+## Resume preemptible Sweeps runs
```
kptkin, line 129: should we point to sweeps docs?

docs/guides/runs/resuming.md
```diff
@@ -142,20 +69,23 @@ If multiple processes use the same `run_id` concurrently unexpected results will
...
+If a run that is marked preempting exits with status code 0, W&B will consider the run to have terminated successfully and it will not be requeued. If a preempting run exits with a nonzero status, W&B will consider the run to have been preempted, and it will automatically append the run to a run queue associated with the sweep. If a run exits with no status, W&B will mark the run preempted 5 minutes after the run's final heartbeat, then add it to the sweep run queue. Sweep agents will consume runs off the run queue until the queue is exhausted, at which point they will resume generating new runs based on the standard sweep search algorithm.
```
kptkin: I don't think our users necessarily know what heartbeat means, or how often we send these...

docs/guides/runs/resuming.md
```diff
@@ -142,20 +69,23 @@ If multiple processes use the same `run_id` concurrently unexpected results will
...
+
+:::tip
+By default, requeued runs begin logging from their initial step. To instruct a run to resume logging at the step where it was interrupted, initialize the resumed run with `wandb.init(resume=True)`.
```
kptkin: i don't think that should be a different between resume=True and the other modes in terms of the starting step, as part of the check for the run we also get back from the server the last history step, so all modes should have the correct step when resuming

docs/guides/runs/resuming.md
```diff
@@ -11,116 +11,43 @@ import TabItem from '@theme/TabItem';
...
+<!-- START -->
 
-#### Automatic and controlled resuming
+### Automatic and controlled resuming
+
+Automatic resuming only works if the process is restarted on top of the same filesystem as the failed process. Set the `WANDB_RUN_ID` environment variable if you can not share a filesystem. For example:
 
-Automatic resuming only works if the process is restarted on top of the same filesystem as the failed process. If you can't share a filesystem, we allow you to set the `WANDB_RUN_ID`: a globally unique string (per project) corresponding to a single run of your script. It must be no longer than 64 characters. All non-word characters will be converted to dashes.
 
 ```python
```
kptkin, line 90: in this example maybe it makes sense to have 2 code snippets or one with a bash script that specify the env variables, since this is the preferred pattern.

docs/guides/runs/resuming.md
```diff
@@ -142,20 +69,23 @@ If multiple processes use the same `run_id` concurrently unexpected results will
 If you resume a run and you have `notes` specified in `wandb.init()`, those notes will overwrite any notes that you have added in the UI.
```
kptkin: re: If multiple processes use the same `run_id` concurrently unexpected...  <- i think it would be good to point the users to the docs about multi-processing. Since I saw users using resume as a way to log to the same run from multiple process, and this the wrong pattern. If the user want's to log from multiple process they should use multiprocessing instead of resume...

docs/guides/runs/resuming.md
```diff
@@ -11,151 +11,120 @@ import TabItem from '@theme/TabItem';
...
+
+:::info
+* If you resume a run and specify `notes` in `wandb.init()`, those notes will overwrite any notes that you make in the W&B App UI.
+:::
+
+
+## Enable runs to automatically resume 
+Automatic resuming only works if the process is restarted on top of the same filesystem as the failed process. 
+
+<!-- This only works if you run your script in the same directory as the one that failed as the file is stored at: `wandb/wandb-resume.json`. -->
```
ngrayluna: @kptkin  Wondering if we can elaborate more on this line "restarted on top of the same filesystem as the failed process".

Some questions:
1. What does that mean, exactly?
2. Does  W&B automatically create the `wandb-resume.json` file for them if a run stopped or failed?  Even if they didn't specify `resume`?

docs/guides/runs/intro.md
```diff
@@ -5,169 +5,176 @@ displayed_sidebar: default
...
+```
 
-wandb.init()
+:::info
+W&B stores runs in a project called "Uncategorized" if a project is not specified.
+:::
 
-assert wandb.run is not None
-```
+W&B will print the name of the run that was created along with a URL path to find out more information about that specific run. The URL path has the form `https://wandb.ai/entity/project-name/run-id`.
```
kptkin: hmmm... should we escape `entity`, `project-name` and `run-id`. I think it is a little confusing given the example above sets the `entity="wandbee"` and `project="model_registry_example"`?

docs/guides/runs/intro.md
```diff
@@ -5,169 +5,176 @@ displayed_sidebar: default
...
-Search for a specific run by name in the sidebar. You can use regex to filter down your visible runs. The search box affects which runs are shown on the graph. Here's an example:
+```python
+import wandb
 
-![](/images/app_ui/project_page_search_for_runs.gif)
+wandb.init()
+wandb.finish()
 
-### Filter runs
+assert wandb.run is None
```
kptkin: import wandb

```
run = wandb.init()
run.finish()
```

lets use this syntax, this will allow us to switch to multiple runs per process. 
since the global only picks up the latest

sorry - just realized that you wanted to show that we clear the run. I don't know if we want to encourage our users to do that based on the issue mentioned above

docs/guides/runs/intro.md
```diff
@@ -5,169 +5,176 @@ displayed_sidebar: default
...
 
-### Organize runs -->
+with wandb.init() as run:
+    pass  # log data here
 
+assert wandb.run is None
```
kptkin, line 184: i guess here is hard to avoid this...

docs/guides/runs/resuming.md
```diff
@@ -11,151 +11,120 @@ import TabItem from '@theme/TabItem';
   <title>Resume W&B Runs</title>
 </head>
 
-You can have W&B automatically resume runs by passing `resume=True` to `wandb.init()`. If your process doesn't exit successfully, the next time you run it W&B will start logging from the last step.
+This page covers how to automatically resume runs that crashed or finished. If you do not specify that you want runs to automatically resume crashed or finished runs, W&B will (by default), create a new run. This new run will overwrite the data of the crashed run if you the new run has the same run ID.
```
kptkin:
```suggestion
This page covers how to automatically resume runs that crashed or finished. If you do not specify that you want to automatically resume crashed or finished runs, W&B will (by default), create a new run. This new run will overwrite the data of the crashed run if you the new run has the same run ID.
```

docs/guides/runs/resuming.md
```diff
@@ -11,151 +11,120 @@ import TabItem from '@theme/TabItem';
...
+Set the `resume` parameter to True (`resume=True`) when you initialize a run with W&B (`wandb.init`). The following code snippet shows how to accomplish this with the W&B Python SDK:
+
+```python
+import wandb
 
-    This is our recommended method for resuming runs with W&B.
+run = wandb.init(
+  entity="<entity>", 
+  project="<project>", 
+  resume=True
```
kptkin: let's replace it with `resume="auto"` instead of `resume=True`. We want to have all of these to be string options (string literal). We convert  resume=True to resume="auto" under the hood

dmitryduev:
Hey @ngrayluna, could you please
```
pip install pre-commit
pre-commit install
pre-commit run blacken-docs <changed files>
```
it looks like the linter didn't run on the code samples

docs/guides/runs/resuming.md
kptkin: nope, it we won't create `wandb-resume.json` if resume is not specified. 
Only if `resume=auto` (or `resume=True` <- which let's avoid in the docs as this is something we want to deprecate).
when the user sets resume to `auto` we create the above file hence we need the next run to be from the same location to find the file and read the run-id from the file. 
An issue that users encountered is that we would pick up the file when they wanted to resume a new run, since the file is still in the same location. 

see this code snippet for the logic: https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_settings.py#L1862-L1880

docs/guides/runs/intro.md
ngrayluna: @kptkin  Yeah, this is tricky. I feel like we should steer them to "best practices". With that in mind. I changed removed
```
import wandb

wandb.init()
wandb.finish()

assert wandb.run is None

with wandb.init() as run:
    pass  # log data here

assert wandb.run is None
```
And left it only as this:

Screen Shot 2023-10-05 at 10 13 20 AM

Thoughts?

kptkin: I don't think the info is clear, also it is not required to call finish. If the user doesn't call finish, we will finish the run for them when the script ends. 

The only reason you would want to call finish mid script is if you want to start a new run (currently we only support one run per process which hopefully will change in the future as well)

Finish would mark "i'm done with this run" and for example you can't log to the run after finish (unless you resume the run)

docs/guides/runs/resuming.md
```diff
@@ -11,151 +11,131 @@ import TabItem from '@theme/TabItem';
   <title>Resume W&B Runs</title>
 </head>
 
-You can have W&B automatically resume runs by passing `resume=True` to `wandb.init()`. If your process doesn't exit successfully, the next time you run it W&B will start logging from the last step.
+This page covers how to automatically resume runs that crashed or finished. If you do not specify that you want to automatically resume crashed or finished runs, W&B will (by default), create a new run. This new run will overwrite the data of the crashed run if you the new run has the same run ID.
```
kptkin: > This new run will overwrite the data of the crashed run if you the new run has the same run ID.


Seems confusing without a context (do you want to say that if they use the same run_id and project)
maybe it should be some callout instead of default text?

docs/guides/runs/resuming.md
```diff
@@ -11,151 +11,131 @@ import TabItem from '@theme/TabItem';
...
+:::
 
-    More context regarding automatic and controlled resuming can be found in [this section](resuming.md#resume-runs).
-2. [`wandb.restore`](../track/save-restore.md#examples-of-wandb.restore)
-   * This will allow you to log new historical values for your metrics to a run starting from where you left off but does not take care of re-establishing the state of your code, you will need to make sure you have written checkpoints that you can load!
-   * You can use [`wandb.save`](../track/save-restore.md#examples-of-wandbsave) to record the state of your run via checkpoint files. Create a checkpoint file through `wandb.save()`, which can then be used through `wandb.init(resume=<run-id>)`. [This report](https://wandb.ai/lavanyashukla/save\_and\_restore/reports/Saving-and-Restoring-Models-with-W-B--Vmlldzo3MDQ3Mw) illustrates how to save and restore models with W&B.
 
-#### Automatic and controlled resuming
+## Resume a run without overriding the existing run
+Resume a run that stopped or crashed without overriding the existing run. This is especially helpful if  if your process doesn't exit successfully. The next time you start W&B, W&B will start logging from the last step.
```
kptkin:
```suggestion
Resume a run that stopped or crashed without overriding the existing run. This is especially helpful if  your process doesn't exit successfully. The next time you start W&B, W&B will start logging from the last step.
```

docs/guides/runs/resuming.md
```diff
@@ -11,151 +11,131 @@ import TabItem from '@theme/TabItem';
...
-os.environ["WANDB_RESUME"] = "allow"
-os.environ["WANDB_RUN_ID"] = wandb.util.generate_id()
-wandb.init()
+import wandb
+
+run = wandb.init(
+  entity="<entity>", 
+  project="<project>", 
+  resume="auto"
+  )
```
kptkin: just a comment auto is our least preferred way of resuming (i understand that it looks the easiest one, but i think we shouldn't highlight this one as much to be honest)

It is prone to most user errors and relies on having access to the file system with the run id file, i think we should flip the order and first focus on the other modes of resuming.

docs/guides/runs/resuming.md
```diff
@@ -11,151 +11,131 @@ import TabItem from '@theme/TabItem';
...
+  project="<project>", 
+  id="<run ID>", 
+  resume="must"
+  )
+```
 
 :::caution
-If multiple processes use the same `run_id` concurrently unexpected results will be recorded and rate limiting will occur.
-:::
+Unexpected results will occur if multiple processes use the same `run_id` concurrently.
```
kptkin: good call out!

ngrayluna:
@kptkin Tried to provide more clarity to resume options. 

Screen Shot 2024-05-14 at 9 44 41 AM


See a live preview on https://docs-beta.wandb.ai/guides/runs/resuming

docs/guides/runs/intro.md
```diff
@@ -5,169 +5,186 @@ displayed_sidebar: default
...
-
-Anything you log with `wandb.log` is recorded in that Run.  For more information on how log objects in W&B, see [Log Media and Objects](../track/log/intro.md).
-
-View Runs within a project within your Project's [Workspace](#view-runs). 
+:::tip
+Some key things to consider when you create and manage runs:
+* Anything you log with `wandb.log` is recorded in that run.  For more information on how log objects in W&B, see [Log Media and Objects](../track/log/intro.md). 
+* Each run is associated to a specific W&B project.
+* View runs and their properties within the run's project workspace on the W&B App UI.
+* There is only at most one active [`wandb.Run`](../../ref/python/run.md) in any process,
```
kptkin, line 24: @ngrayluna could you create a ticket for this part, we want to relax it in the new client, so when we have this would be good to re-write this part. For now it is totally correct. Just don't want to lose track of this!

docs/guides/runs/intro.md
```diff
@@ -5,169 +5,186 @@ displayed_sidebar: default
...
-assert wandb.run is not None
-```
 
+W&B will print the name of the run that was created along with a URL path to find out more information about that specific run. 
 
-You need to finish a Run that has not completed in order to start one or more Runs in the same notebook or script. 
+For example, the code snippet above produced this output:
 
 
+![](/images/runs/run_example.png)
```
kptkin, line 52: this one also going to change with the new client, so let's create a ticket to revise this once we have the new client! @ngrayluna

docs/guides/runs/intro.md
```diff
@@ -5,169 +5,186 @@ displayed_sidebar: default
...
 
 
-## End a Run
-W&B automatically calls [`wandb.finish`](../../ref/python/finish.md) to finalize and cleanup a run. However, if you call [`wandb.init`](../../ref/python/init.md) from a child process, you must explicitly call `wandb.finish` at the end of the child process. 
+Since we did not specify a name for the run, a random name generated for it by W&B called `likely-lion-9`. The run is stored in a project called `model_registry_example`. And it was given a *run ID* of `xlm66ixq`.
```
kptkin: maybe let's make it clear that is is a display name

kptkin: and the other is the run's name (or id) - 
just for your info - no need to communicate to the reader, we actually have a different id which is the backend id, and what we call id is a name and what we call a name is a display name.

docs/guides/runs/intro.md
```diff
@@ -5,169 +5,186 @@ displayed_sidebar: default
...
+
+
+:::info
+* W&B stores runs in a project called "Uncategorized" if a project is not specified.
```
kptkin: if we write things like this in the docs we need to be able to change it if we change the default, i feel like we don't have a good way to do this loop back... just a thought

docs/guides/runs/intro.md
```diff
@@ -5,169 +5,186 @@ displayed_sidebar: default
...
+
+```python
+import wandb
+
+run = wandb.init()
+run.finish()
+```
 
+:::info
+If you call [`wandb.init`](../../ref/python/init.md) from a child process, you must explicitly call [`wandb.finish`](../../ref/python/finish.md) at the end of the child process.
```
kptkin: that's not true, but it's a good practice

docs/guides/runs/intro.md
```diff
@@ -5,169 +5,186 @@ displayed_sidebar: default
...
-Search for a specific run by name in the sidebar. You can use regex to filter down your visible runs. The search box affects which runs are shown on the graph. Here's an example:
+The following code example demonstrates how to end a run from a `with` Python statement:
 
-![](/images/app_ui/project_page_search_for_runs.gif)
+```python
+import wandb
 
-### Filter runs
+wandb.init()
+wandb.finish()
```
kptkin, line 177: we try to encourage:
```
run = wandb.init()
run.finish()
```

usage pattern

docs/guides/runs/resuming.md
```diff
@@ -11,157 +11,137 @@ import TabItem from '@theme/TabItem';
...
-  <TabItem value="keras">
+:::tip
+W&B encourages you to provide the name of the W&B Project where you want to store the run.
+:::
+
+Pass one of the following arguments to the `resume` parameter to determine how a run should respond. In each case, W&B first checks if the run ID already exists. 
+
+|Argument | Description | Run ID exists| Run ID does not exist | Use case |
+| --- | --- | -- | --| -- |
+| `"must"` | W&B must use the run ID specified if run ID exists. | W&B resumes run with the same run ID. | W&B does nothing. | Resume a run that must use the same run ID.  |
```
kptkin: this is not correct in `must` if the run doesn't exists we error out!

docs/guides/runs/resuming.md
```diff
@@ -11,157 +11,137 @@ import TabItem from '@theme/TabItem';
...
+| `"allow"`| Allow W&B to resume run if run ID exists. | W&B resumes run with the same run ID. | W&B initializes a new run with specified run ID. | Resume a run without overriding an existing run. |
+| `"never"`| Never allow W&B to resume a run if run ID exists. | W&B does nothing. | W&B initializes a new run with specified run ID. | |
```
kptkin: it will error out if the run id exists

docs/guides/runs/resuming.md
```diff
@@ -11,157 +11,137 @@ import TabItem from '@theme/TabItem';
...
+<!-- - `"must"`:  If the run ID exists, W&B resumes the run with that run ID. If the run ID does not exist, W&B does nothing. 
+- `"allow"`:  If the run ID exists, W&B resumes the run with that run ID. If the run ID does not exist, W&B initializes a new run with the specified run ID.
+- `"never"`: If the run ID exists, W&B does nothing. If the run ID does not exist, W&B starts a new run with the specified run ID.  -->
+
+You can also specify `resume="auto"` to let W&B to automatically try to restart the run on your behalf. However, you will need to ensure that you restart your run from the same directory. See the [Enable runs to automatically resume](#enable-runs-to-automatically-resume) section for more information.
+
+For all the examples below, replace values enclosed within `<>` with your own.
+
+## Resume a run that must use the same run ID
+Resume a run that uses the same run ID if it crashes or fails. To do so, initialize a run and specify the following:
```
kptkin: it doesn't have to be crashed or failed, it could be just stopped.

docs/guides/runs/resuming.md
```diff
@@ -11,157 +11,137 @@ import TabItem from '@theme/TabItem';
...
-   * You can use [`wandb.save`](../track/save-restore.md#examples-of-wandbsave) to record the state of your run via checkpoint files. Create a checkpoint file through `wandb.save()`, which can then be used through `wandb.init(resume=<run-id>)`. [This report](https://wandb.ai/lavanyashukla/save\_and\_restore/reports/Saving-and-Restoring-Models-with-W-B--Vmlldzo3MDQ3Mw) illustrates how to save and restore models with W&B.
+The following example shows how to specify the W&B `WANDB_RUN_ID` variable in a bash script: 
 
-#### Automatic and controlled resuming
+```bash title="run_experiment.sh"
+RUN_ID="$1"
 
-Automatic resuming only works if the process is restarted on top of the same filesystem as the failed process. If you can't share a filesystem, we allow you to set the `WANDB_RUN_ID`: a globally unique string (per project) corresponding to a single run of your script. It must be no longer than 64 characters. All non-word characters will be converted to dashes.
+WANDB_RESUME=auto WANDB_RUN_ID="$RUN_ID" python eval.py
```
kptkin: this doesn't make sense... the entire idea of auto is the user doesn't need to keep track of the run id, and instead wandb does it for them

docs/guides/runs/intro.md
ngrayluna: Hmm...good point. 🤔

docs/guides/runs/resuming.md
ngrayluna: Good catch. I changed it to "allow"

docs/guides/runs/intro.md
ngrayluna: Jira ticket: https://wandb.atlassian.net/browse/DOCS-866

kptkin (approved):

PR #459, opened 2023-10-30
changed the early stopping and callbacks links to correct url

al-stev:
## Description

The early stopping and callback links gave a 404, amended them to the new links


cvphelps:
Thanks @al-stev ! @ngrayluna could you please review

ngrayluna (approved):

PR #555, opened 2024-01-11
adding warning to call teardown after sweeps within the same process

rymc:
## Description

This adds a caution to the docs after the code sample for starting an agent within a script or notebook. This caution lets users know that if they want to start a new run within the same script or notebook, they must call wandb.teardown() before initialising the new run. 

## Ticket
NA

## Checklist

- [ ] Build (`yarn docusaurus build`) was run locally and successfully without errors or warnings.


rymc:
@kptkin is this OK to recommend calling `wandb.teardown()` like this from the docs? I have seen only calling `wandb.finish() `cause problems a few times for customers where they want to initialize a new run within the same script or notebook after a sweep is complete, where they need to call `wandb.teardown()` instead.

PR #694, opened 2024-05-31
docs(launch): job inputs docs

bcsherma:
## Description

## Ticket

## Checklist

- [ ] Files I edited were previewed on my local development server with `yarn start`. My changes did not break the local preview.
- [ ] Build (`yarn docusaurus build`) was run locally and successfully without errors or warnings.
- [ ] I merged the latest changes from `main` into my feature branch before submitting this PR.

### DIFF
--- a/docs/guides/launch/job-inputs.md
+++ b/docs/guides/launch/job-inputs.md
@@ -1,0 +1,141 @@
+---
+displayed_sidebar: default
+---
+
+# Manage job inputs
+
+Add variable inputs to your job.
+
+## Intro
+
+Jobs have customizable inputs, enabling users to reconfigure the job when launching from the W&B CLI or UI.
+
+By default W&B Jobs capture the entire Run.config as the inputs to a job. The launch sdk provides function to select keys in the run config, or specify JSON or YAML files as inputs.
+
+For information on how job inputs can be set when launching from the CLI or UI, see the [Enqueue jobs](./add-job-to-queue.md) guide.
+
+:::info
+`launch` sdk functions require `wandb-core`. See the [`wandb-core` README](https://github.com/wandb/wandb/blob/main/core/README.md) for more information.
+:::
+
+## Reconfigure the `Run` object
+
+The `Run` object returned by `wandb.init` in a job can be reconfigured, by default. The `launch` sdk provides a way to customize what parts of the `Run.config` object can be reconfigured when launching the job.
+
+
+```python
+import wandb
+from wandb.sdk import launch
+
+# Required for launch sdk use.
+wandb.require("core")
+
+config = {
+    "trainer": {
+        "learning_rate": 0.01,
+        "batch_size": 32,
+        "model": "resnet",
+        "dataset": "cifar10",
+        "private": {
+            "key": "value",
+        },
+    },
+    "seed": 42,
+}
+
+
+with wandb.init(config=config)
+    launch.manange_wandb_config(
+        include=["trainer"], 
+        exclude=["trainer.private"],
+    )
+    # Etc.
+```
+
+The function `launch.manage_wandb_config` configures the job to accept input values for the `Run.config` object. The optional `include` and `exclude` options take path prefixes within the nested config object.
+
+If `include` prefixes are provided, only paths within the config that match an `include` prefix will accept input values. If `exclude` prefixes are provided, no paths that match the `exclude` list will be filtered out of the input values. If a path matches both an `include` and an `exclude` prefix, the `exclude` prefix will take precedence.
+
+In the preceding example, the path `["trainer.private"]` will filter out the `private` key from the `trainer` object, and the path `["trainer"]` will filter out all keys not under the `trainer` object.
+
+:::tip
+Use a `\`-escaped `.` to filter out keys with a `.` in their name. 
+
+For example, `r"trainer\.private"` filters out the `trainer.private` key rather than the `private` key under the `trainer` object.
+
+Note that the `r` prefix above denotes a raw string.
+:::
+
+If the code above is packaged and run as a job, the input types of the job will be:
+
+```json
+{
+    "trainer": {
+        "learning_rate": "float",
+        "batch_size": "int",
+        "model": "str",
+        "dataset": "str",
+    },
+}
+```
+
+When launching the job from the W&B CLI or UI, the user will be able to override the `trainer` parameters.
+
+### Access run config inputs
+
+Jobs launched with run config inputs can access the input values through the `Run.config`. The `Run` returned by `wandb.init` in the job code will have the input values automatically set. Use 
+```python
+from wandb.sdk import launch
+
+run_config_overrides = launch.load_wandb_config()
+```
+to load the run config input values anywhere in the job code.
+
+## Reconfigure a file
+
+The launch sdk can manage input values stored in config files in the job code.
+
+The `launch` sdk provides a way to manage input values stored in config files in the job code. The `launch.manage_config_file` function can be used to add a config file as an input to the job and apply input values to the file at runtime.
+
+By default, no run config inputs will be captured if `launch.manage_config_file` is used. Calling `launch.manage_wandb_config` overrides this behavior.
+
+Consider the following example:
+
+```python
+import yaml
+import wandb
+from wandb.sdk import launch
+
+# Required for launch sdk use.
+wandb.require("core")
+
+launch.manage_config_file("config.yaml")
+
+with open("config.yaml", "r") as f:
+    config = yaml.safe_load(f)
+
+with wandb.init(config=config)
+    # Etc.
+```
+
+Imagine the code is run with an adjacent file `config.yaml`:
+
+```yaml
+learning_rate: 0.01
+batch_size: 32
+model: resnet
+dataset: cifar10
+```
+
+The call to `launch.manage_config_file` will add the `config.yaml` file as an input to the job, making it reconfigurable when launching from the W&B CLI or UI. 
+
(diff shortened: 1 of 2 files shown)


docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,139 @@
+---
+displayed_sidebar: default
+---
+
+# Reconfigure jobs
+
+Add inputs to a job to make it reconfigurable.
```
KyleGoyette: The title reconfigure jobs feels incorrect to me. A better title would be Configuring Job Inputs IMO

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Intro
+
+Job input types are an object schema defining the variable inputs of a job. Job output types are an object schema for the keys logged with `wandb`. The `wandb` library captures and saves the input and output types of the job at runtime.
```
KyleGoyette: I don't think we should dive right into the types here. First lets lets describe inputs, how they are configured. I would totally ignore output types right now they don't do anything

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,139 @@
...
+
+Jobs create a single `wandb.Run` to track their execution. By default, the structure of `Run.config` and `Run.summary` determine the input and output types of a job. You can override these defaults by specifying the `input_types` and `output_types` of a job.
```
KyleGoyette: This will confuse users IMO, they will think Jobs automatically call wandb.init  when they don't.

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,139 @@
...
+
+By default the full structure the input types of a job are taken from the full `Run.config`. This set of keys can be filtered using the `launch` sdk.
+
+```python
+import wandb
+from wandb.sdk import launch
+
+# Required for launch sdk use.
+# See https://github.com/wandb/wandb/blob/main/core/README.md for more info.
+wandb.require("core")
```
KyleGoyette, line 31: This deserves a call out IMO

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,139 @@
...
+
+The `launch` sdk provides ways to customize inputs. Job outputs cannot be customized. 
+
+:::tip
+For information on how job inputs can be set when launching from the CLI or UI, see the [Enqueue jobs](./add-job-to-queue.md) guide.
+:::
+
+## Reconfigure with `Run.config`
+
+### Customize run config inputs
```
KyleGoyette: It's really more filtering or controlling as opposed to customizing no?

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,139 @@
...
+with wandb.init(config=config)
+    launch.manange_wandb_config(
+        config, 
+        include=["trainer"], 
+        exclude=["trainer.private"],
+    )
+    # Etc.
+```
+
+The call to `launch.manage_wandb_config` will filter the `Run.config` to only include the keys in `include` and exclude the keys in `exclude`. The `include` and `exclude` arguments are optional.
```
KyleGoyette: This is unclear IMO. Whats the behavior if the optional args are left blank?

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,139 @@
...
+
+Both `include` and `exclude` take a list of paths within the corresponding object and filter keys in or out of the job input types. The paths are specified as a list of keys separated by a period. For example, the path `["trainer.private"]` will filter out the `private` key from the `trainer` object, and the path `["trainer"]` will filter out all keys not under the `trainer` object.
```
KyleGoyette: Overall I don't love the wording here, and it wasn't clear to me that we were referring to the example above.

I'd rephrase this. And make it in terms of precedence. Default: whole config, include limits to the include list, exclude takes precedence over include I think?

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,139 @@
...
+        "dataset": "str",
+    },
+}
+```
+
+When launching the job from the W&B CLI or UI, the user will be able to override the `trainer` parameters.
+
+### Access run config inputs
+
+Jobs launched with run config inputs provided access the input values through the `Run.config`. Launch provides the input values through the `WANDB_CONFIG` environment variable in the job environment. The `WANDB_CONFIG` environment variable is a JSON string of the input values. The run returned by `wandb.init` in the job will have the input values automatically set in `Run.config`.
```
KyleGoyette: I'd drop the WANDB_CONFIG info and say access it via the run object config key, or if you're in DDP use the util function

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+The call to `launch.manage_config_file` will add the `config.yaml` file as an input to the job, making it reconfigurable when launching from the W&B CLI or UI. 
+
+The `include` and `exclude` keyword arugments may be used to filter the input values for the config file in the same way as `launch.manage_wandb_config`.
+
+
+### Access config file inputs
+
+When `launch.manage_config_file` is called `launch` patches the contents of the config file with the input values provided by the user. The patched config file is available in the job environment.
```
KyleGoyette: This only happens when the job is launched, we should make that more clear

KyleGoyette (commented):
A few things:
1. I would ignore outputs entirely here. This doc is solely about inputs. If we discuss outputs we should put it at the bottom, we should also have a tutorial lined up showing how they are useful.
2. The structure here feels off to me. I think we should start by: describing job inputs, note they default to the run config. Introduce the manage api, show how it can be used to filter config and control files.

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,143 @@
...
+
+# Configure job inputs
+
+Add variable inputs to your job.
+
+## Intro
+
+Jobs have customizable inputs, enabling users to reconfigure the job when launching from the W&B CLI or UI.
+
+All jobs accept overrides to their entry point command and arguments.
```
KyleGoyette: This doesn't feel relevant to this section. Here we're discussing the inputs defined by confgs and files

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,143 @@
...
+
+Inputs can also be set using the `Run.config` or a config file in the job code. The `launch` sdk provides functions to manage these inputs.
```
KyleGoyette: "By default W&B Jobs capture the entire `Run.config` as the inputs to a job. The `launch` sdk provides function to select keys in the run config, or specify JSON or YAML files as inputs"

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,143 @@
...
+
+For information on how job inputs can be set when launching from the CLI or UI, see the [Enqueue jobs](./add-job-to-queue.md) guide.
```
KyleGoyette, line 15: Out of scope here. But I think we should combine aspects of "Create a launch job" and "Enqueue jobs" into a single page about launching jobs. There we should discuss configuring inputs.

:/ On that note "Manage job inputs may make more sense as the title. Since this API is the manage API.

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,143 @@
...
+    launch.manange_wandb_config(
+        include=["trainer"], 
+        exclude=["trainer.private"],
+    )
+    # Etc.
+```
+
+The function `launch.manage_wandb_config` configures the job to accept input values for the `Run.config` object. The optional `include` and `exclude` options take path prefixes within the nested config object.
+
+If `include` prefixes are provided, only paths within the config that match an `include` prefix will accept input values. If `exclude` prefixes are provided, no paths that match the `exclude` list will be filtered out of the input values. If a path matches both an `include` and an `exclude` prefix, the `exclude` prefix will take precedence.
```
KyleGoyette, line 57: Much clearer to me now

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,143 @@
...
+
+Inputs can also be set using the `Run.config` or a config file in the job code. The `launch` sdk provides functions to manage these inputs.
+
+For information on how job inputs can be set when launching from the CLI or UI, see the [Enqueue jobs](./add-job-to-queue.md) guide.
+
+:::info
+`launch` sdk functions require `wandb-core`. See the [`wandb-core` README](https://github.com/wandb/wandb/blob/main/core/README.md) for more information.
+:::
+
+## Reconfigure the `Run` object
```
KyleGoyette, line 21: I recommend saying Reconfiguring the `Run Config`

docs/guides/launch/job-inputs.md
```diff
@@ -0,0 +1,143 @@
...
+
+The `Run` object returned by `wandb.init` in a job can be reconfigured, by default. The `launch` sdk provides a way to customize what parts of the `Run.config` object can be reconfigured when launching the job.
```
KyleGoyette, line 23: And referring to the `Run.config` here, we don't allow users to reconfigure the whole run. Just the config

KyleGoyette (approved):
LGTM pending my recommended changes

docs/guides/launch/job-inputs.md
bcsherma: That's how I was phrasing this everwhere before and I deliberately changed it. I think saying "reconfigure the configuration of X" is less clear than "reconfigure X"

KyleGoyette: I'm fine with trying it this way, I think if users get confused we'll find out pretty quickly

jcrawley (approved):

PR #737, opened 2024-07-02
Download and use an artifact from a registry

ngrayluna:
## Description

Adds a doc on how to download and use an artifact from registry

## Ticket

https://wandb.atlassian.net/browse/DOCS-900

## Checklist


docs/guides/model_registry/consume-models.md
noaleetz, line 1: since users will still be consuming models for a period of time until we deprecate the old model registry (the timeline for this will be end of next Q (Q4), should we consider leaving this up (we can (1) rename it to make it clear that is specifically for the old team model registry and (2) put a banner saying that the old team model registry will be deprecated, to learn how to consume models from W&B registry, see X)?

noaleetz (approved):
added a new commit version w changes

PR #833, opened 2024-09-25
Custom CAs in operator

mkaesz-wandb:
## Description

How to configure Custom CAs in operator.

## Ticket

## Checklist

- [ x] Files I edited were previewed on my local development server with `yarn start`. My changes did not break the local preview.
- [x ] Build (`yarn docusaurus build`) was run locally and successfully without errors or warnings.
- [ x] I merged the latest changes from `main` into my feature branch before submitting this PR.


flamarion (commented):
Not directly connected with the Custom CA, but I wonder if we really [this](https://github.com/wandb/docodile/blob/custom-ca/docs/guides/hosting/operator.md?plain=1#L440-L441) or if this is only an example.
If it's an example, I suggest we use a different env var, this can be confused. In the end, when setting up the bucket, we already set the `region` option that will become that same env var, so adding it in a different place will override what we're adding for the bucket, or vice versa.

flamarion (approved):
LGTM 🚢

docs/guides/hosting/operator.md
```diff
@@ -705,6 +703,60 @@ global:
...
+    SAAwRQIhAIzXZMW44l6XMf9Nf4TxTevK8vE4Ic6E8UFqsCcILdXjAiA7iTluM0IU
+    aIgJYVqKxXt25blH/VyBRzvNhViesfkNUQ==
+    -----END CERTIFICATE-----
+```
+
+## Configuration Reference (W&B Operator)
+
+This section describes the configuration options for W&B Kubernetes operator (wandb-controller-manager). The operator receives its configuration in the form of a YAML file. 
+
+By default, the W&B Kubernetes operator does not need any configuration file and should only be created if actually needed.
```
ngrayluna: @mkaesz-wandb  What is an example or two if when someone would need to create a config file?

ngrayluna (commented):
Left one QQ.

docs/guides/hosting/operator.md
mkaesz-wandb: Added two examples. Thanks!

PR #845, opened 2024-10-02
Fix links, images, metadata, Vale rules, and titles

johndmulhausen:
## Description

- Fixes all known broken links
- Brings in all externally-sourced images from providers like imgur, and optimizes them (both actions will vastly improve their load time and make them 'tracked' by Docusaurus if moved/broken)
- Fixes Vale to turn off some of the noisier comments
- Changes titles for Guides docs to be action-based and sentenced-cased, per Google style guide
- Changes titles to use front-matter consistently enabling proper metadata structure and consolidation of titling approach
- Changes colab links in tutorials to use react tag
- Changes links to docs.wandb.com and docs.wandb.ai to be relative/local so they will be tracked by Docusaurus
- Various "that looks bad" fixes, as I was sweeping through, especially in tutorials
- Establishes Lychee CI job so that external links will be checked in future PRs


ngrayluna (approved):
Siiiiiiick.  Approved.

Curious what the network errors are about? 

For future reference, *some* of the tutorials are autogenerated from a separate repo from notebooks. Hence, why they use non relative links, etc. Here's the list of notebooks autogenerated: https://github.com/wandb/examples_ci/blob/master/tutorials.txt

PR #914, opened 2024-11-06
Move Launch docs out of main navigation

johndmulhausen:
Moves docs out of main navigation and into their own isolated navigation. This is to stop promotion of Launch workflows for new users while still supporting users who currently use Launch.

To see it in action at the "secret URL" go to https://launch-docs.docodile.pages.dev/guides/launch


ngrayluna:
We can ignore the Check markdown header since we are being intentional here about not wanting Launch in the docs.

QQ: Should we add a banner at the top of the launch docs saying that W&B has no plans to support after X date?

ngrayluna (approved):

PR #939, opened 2024-12-03
Removed obsolete Redis config

mkaesz-wandb:
## Description

Removed obsolete Redis config

## Ticket

N/A

## Checklist


venky-wandb:
@mkaesz-wandb This is being handled by the operator helm chart now, is that why we are removing this?

mkaesz-wandb:
> @mkaesz-wandb This is being handled by the operator helm chart now, is that why we are removing this?

Old, at the wrong place and completely out of context. I want to add a new section at the right place.

venky-wandb:
Would it be possible to also add it to the same PR? I think this is still very relevant to anyone non-operator, so if we remove it and it's not available anywhere we will be scrambling to find the docs for REDIS.

ngrayluna (approved):

PR #980, opened 2025-01-13
Update docsearchConfig.json for Hugo

ngrayluna:
## Description

Fixes docsearchConfig.json so we can index the site and use search on our new Hugo site.  

## Ticket

https://wandb.atlassian.net/browse/DOCS-1195


docsearchConfig.json
```diff
@@ -2,27 +2,31 @@
...
+  "stop_urls": [
+    "https://docs.wandb.ai/ja",
+    "https://docs.wandb.ai/ko",
+    ".*_print.*"
+  ],
   "sitemap_alternate_links": false,
   "selectors": {
     "lvl0": {
-      "selector": "(//ul[contains(@class,'menu__list')]//a[contains(@class, 'menu__link menu__link--sublist menu__link--active')]/text() | //nav[contains(@class, 'navbar')]//a[contains(@class, 'navbar__link--active')]/text())[last()]",
+      "selector": "(//ul[contains(@class,'navbar_nav')]//a[contains(@class, 'nav_link active')]/text() | //ul[contains(@class,'ul_0')]//a[contains(@class, 'td-sidebar-link active')]/text())",
```
mdlinville, line 13: Were you able to confirm that the nav is no longer being indexed as content, with this change?

ngrayluna: How do you mean?

mdlinville: Oh, I thought John mentioned something about the search results seeming to show up multiple times, one for the content and one for the navigation. We speculated that maybe this code was trying to exclude the top and left nav themselves from showing up as search results. Maybe I completely misunderstood.

ngrayluna: Ah. Yeah, that existed before too when we were on Docusaurus 😬

mdlinville (approved):

PR #981, opened 2025-01-14
Docs share panels

ngrayluna:
## Description

Docs on linking to a specific panel.

## Ticket

https://wandb.atlassian.net/browse/DOCS-1156

### DIFF
--- a/content/guides/models/app/features/panels/_index.md
+++ b/content/guides/models/app/features/panels/_index.md
@@ -103,7 +103,7 @@ To remove all panels from a manual workspace, click its action `...` menu, then 
 
 To remove all panels from an automatic or manual workspace, you can [reset the workspace](#reset-a-workspace). Select **Automatic** to start with the default set of panels, or select **Manual** to start with an empty workspace with no panels.
 
-### Show specific panels to members of your organization
+### Share panels
 Direct colleagues to a specific panel in your project. The link redirects users to a full screen view of that panel when they click that link. To create a link to a panel:
 
 1. Hover your mouse over the panel.
--- a/content/guides/models/app/features/panels/_index.md
+++ b/content/guides/models/app/features/panels/_index.md
@@ -59,7 +59,7 @@ To add a custom panel to your workspace:
 1. Follow the prompts to configure the panel.
 
 To learn more about the options for each type of panel, refer to the relevant section below, such as [Line plots](line-plot/intro.md) or [Bar plots](bar-plot.md).
-## Share panels
+
 
 ## Manage panels
 
--- a/content/guides/models/app/features/panels/_index.md
+++ b/content/guides/models/app/features/panels/_index.md
@@ -103,7 +103,7 @@ To remove all panels from a manual workspace, click its action `...` menu, then 
 
 To remove all panels from an automatic or manual workspace, you can [reset the workspace](#reset-a-workspace). Select **Automatic** to start with the default set of panels, or select **Manual** to start with an empty workspace with no panels.
 
-### Share panels
+### Share a full-screen panel directly
 Direct colleagues to a specific panel in your project. The link redirects users to a full screen view of that panel when they click that link. To create a link to a panel:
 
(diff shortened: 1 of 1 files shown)


content/guides/models/app/features/panels/_index.md
```diff
@@ -102,6 +103,17 @@ To remove all panels from a manual workspace, click its action `...` menu, then
 
 To remove all panels from an automatic or manual workspace, you can [reset the workspace](#reset-a-workspace). Select **Automatic** to start with the default set of panels, or select **Manual** to start with an empty workspace with no panels.
 
+### Show specific panels to members of your organization
```
mdlinville:
```suggestion
### Share a panel
```
From reading the description, it doesn't look like it's necessarily limited to members of the organization. My instinct is to make the heading shorter and more generic (and more searchable). WDYT?

content/guides/models/app/features/panels/_index.md
```diff
@@ -102,6 +103,17 @@ To remove all panels from a manual workspace, click its action `...` menu, then
...
+Direct colleagues to a specific panel in your project. The link redirects users to a full screen view of that panel when they click that link. To create a link to a panel:
+
+1. Hover your mouse over the panel.
+2. Select the action `...` menu.
+3. Click **Copy panel URL**.
+
+The settings of the project determine who can view the panel. This means that if the project is private, only members of the project can view the panel. If the project is public, anyone with the link can view the panel.
+
+If you have multiple panels with the same name, W&B directs users to the first panel with that name.
```
mdlinville:
```suggestion
If multiple panels have the same name, W&B shares the first panel with the name.
```
The first according to what? Maybe we should advise not to get into this situation.

mdlinville (approved):

content/guides/models/app/features/panels/_index.md
ngrayluna: It is limited to folks within the org. If folks don't have org access, you get denied page

ngrayluna: Yeah, not a fan of my original title.  Trying to come up with a title that is accurate and yet not called the same thing as another feature...

From Tristan: 

"The main point of confusion may relate to the difference between this and the “Share Panel” option.   This “Direct Panel Sharing” option, the new feature we’re talking about, respects your project’s access controls and can’t be shared publicly with the magic link.  Unlike the share panel one." 

(Taken from ticket)

mdlinville: Nothing about the UI makes me realize that this one respects project access controls and the other one can't. Why would we want to leave something around that defeats a project's access controls? Maybe I'm missing some background. It's hard to see how docs could clear up confusion (which makes sense) between this and a different "Share panel" option.

ngrayluna: It's a magic link.  See this ticket for what Tristan says: https://wandb.atlassian.net/browse/DOCS-1156

content/guides/models/app/features/panels/_index.md
```diff
@@ -102,6 +103,17 @@ To remove all panels from a manual workspace, click its action `...` menu, then
 
 To remove all panels from an automatic or manual workspace, you can [reset the workspace](#reset-a-workspace). Select **Automatic** to start with the default set of panels, or select **Manual** to start with an empty workspace with no panels.
 
+### Share a full-screen panel directly
```
mdlinville, line 106: LGTM

mdlinville (approved):

content/guides/models/app/features/panels/_index.md
tristan-spaulding2: The old share panel is meant for publishing a panel to the public, e.g. in a tweet.   It is not from what I can tell a significant use case in terms of usage, more of a viral marketing concept from the early days of W&B.  I would expect our direction to more and more emphasize this within-org sharing action in favor of this, though we have not made those changes in this initial release.

PR #991, opened 2025-01-15
Fix mixed-up CLI / Python labels for Sweep configuration examples

AudreyBeard:
Sweep configuration options (CLI / Python) examples were mixed up, such that the Python dictionary specification was labeled "CLI" and the YAML specification was labeled "Python script or Jupyter notebook".

## Description

This change fixes the mislabeled Sweep configuration by correctly labeling each of the two examples.

## Ticket

N/A

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -173,9 +173,9 @@ params:
       enable: true
       # The responses that the user sees after clicking "yes" (the page was helpful) or "no" (the page was not helpful).
       'yes': >-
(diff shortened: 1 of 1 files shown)


johndmulhausen:
@AudreyBeard Thank you for this PR! I think we would want to swap _the actual code inside the tabs_ in order to keep the order of the tabs consistent on the two places where they appear. I also need to get in there and figure out why there's some `</div>` being printed on the page https://docs.wandb.ai/guides/sweeps/define-sweep-configuration/

At any rate, do you want to do another pass at this with the code swapped or should I just jump in? The PR is very much appreciated 🙏

PR #997, opened 2025-01-16
Add support for Hugo-native multi-language functionality

johndmulhausen:
## Description

Rather than use multiple branches and shell script logic to produce the Japanese and Korean builds, this PR aims to simply use the native functionality built directly into Hugo for featuring docsets for difference locales. 

This PR: 
- Adds script that wipes out, then replaces, `ko` and `ja` subdirectories within `content` with the English docs, modified for localized build (given target `lang` values in relref, and given language-specific menu IDs).
- Modifies a number of links that somehow became ambiguous for Hugo in a multi-language situation
- Modifies config file to refer to new content sources
- Adds a fresh "seed" of English content, ready to be translated

TODO: 
- Actually translate the "seed" of content. 

## Ticket

Resolves DOCS-1188

## To preview

See: https://docs-i18n-test.pages.dev

## Note

The production script, `/scripts/production.sh` is deleted in this branch, as CloudFlare's native/default Hugo environment becomes compatible with our repo once this merges, cutting build times from 6:30-7:00 minutes to 2:00-2:30 minutes!


johndmulhausen:
Closing due to commit/rebase drift

PR #998, opened 2025-01-16
Registry: Add doc on collection cards

ngrayluna:
## Description

Unlike how we documented Model Registry cards, I kept this page collection type agnostic since users can create any collection type. 

## Ticket

https://wandb.atlassian.net/browse/DOCS-1096


ngrayluna:
Duplicate. Closing in favor of: https://github.com/wandb/docs/pull/999

PR #1025, opened 2025-01-24
Override Docsy lastmod template to remove subject

mdlinville:
Override Docsy lastmod template to remove subject

Screenshot 2025-01-24 at 2 40 14 PM

Adapted from https://github.com/google/docsy/blob/a854cb315ac2885f8e27a5780d3c320a30d1b457/layouts/partials/page-meta-lastmod.html#L2

## Ticket

Ref: https://weightsandbiases.slack.com/archives/C081R4TDAH1/p1737746004048749


johndmulhausen (approved):
Hopefully they don't edit this file upstream and we don't get some crucial update in the future, but I agree it looks better this way. I suppose if we notice something off we can always cross-check this file against `google/docsy` and pull in the update

PR #1030, opened 2025-01-26
Update _index.md

laminendao:
## Description

What does the pull request do? 

## Ticket

Does this PR fix an existing issue? If yes, provide a link here.


mdlinville:
Thanks for the contribution. Unless I'm missing something, this is a whitespace-only change and has no result in the output. Typically, we would not accept a whitespace-only change unless the existing output is causing an issue. Please let me know if I'm missing the point of the contribution.

PR #1034, opened 2025-01-27
Try to fix Lychee github action error

mdlinville:
This fixes the Lychee GitHub action so it doesn't fail on each PR

- Do not run as a PR check
- Run against `main` using a cron schedule, for 12:01 AM on the 1st of each month
- Ignore 429 errors
- Adjust the syntax for specifying args
- Give it permission to create GH issues
- Upgrade the action

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -3,7 +3,7 @@ on:
   repository_dispatch:
   workflow_dispatch:
   schedule:
-    - cron: "1 0 1 * *" # 12:01 on the 1st of each month
+    - cron: "1 0 2 * *" # 12:01 on the 1st of each month
 
 jobs:
   linkChecker:
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -1,6 +1,9 @@
 name: Lychee check
 on:
-  pull_request:
+  repository_dispatch:
+  workflow_dispatch:
+  schedule:
+    - cron: "41 14 * * *"
 
 jobs:
   linkChecker:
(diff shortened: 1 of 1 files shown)


.github/workflows/linkcheck.yml
```diff
@@ -21,11 +40,7 @@ jobs:
         uses: lycheeverse/lychee-action@v1.10.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: > 
-          --scheme https --scheme http --verbose --no-cache
-          --user-agent "Mozilla/5.0 (Windows NT 10.0; Win64; x64)"
-          --max-concurrency 5 --retry-count 3 --retry-delay 2
-          ${{ steps.changed-files.outputs.all_changed_files }}
+          args: --base content --scheme https --scheme http --verbose --no-cache --user-agent "Mozilla/5.0 (Windows NT 10.0; Win64; x64)" --max-concurrency 5 --retry-count 3 --retry-delay 2 ${{ steps.changed-files.outputs.all_changed_files }}
```
johndmulhausen:
```suggestion
          args: --accept 200,429 --base content --scheme https --scheme http --verbose --cache --user-agent "Mozilla/5.0 (Windows NT 10.0; Win64; x64)" --max-concurrency 5 --retry-count 1 --retry-delay 2 ${{ steps.changed-files.outputs.all_changed_files }} --github-token "${{ secrets.GITHUB_TOKEN }}"
```

This is to reduce this problem: https://lychee.cli.rs/troubleshooting/rate-limits/

johndmulhausen (dismissed):
The reason I think we should accept 429 responses is to clean up GitHub's rate limiting errors from the reports. We pass in a token and that should be giving us 1000 requests and hour but with the number of commits we push and number of links we have it seems to be hitting the ceiling.

In addition, I think we should convert this to a cron job that we run, say, monthly? It's going to file github issues when there's an issue (and it already has filed several from your tests).

.github/workflows/linkcheck.yml
mdlinville: OK I tried to address this one, your feedback above, and also to add the things needed to use `--cache`, in the most recent commit. I was hoping that setting the cron to a specific time I could watch the Actions tab in the repo and find out if it worked. But maybe the PR has to be committed for that to work. I modeled the change on https://github.com/lycheeverse/lychee-action, which passes the args using a different syntax than we were using previously.

.github/workflows/linkcheck.yml
```diff
@@ -1,37 +1,54 @@
 name: Lychee check
 on:
-  pull_request:
-    paths:
-      - '**.md'
+  repository_dispatch:
+  workflow_dispatch:
+  schedule:
+    - cron: "1 0 1 * *" # 12:01 on the 1st of each month
```
mdlinville:
```suggestion
    - cron: "1 0 2 * *" # 12:01 on the 1st of each month
```
Temporarily set this to the 2nd so we can test it

johndmulhausen (approved):

PR #1040, opened 2025-01-27
[DOCS-1214] Document 'ignore outliers' behavior for full fidelity plots

mdlinville:
[DOCS-1214] Document 'ignore outliers' behavior for full fidelity plots

[DOCS-1214]: https://wandb.atlassian.net/browse/DOCS-1214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jtulk (commented):
For sampling: there's no logic that actually removes "outliers" in the sense that outliers are points that deviate significantly from the overall trend line of the plot. It naively removes the top and bottom 5% of points. 

For full fidelity: "Exclude outliers" doesn't actually have anything to do with outliers. It toggles the shaded area that represents the min/max bounds of each "bucket". A "bucket" is a series of points condensed down into a single point. We render the last value in the bucket. So a series of points like:
```
[0,0], [1,2], [2,4], ..., [10,20], [11, 4]
```
would render as a single point at [11,4]. A shaded area would render at [11, 0, 20] where 0 and 20 represent the min and max values in the bucket. 

This shows "outliers" because they don't disappear into the aggregation in the bucket. You can still see spikes in the shaded min/max areas.

If you toggle on `exclude outliers` you will no longer see these shaded regions, but will instead only see the line produced by rendering the last point in each bucket.

mdlinville:
> For sampling: there's no logic that actually removes "outliers" in the sense that outliers are points that deviate significantly from the overall trend line of the plot. It naively removes the top and bottom 5% of points.
> 
> For full fidelity: "Exclude outliers" doesn't actually have anything to do with outliers. It toggles the shaded area that represents the min/max bounds of each "bucket". A "bucket" is a series of points condensed down into a single point. We render the last value in the bucket. So a series of points like:
> 
> ```
> [0,0], [1,2], [2,4], ..., [10,20], [11, 4]
> ```
> 
> would render as a single point at [11,4]. A shaded area would render at [11, 0, 20] where 0 and 20 represent the min and max values in the bucket.
> 
> This shows "outliers" because they don't disappear into the aggregation in the bucket. You can still see spikes in the shaded min/max areas.
> 
> If you toggle on `exclude outliers` you will no longer see these shaded regions, but will instead only see the line produced by rendering the last point in each bucket.

I tried again to explain. Please let me know if I've got it right this time.

jtulk (approved):
Works for me - think you've got a typo in `consdensed`

content/guides/models/app/features/panels/line-plot/reference.md
```diff
@@ -67,7 +67,10 @@ You can set the [smoothing coefficient]({{< relref "/support/formula_smoothing_a
 
 ## Ignore outliers
 
-Ignore outliers sets the graph's Y-axis range from 5% to 95%, rather than from 0% to 100% to make all data visible.
+Rescale the plot to exclude outliers from the default plot min and max scale. The setting's impact on the plot depends on the plot's sampling mode.
+
+- For plots that use [random sampling mode]({{< relref "sampling.md#random-sampling" >}}), when you enable **Ignore outliers**, only points from 5% to 95% are shown. When outliers are shown, they are not formatted differently from other points.
+- For plots that use [full fidelity mode]({{< relref "sampling.md#full-fidelity" >}}), all points are always shown, consdensed down to the last value in each bucket. When **Ignore outliers** is enabled, the minimum and maximum bounds of each bucket are shaded. Otherwise, no area is shaded.
```
mdlinville:
```suggestion
- For plots that use [full fidelity mode]({{< relref "sampling.md#full-fidelity" >}}), all points are always shown, condensed down to the last value in each bucket. When **Ignore outliers** is enabled, the minimum and maximum bounds of each bucket are shaded. Otherwise, no area is shaded.
```

ngrayluna (approved):

PR #1045, opened 2025-01-29
Fix issue where double dashes are currently not shown correctly on the website

V0XNIHILI:
Currently, only a single dash is shown for these arguments.

{Short description}

{More details, or remove this line}

-- Remove this line and following from the merge commit message --

{JIRA / GitHub issue, or remove this line}
{Upstream code PR, or remove this line}

Confirm that tests pass and this change is ready for review:
- [ ] Local `hugo serve`
- [ ] Local `vale ./content`


mdlinville (changes requested):
Thanks for this. All of the pages under `ref/cli` seem to have the same issue. I wonder if you would consider fixing them all. There does not seem to be a way to automatically format the first column of these tables to be in code font, like there is a way to set the alignment manually (this table sets it to left-justified by having the `:` only on the left side).

johndmulhausen:
Hi, we appreciate this, but we have to close this PR as it is editing generated content. This is also re: #1060 and we are internally tracking a fix that will ship across all CLI docs. Thanks so much, @V0XNIHILI we hope to see more contributions someday :)

PR #1055, opened 2025-01-31
Make site compatible with DocSearch crawler

johndmulhausen:
Related to: DOCS-1251

The idea here is to make the site properly index using Algolia's DocSearch product, then build the index in their dashboard, and finally ship a PR that activates DocSearch using the new index.


mdlinville:
Screenshot 2025-01-31 at 1 41 10 PM

I wonder if it is possible to disable having each heading (in the same file) show up as a separate search result. Overall, this seems better than the situation with Algolia before.

hugo.yaml
```diff
@@ -88,11 +88,11 @@ services:
...
 
 params:
-  #search:
-  #  algolia:
-  #    appId: 2D210VB5MP
-  #    apiKey: 730cfa02025b8ba2e95d4c33b1e38cc7
-  #    indexName: docodile
+  search:
+    algolia:
+      appId: 6HH47A49YJ
```
ngrayluna: Should we be hiding our API keys? E.g. in a .env or .gitignore? (apiKey)

hugo.yaml
```diff
@@ -140,7 +140,7 @@ params:
   github_branch: main
 
   # Google Custom Search Engine ID. Remove or comment out to disable search.
-  gcs_engine_id: 674c0c3807d874d09
+  # gcs_engine_id: 674c0c3807d874d09
```
mdlinville: I suspect we should remove this so we don't expose the ID since we are not currently using it.

ngrayluna:
Some maybe useful pages within the Algolia account app? 

Searchable attributes:
Search -> Index -> Configuration
Screenshot 2025-01-31 at 2 05 37 PM

UI return results demo. Could be useful for seeing what it takes to generate bread crumbs...
Search -> Index ->  UI Demos -> Generate UI Demos -> Generate a new demo -> Instant Search

It would seem that based on the Configuration, that you can demo the results...how that then translates back to the site is not entirely clear...
Screenshot 2025-01-31 at 2 08 51 PM

hugo.yaml
johndmulhausen: We never did before; I think it's fine, that's the pattern you use for things like GTM as well, and this is the documented way Docsy says to specify your `appId`

hugo.yaml
johndmulhausen:
```suggestion
```

hugo.yaml
ngrayluna: Our repo was private before ;)

PR #1069, opened 2025-02-04
Prism "copy" button fix

johndmulhausen:
Fixes HTML bug in instances where an external file is being loaded in the `prism` shortcode. 

Fixes DOCS-1227

To test go to https://prism-fix.docodile.pages.dev/guides/artifacts/project-scoped-automations/ , then search for "Bash script" and try to copy the code block that features the "Download" button. The code does make it to your clipboard, unlike the live version at https://docs.wandb.ai/guides/artifacts/project-scoped-automations/


mdlinville (approved):
I don't understand it but it works!

ngrayluna (approved):

PR #1104, opened 2025-02-19
[Do no merge] Lazydocs SDK DataTypes doc preview.

ngrayluna:
This PR is for previewing modifications made to `wandb/wandb/sdk/data_types/*.py` created from this PR: https://github.com/wandb/wandb/pull/9220/files

To generate these files I used [ngrayluna/generate-wandb-python-reference](https://github.com/ngrayluna/generate-wandb-python-reference). As discussed here, there are some methods, etc. that are public even though they should be hidden by preprending an underscore (`_`) to said methods. Changing these methods today would require updating CI tests (which was/is not advices by SDK Team). The work around is to add this string `<!-- lazydoc-ignore: internal -->` within each method's, etc. docstring that is exposed, but shouldn't be.

**Doc Team**: Do not review (yet). 

**SDK Team**: For this first pass, can ya'll comment on:
[ ] Methods, Classes etc. that should not be shown.
[ ] Whether or not the descriptions I added as placeholders (e.g. for from_json) are OK.
[ ] Anything else that looks off? 
[ ] `BoundingBoxes2D` and `WBTraceTree` is currently in our live docs, but it is not in `__init__template.pyi` (where/how I make my list for lazydocs to pull APIs from), should they be?


ngrayluna:
Closing in favor of: https://github.com/wandb/docs/pull/1213

PR #1105, opened 2025-02-20
JA automatic translation

tcapelle:
Let's check the files from guides first...


johndmulhausen:
Closing per Slack discussion; let's re-run this when the re-org is done

PR #1113, opened 2025-02-21
Japanese Translation Gemini-2.0-flash

tcapelle:
A friend suggested that I try Gemini for Japanese translation (he lives in Japan) and is surprised with the quality; let's check this one too, then!


johndmulhausen:
Closing per Slack discussion; let's re-run this when the re-org is done

PR #1115, opened 2025-02-21
Tutorials landing page update

ngrayluna:
Another crude, simple, landing page update. This time, the tutorials landing page.


johndmulhausen (changes requested):
Are you having the same issue I am where the outbound links on the bottom don't do anything? https://tutorials-landing-page.docodile.pages.dev/tutorials/

I would convert them into non-cards/regular links since they're not on the docs site, anyway.

johndmulhausen (approved):

Issue #1132, opened 2025-02-28
Trailing </div> tag


johndmulhausen:
URL: https://docs.wandb.ai/guides/sweeps/define-sweep-configuration/

ngrayluna:
Fixed.

PR #1139, opened 2025-03-04
Trailing </div> tag

ngrayluna:
Removes trailing div tag from sweep doc config page. See GitHub Issue: https://github.com/wandb/docs/issues/1132


mdlinville (approved):
I don't understand it but it's fixed!

PR #1145, opened 2025-03-06
Document new reinit='allow' setting

timoffex:
Updates documentation for https://github.com/wandb/wandb/pull/9562.

NOTE: This original example on this page was wrong: `reinit` was not necessary because the example already called `run.finish()`. Setting `reinit=True` simply calls `run.finish()` automatically for the user. I removed mention of `reinit=True` as we would like to deprecate this setting completely.

### DIFF
--- a/content/support/kb-articles/launch_multiple_runs_one_script.md
+++ b/content/support/kb-articles/launch_multiple_runs_one_script.md
@@ -6,29 +6,60 @@ type: docs
 support:
    - experiments
 ---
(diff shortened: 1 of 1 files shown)


ngrayluna (approved):

timoffex:
@ngrayluna I updated this PR to include the most recent information.

I noticed that there Japanese and Korean versions of this page. Do I need to do anything to update them as well?

PR #1148, opened 2025-03-06
Remove highlighting syntax in Hugo docs

ngrayluna:
Removes Docusaurus-specific syntax highlighting from code blocks because:
* Hugo requires a different syntax for code highlighting. 
* Docs Team has not investigated, agreed if highlighting is a good practice (e.g. accessibility)
* Not many docs use it

[Hugo does support code highlights](https://gohugo.io/content-management/syntax-highlighting/), however, it renders incorrectly with current CSS. Will need to update CSS to re-introduce highlighting.


mdlinville (approved):
Yeah surely there is a better way to do this.

PR #1178, opened 2025-03-13
Reapply "Link to commit in last modified date, remove hash""

mdlinville:
Reverts wandb/docs#1177

To reapply this fix. the CSS bug in the footer / right nav needs to be fixed. Let me know how I can be helpful. :)


johndmulhausen:
Can you just revert the css changes altogether and keep the ones in the HTML file? Then I think we can ship this

johndmulhausen (approved):

PR #1192, opened 2025-03-17
Fix: missing escape character

azureology:
Add `\` before `|`  to get correct rendering in markdown table.


mdlinville (approved):
Wow, thank you!

PR #1199, opened 2025-03-19
chore: update references for installing fastcore to plum-dispatch for metaflow integration.

jacobromero:
This change updates instructions for installing `fastcore`, to instead mention installing `plum-dispatch`.

This follows the changes in the SDK which is migrating to `plum-dispath` in https://github.com/wandb/wandb/pull/9599

### Testing

image

image

image


content/guides/integrations/metaflow.md
```diff
@@ -34,6 +34,11 @@ For a more streamlined approach, you can generate an API key by going directly t
 
 To install the `wandb` library locally and log in:
 
+{{% alert %}}
+For wandbkversions <= 0.19.8, install `fastcore<1.8.0` instead of `plum-dispatch`:
```
kptkin:
```suggestion
For wandb versions <= 0.19.8, install `fastcore<1.8.0` instead of `plum-dispatch`:
```

content/guides/integrations/metaflow.md
```diff
@@ -34,6 +34,11 @@ For a more streamlined approach, you can generate an API key by going directly t
 
 To install the `wandb` library locally and log in:
 
+{{% alert %}}
+For wandb versions <= 0.19.8, install `fastcore<1.8.0` instead of `plum-dispatch`:
```
mdlinville: Is there any way to avoid hard-coding all these versions, which will be difficult to maintain? Is there any reason not to have them install exactly 1.8.0, since it seems anything newer than that won't work?

content/guides/integrations/metaflow.md
```diff
@@ -48,7 +53,7 @@ To install the `wandb` library locally and log in:
 
 
     ```shell
-    pip install -Uqqq metaflow fastcore wandb
+    pip install -Uqqq metaflow "plum-dispatch<3.0.0" wandb
```
mdlinville, line 56: Similar confusion about asking for 3.0.0 or earlier? If it's the minimum, why wouldn't we just say to use exactly that version?

mdlinville (commented):
Apologies if I'm misunderstanding something here -- I think we have an opportunity to try to head off the potential for technical debt with the version numbers over time.

content/guides/integrations/metaflow.md
jacobromero: The SDK team is trying to be stricter on our required version to avoid randomly breaking when a dependency updates with a breaking change. 

We are limiting `plum-dispatch` to the latest tested major version. (We have tested with 2.5.7)

content/guides/integrations/metaflow.md
jacobromero: > Is there any reason not to have them install exactly 1.8.0

There was some breaking changes in fastcore 1.8.0, by installing `fastcore<1.8.0` pip will pick up the latest non-breaking change.

We will be moving away from fastcore in the next release of the wandbsdk ([ref](https://github.com/wandb/wandb/pull/9599))

content/guides/integrations/metaflow.md
mdlinville: Got it. Thanks for explaining. I guess I'm wondering if there is any benefit in just specifying the version they should use that we have tested, with the `==` operator: ```pip install -Uqqq metaflow "plum-dispatch==2.5.7" wandb``` That way the customer may not end up getting ahead of our testing. Maybe it's not a big worry.

jacobromero: by specifying `<3.0.0` this stays in line with the version we are installing and using in [our tests](https://github.com/wandb/wandb/blob/main/requirements_dev.txt#L51).

It ensures that we and the user will have any of the latest fixes

content/guides/integrations/metaflow.md
mdlinville:
```suggestion
For `wandb` version 0.19.8 or below, install `fastcore` version 1.8.0 or below (`fastcore<1.8.0`) instead of `plum-dispatch`.
```
This might be more readable in the prose, while also giving the correct syntax. Also, the alert doesn't have any more contents, so we should end the sentence with a period.

mdlinville (approved):
One further suggestion. I believe @kptkin still needs to take a look.

kptkin (approved):

PR #1202, opened 2025-03-19
convert the support section to a hugo taxonomy

hazel-nut:
this PR converts the support section of the w&b site into a taxonomy.

## details

as far as i can tell, the docsy taxonomy treatment has no concept of a section menu for a taxonomy, and the main taxonomy page is mostly just a word cloud.

i couldn't find a customization option in docsy to add a left menu or more docs-style treatment to taxonomy pages, so i wrote it myself: there are two new taxonomy layouts (largely taken directly from the [docsy default layouts](https://github.com/google/docsy/tree/main/layouts/_default) for consistency).

there is also now a slightly new structure in content to support the taxonomy. the actual taxonomy content lives in files in `content/support/terms`, and the rest of the `support` directory is the [typical hugo taxonomy metadata config](https://gohugo.io/templates/taxonomy/#display-metadata).

i also added an [archetype](https://gohugo.io/content-management/archetypes/) for support terms, so this command:

```
hugo new support/terms/example.md
```

creates a new file with the right front matter already added for a new support article.

a couple notable changes:

- the slugs for term pages are now at `/support/<term name>` rather than `/support/index_<term>`. i didn't add aliases but it would be easy to do so.

- the term pages now have a card treatment for support articles. i personally think it's a little more skimmable. plus, if you add a front matter `description` to support articles, they'll display on the cards.

otherwise, the page layouts and slugs should match what currently exists on prod. i tested locally with hugo v0.145.0 extended and node.js v12.18.3.


hazel-nut:
aha, are you on a slightly older hugo version? the build is failing because of `:contentbasename`, which is [available starting in v0.144.0](https://gohugo.io/configuration/permalinks/#contentbasename).

mdlinville:
Wow, thanks for this enhancement! I had a look at the Hugo build failure and it looks like maybe a problem with the placeholder. I couldn't find any reference to `:contentbasename` in the Hugo docs, but I did find this: https://github.com/gohugoio/hugo/blob/52561d561a591a6fb757b6d417138e0a79bce790/docs/content/en/methods/page/File.md?plain=1#L51

hazel-nut:
from `.hugo-version`, it looks like you're on 0.140.2, so i see two fix options:

1. upgrade to 0.145.0 (latest at time of writing), which should involve a good pass for any other breaking or sneaky changes, like the `:filename` deprecation in 0.144.0.

2. just kick the can down the road a bit and change `:contentbasename` in this PR to `:filename`. :slightly_smiling_face: 

(for sake of not blocking my own PR on this, i've implemented option 2 in cce93e3, but naturally option 1 is the right long term approach.)

content/guides/models/app/features/panels/line-plot/reference.md
```diff
@@ -62,7 +62,7 @@ You can aggregate all of the runs by turning on grouping, or group over an indiv
 
 ## Smoothing
 
-You can set the [smoothing coefficient]({{< relref "/support/formula_smoothing_algorithm.md" >}}) to be between 0 and 1 where 0 is no smoothing and 1 is maximum smoothing.
+You can set the [smoothing coefficient]({{< relref "formula_smoothing_algorithm.md" >}}) to be between 0 and 1 where 0 is no smoothing and 1 is maximum smoothing.
```
mdlinville, line 65: Silly change to retrigger Cloudflare (maybe)
```suggestion
You can set the [smoothing coefficient]({{< relref "formula_smoothing_algorithm.md" >}}) to be between 0 and 1 where 0 is no smoothing and 1 is maximum smoothing.

```

hazel-nut: if this is just to retrigger CI, in the future you can make an empty commit, too:

```
git commit --allow-empty -m "empty commit to trigger build"
```

i use this all the time when i'm testing new github actions integrations in my repos. :slightly_smiling_face:

mdlinville: Good tip. Nevertheless, it didn't work. Feel free to pop my commit off, and then perhaps if you rebase on `main` (low risk since it hasn't had review yet), maybe Cloudflare will wake up.

hazel-nut:
@mdlinville, i dropped your commit and force-pushed. i can rebase on main but i wonder if some of the issue might be that my changes are in a fork, not a direct branch of your repo.

mdlinville:
> @mdlinville, i dropped your commit and force-pushed. i can rebase on main but i wonder if some of the issue might be that my changes are in a fork, not a direct branch of your repo.

That's exactly it! I'm always forgetting about that limitation. I do not believe this repo requires use of forks, for what it's worth! But let me try building this locally to see what it looks like.

mdlinville:
Screenshot 2025-03-20 at 1 08 59 PM
Not sure what this is about, but going to `/support` seems to load the last article, the WYSIWYG one, no matter which other one I try to click on.

hazel-nut:
@mdlinville bah, i thought i worked around a frustrating docsy issue with the left menu, but i got duped by hugo's `/public` locally as i was testing.

3875422 should fix it, and this does mean you'll want to stick a `url` in each support article now. ~an [archetype](https://gohugo.io/content-management/archetypes/) might be convenient.~ i've added an archetype in f7a57c7 & updated the PR description.

the issue is that docsy understands section menus only by top-level content directory, which forces the support articles to live in `support/`. this fights with what hugo wants in that directory, which is custom metadata for the support taxonomy terms. (maybe this is why docsy doesn't have a left menu for taxonomies...)

johndmulhausen:
Well the good news is I got deploy previews working for forks. The bad news is I broke the build because I messed up some support page link in `tensorflow-sweeps.md`. All in all this is looking great! I might want to do a pass with the GPT Metadata Generator (still in flight) after we merge this so the `description`s in the front-matter show up in the cards and give you a bit of an article preview. But that's for then. 

As long as the site builds after upgrading to the latest Hugo and we work around the `cascade` of `:filename` in the front-matter (which is how we're currently supporting legacy URLs) there really is no reason not to use a later version/apply the more long-term fix, IMO. Docsy currently specifies a fairly old version in their repo but I think it's just for lack of testing, not because they know they are running afoul of some site-breaking change and haven't addressed it yet.

hazel-nut:
lol. no worries, i believe i've fixed the build and resolved the merge issue @johndmulhausen.

hazel-nut:
on upgrading hugo, i'd recommend splitting that out into a different change, just to keep the scope distinct. this PR will work with the setup you have.

updating to hugo 0.145 will involve a pretty mundane `grep -rl [...] | xargs sed -i [...]` situation to convert the instances of `:filename` to `:contentbasename`, but it's worth making sure there were no other changes in the intervening versions that might rain on your parade, too.

mdlinville:
> Well the good news is I got deploy previews working for forks. The bad news is I broke the build because I messed up some support page link in `tensorflow-sweeps.md`. All in all this is looking great! I might want to do a pass with the GPT Metadata Generator (still in flight) after we merge this so the `description`s in the front-matter show up in the cards and give you a bit of an article preview. But that's for then.
> 

This is awesome. I note that it seems to be only a commit preview, not the PR branch preview, and I'm not sure if it actually updated based on Hazel's last pushes.

hazel-nut:
the current cloudflare comment says it's built on 80cab3c, which is my last commit, so it looks good to me. :slightly_smiling_face:

mdlinville:
> the current cloudflare comment says it's built on [80cab3c](https://github.com/wandb/docs/commit/80cab3cf46176aa88de68914fb1300a9786e98d6), which is my last commit, so it looks good to me. 🙂

Must have been caching, I can confirm the comment looks updated now. I guess the question for @johndmulhausen is did he mean for there to be a branch preview too?

johndmulhausen:
> > the current cloudflare comment says it's built on [80cab3c](https://github.com/wandb/docs/commit/80cab3cf46176aa88de68914fb1300a9786e98d6), which is my last commit, so it looks good to me. 🙂
> 
> Must have been caching, I can confirm the comment looks updated now. I guess the question for @johndmulhausen is did he mean for there to be a branch preview too?

There is a branch preview; it's linked to in the CI Checks area of the PR (where it says **X Checks Have Passed**). For convenience, I should change the generated comment to link to the branch preview, as well as the commit-specific preview. The branch preview is generally the better "shareable" URL, after all, since it will always have the latest. 

> lol. no worries, i believe i've fixed the build and resolved the merge issue @johndmulhausen.

Thank you 🙏

johndmulhausen:
I think this looks great, just a few things:

- It feels a bit strange that the reading time estimates are showing up in the cards in the indexes. Is that Docsy doing that? 
- I think `terms` is a confusing name for the folder that serves as the big bucket full of support articles. Is `kb-articles` too much? 
- Docsy has exposure for tags-based navigation (see the right-hand section of the example site: https://example.docsy.dev/docs/) We override some of the strings in Docsy already so we can definitely replace "Tag Cloud" as the heading. And we can suppress the section for "Categories," the utility of which escapes me. So, if it's fairly free candy to turn on the tags-based navigation (say, just flipping the config switch and it "just works" in /support, the only section that has tags), we might want to preserve the use of the original front-matter annotation for collecting taxonomy terms, which was `tags` rather than the new `support`. I *think* I know why this was changed (because Docsy's native tags browsing is weird and not meant to be used in conjunction w/taxonomies), but curious of your thoughts here @hazel-nut.

hazel-nut:
> It feels a bit strange that the reading time estimates are showing up in the cards in the indexes. Is that Docsy doing that?

yep.

you can hide it with the front matter `hide_readingtime: true`, but that hides it from the header on the page itself in addition to the card (and presumably anywhere else docsy would display it).

presuming you just want it removed from the cards on support taxonomy term pages specifically, i've done that in 49607bd3.

> I think `terms` is a confusing name for the folder that serves as the big bucket full of support articles. Is `kb-articles` too much?

nope, you can call it whatever you want.

i've changed it to `kb-articles` in 6638355a.

hazel-nut:
> Docsy has exposure for tags-based navigation

this is actually taxonomy-based navigation. "tags" is just the name of one of the two default hugo taxonomies — the other of which is "categories".

> we might want to preserve the use of the original front-matter annotation for collecting taxonomy terms

no need. by default, doscy generates taxonomy clouds in the right gutter for all taxonomies on the site.

_(asterisk here. there's a different reason why they weren't showing up before. more on this at the end...)_

> we can suppress the section for "Categories," [...] we can definitely replace "Tag Cloud" as the heading

you _can_ customize which taxonomies are displayed in the right gutter with [some docsy-specific site config params](https://www.docsy.dev/docs/adding-content/taxonomy/#parameters), but there's no taxonomy named "categories" on the site, so therefore nothing to suppress.

one of said site config params, `taxonomyCloudTitle`, also lets you change the title of the right gutter taxonomy cloud. (by default it's based off of the singular of the taxonomy name.)

> if it's fairly free candy to turn on the tags-based navigation

it *is* basically free candy... but i don't think docsy's native taxonomy cloud treatment is what you want.

docsy displays taxonomy terms in two ways: a cloud in the right gutter and in the header of a page (and that page's card treatment).

but docsy's taxonomy clouds don't show only taxonomy terms you've added to a given page in the right gutter. they show _every_ taxonomy term in the right gutter of _every_ page:

image

in contrast, their taxonomy page header treatment displays the taxonomy terms you've added to a page (using its front matter) in that page's header and on its cards, like this:

image

image

i don't think the right gutter taxonomy cloud works the way you want, but i could be wrong.

> (say, just flipping the config switch and it "just works" in /support, the only section that has tags)

_(here's that asterisk)_

so, it *is* as simple as specifying docsy's site config params, but you yourselves hide both the taxonomy cloud and page header treatment at the CSS level!

https://github.com/wandb/docs/blob/main/assets/scss/_variables_project.scss#L188-L189

so:

* i reversed the CSS that hides all the taxonomy page header and cloud treatments in 057cc128 because you should hide those using docsy's site config params instead.

* i took a guess and enabled the taxonomy page header treatment + disabled taxonomy cloud treatment entirely in 2e43b75a.

johndmulhausen:
@hazel-nut I think you nailed it; the right-nav tag cloud approach would look a lot clunkier than the sleek tags you have here. This feels a lot more joyous to use and will absolutely be a lot easier to maintain than the manual updates we were doing with Python. (A necessity in the dark days of Docusaurus.) Also appreciate tidying up the config/CSS that I hacked together to hide the unwanted right-nav fluff. 

Thanks so much for helping the docs site take several steps in a better and more sustainable direction!

johndmulhausen (approved):

PR #1205, opened 2025-03-21
Add CI that builds CloudFlare Pages previews on forks

johndmulhausen:
CloudFlare Pages integration doesn't run on forks. But, because the forks are public, we can do custom CI that manually `git clone`s those forks, and, after checking them out, runs the CloudFlare Pages build for us, commenting back in the originating PR with a link.

See #1206 for an example of this CI in action

### DIFF
--- a/.github/workflows/forks-build-on-cloudflare.yml
+++ b/.github/workflows/forks-build-on-cloudflare.yml
@@ -145,8 +145,9 @@ jobs:
       
       - name: Check Deployment Status
         id: check-status
+        if: always()  # Run this step even if previous steps failed
         run: |
-          if [ "${{ steps.cloudflare-deployment.outcome }}" == "success" ]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/forks-build-on-cloudflare.yml
```diff
@@ -0,0 +1,228 @@
...
+          if [ -d "public" ]; then
+            echo "✅ Hugo generated 'public' directory"
+            echo "Contents of public directory:"
+            ls -la public
+          else
+            echo "❌ Hugo did not generate 'public' directory"
+          fi
+      
+      # Debug directory structure
+      - name: Debug directory structure
```
ngrayluna, line 91: What happens w/out these checks? Curious what errors you were hitting.

ngrayluna (approved):
Left a question (for my understanding). Otherwise looks good.

Issue #1184, opened 2025-03-14
Add line numbers to the code blocks within documentation pages


SinaMostafanejad:
# Summary

The following is an example of a documentation page where there are multiple references in the text to the line numbers in the code blocks but there are no line numbers in the margins of the code blocks for cross-referencing. 

URL: https://docs.wandb.ai/guides/sweeps/add-w-and-b-to-your-code/

ngrayluna:
Ahoy @SinaMostafanejad ! Thanks for flagging. I temporarily removed mentioning line numbers from code snippets in our docs (as opposed to re adding them in) because I discovered some unwanted CSS behaviour that'll need further investigation.

SinaMostafanejad:
> Ahoy [@SinaMostafanejad](https://github.com/SinaMostafanejad) ! Thanks for flagging. I temporarily removed mentioning line numbers from code snippets in our docs (as opposed to re adding them in) because I discovered some unwanted CSS behaviour that'll need further investigation.

Hi @ngrayluna. I guessed the issue may not be as simple as it looked because I could see the ` ```python showLineNumbers``` ` around the code-blocks but apparently something interfered with their actions. OK then, I'll try to remove the references whenever I see them in the docs because I know there are quite a few of them.

ngrayluna:
Thanks @SinaMostafanejad ! Funny enough that syntax is a relic of our old docs platform. We can add line numbers, but there is an odd highlighting happening when introduced.  

I took a pass removing line numbers and mentions of line numbers throughout the docs so (*hopefully*) you shouldn't encounter more of them.

PR #1208, opened 2025-03-21
Remove line numbers

ngrayluna:
Removes line numbers from code blocks (and mention of line numbers). [As discussed here](https://wandb.atlassian.net/browse/DOCS-1354), we can add line numbers but there are weird CSS issues. For now, removing line codes while we resolve that.

GitHub Issue: https://github.com/wandb/docs/issues/1184

### DIFF
--- a/content/guides/core/artifacts/create-a-new-artifact-version.md
+++ b/content/guides/core/artifacts/create-a-new-artifact-version.md
@@ -38,12 +38,12 @@ Based on your use case, select one of the tabs below to create a new artifact ve
   {{% tab header="Inside a run" %}}
 Create an artifact version within a W&B run:
 
-1. Create a run with `wandb.init`. (Line 1)
-2. Create a new artifact or retrieve an existing one with `wandb.Artifact` . (Line 2)
-3. Add files to the artifact with `.add_file`. (Line 9)
-4. Log the artifact to the run with `.log_artifact`. (Line 10)
+1. Create a run with `wandb.init`.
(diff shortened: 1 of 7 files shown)


content/guides/core/artifacts/create-a-new-artifact-version.md
```diff
@@ -38,12 +38,12 @@ Based on your use case, select one of the tabs below to create a new artifact ve
...
-2. Create a new artifact or retrieve an existing one with `wandb.Artifact` . (Line 2)
-3. Add files to the artifact with `.add_file`. (Line 9)
-4. Log the artifact to the run with `.log_artifact`. (Line 10)
+1. Create a run with `wandb.init`.
+2. Create a new artifact or retrieve an existing one with `wandb.Artifact`.
+3. Add files to the artifact with `.add_file`.
+4. Log the artifact to the run with `.log_artifact`.
 
-```python showLineNumbers
+```python
```
mdlinville, line 46: To put them back in the Hugo way, you could instead add this right after the backticks to start the code fence: ```{linenos=true}``` If you also want to highlight some lines, you can add something like ```{hl_lines=[5]}```. If we want to enable line numbers on _all_ code blocks, we could set `linenos` to `true` in the site config.

ngrayluna: Right. I am aware of that. :) I removed them because of the weird CSS happening https://wandb.atlassian.net/browse/DOCS-1354. [Different kind of highlighting]

mdlinville (approved):

PR #1209, opened 2025-03-21
[DOCS-1131] Document run metrics automations

mdlinville:
[DOCS-1131] Document run metrics automations

Ready for peer review

## Previews
Major:
https://run-metrics-automations.docodile.pages.dev/guides/core/automations/automation-events/#run-metrics-events
Minor:
https://run-metrics-automations.docodile.pages.dev/guides/core/automations/
https://run-metrics-automations.docodile.pages.dev/guides/core/automations/create-automations/slack/
https://run-metrics-automations.docodile.pages.dev/guides/core/automations/create-automations/webhook/

[DOCS-1131]: https://wandb.atlassian.net/browse/DOCS-1131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


content/guides/core/automations/automation-events.md
```diff
@@ -29,13 +29,48 @@ You can create a project automation in these scopes:
...
+This section describes the events related to an artifact that can trigger an automation.
 
 - **Linking a new artifact**: Test and validate new models, datasets or dataset automatically.
 - **Creating a new version of an artifact**: Apply recurring actions to each version of an artifact. For example, start a training job when a new dataset artifact version is created.
 - **Adding a new alias to a version of an artifact**: Trigger a special step of your workflow when a new artifact version in a project or collection has a specific label or alias applied. For example, run a series of downstream processing steps when an artifact has the `test-set-quality-check` alias applied.
 
+### Run metrics events
+This section describes the run metrics events that can trigger an automation.
+
+- **Run metrics threshold met**: Trigger a workflow when, for a given metric, a run or the average of a number of runs meets the threshold you specify for a given metric.
```
ssisk:
```suggestion
- **Run metrics threshold met**: Trigger a workflow when, for a given metric, the average of the metric value meets the threshold you specify for a given metric.
```

It's not across runs, it's all within a single run.

content/guides/core/automations/automation-events.md
```diff
@@ -29,13 +29,48 @@ You can create a project automation in these scopes:
...
+- **Run metrics change threshold met**: Trigger a workflow when, for a given metric, a run or the average of a number of runs increases or decreases by the threshold you specify for a given metric.
```
ssisk: ditto above
```suggestion
- **Run metrics change threshold met**: Trigger a workflow when, for a given metric, the average of the metric value increases or decreases by the threshold you specify for a given metric.
```

content/guides/core/automations/automation-events.md
```diff
@@ -29,13 +29,48 @@ You can create a project automation in these scopes:
...
+- The metric.
+- The threshold for evaluation.
+- The comparison to make. Choices differ for each type of event. Refer to [Run metrics comparison choices]({{< relref "#run-metrics-comparison-choices" >}}).
+
+#### Run metrics comparison choices
+When configuring an automation for a run metric event, you can configure how to compare the run metric value with the threshold you specify. Your choices depend on the event type.
+
+For **Run metrics threshold met** events, you can configure:
+1. The number of logged values to average across (defaults to 5).
+1. How to compare the values with the threshold:
```
ssisk: Do we really need to list out the values in the options dropdown? I don't see what users will learn from this. if we needed to explain what they did, we should do that, but I don't think we need to

content/guides/core/automations/automation-events.md
```diff
@@ -29,13 +29,48 @@ You can create a project automation in these scopes:
...
+
+To set up the metric for a run metrics event, you specify:
+
+- An optional run name filter. Only runs matching this filter can trigger the automation.
+- The metric.
+- The threshold for evaluation.
+- The comparison to make. Choices differ for each type of event. Refer to [Run metrics comparison choices]({{< relref "#run-metrics-comparison-choices" >}}).
+
+#### Run metrics comparison choices
+When configuring an automation for a run metric event, you can configure how to compare the run metric value with the threshold you specify. Your choices depend on the event type.
```
ssisk: So this section has a tricky job: it needs to explain how the change threshold metric works. This isn't that complicated, but there are a lot of ways that this *could* be done, and we need to make sure we're very explicit about how it works so that users 

I think it's worth separating out the two types of run metrics (threshold vs change threshold) and explaining the algorithms separately, specifically what values are included in the windows (eg that the current & prior don't overlap and relative ordering), and how the windows are compared (for the change threshold)

Let me know if you'd like to meet to discuss the details of this.

content/guides/core/automations/automation-events.md
```diff
@@ -29,13 +29,48 @@ You can create a project automation in these scopes:
...
+### Run metrics events
+This section describes the run metrics events that can trigger an automation.
+
+- **Run metrics threshold met**: Trigger a workflow when, for a given metric, a run or the average of a number of runs meets the threshold you specify for a given metric.
+- **Run metrics change threshold met**: Trigger a workflow when, for a given metric, a run or the average of a number of runs increases or decreases by the threshold you specify for a given metric.
+
+To set up the metric for a run metrics event, you specify:
+
+- An optional run name filter. Only runs matching this filter can trigger the automation.
+- The metric.
```
ssisk: it's probably worth discussing system metrics and how to get the names of those

mdlinville: Done

content/guides/core/automations/automation-events.md
mdlinville: Done

content/guides/core/automations/automation-events.md
mdlinville: Done

content/guides/core/automations/automation-events.md
mdlinville: Done

content/guides/core/automations/automation-events.md
mdlinville: I think this is done, PTAL!

content/guides/core/automations/_index.md
```diff
@@ -21,6 +21,8 @@ For an artifact in a [Registry]({{< relref "/guides/core/registry/">}}), you can
 For an artifact in a [project]({{< relref "/guides/models/track/project-page.md" >}}), you can configure an automation to run:
 - When a new version is added to an artifact. For example, start a training job when a new version of a dataset artifact is added to a given collection.
 - When an alias is added to an artifact version. For example, trigger a PII redaction workflow when the alias "redaction" is added to a dataset artifact.
+- When a metric for a run or set of runs meets or exceeds a configured threshold.
+- When a metric for a run or set of runs changes by a configured threshold.
```
ssisk: While the alert config does support multiple runs, the alert itself only fires in relation to a particular run - if 3 runs all log the same metric, they are evaluated independently of each other against the alert, not together. I worry that we're confusing users by mentioning set of runs here.

content/guides/core/automations/automation-events.md
```diff
@@ -41,11 +41,43 @@ You can create a project automation at these scopes:
...
 - **An artifact alias is added**: Trigger a special step of your workflow when a new artifact version in a project or collection has a specific alias applied. For example, run a series of downstream processing steps when an artifact has the `test-set-quality-check` alias applied.
 
+### Run metrics events
+An automation can watch for changes in a run metric:
+
+- **Run metrics threshold met**: Trigger a workflow when for a given metric, a single logged value or the average logged values meets the threshold you specify.
+- **Run metrics change threshold met**: Trigger a workflow when for a given metric, a single logged value or the average logged value changes by the threshold you specify.
+
+{{% alert %}}
+You can create aautomations triggered by system metrics such as the `cpu` metric, which tracks the percentage of CPU utilization. W&B logs system metrics automatically every 15 seconds. Find each metric's name and details in [System metrics]({{< relref "/guides/models/app/settings-page/system-metrics.md" >}}).
```
ssisk:
```suggestion
You can create automations triggered by system metrics such as the `cpu` metric, which tracks the percentage of CPU utilization. W&B logs system metrics automatically every 15 seconds. Find each metric's name and details in [System metrics]({{< relref "/guides/models/app/settings-page/system-metrics.md" >}}).
```

ssisk (approved):
one minor change, but I think this is reasonable!

content/guides/core/automations/automation-events.md
```diff
@@ -41,11 +41,43 @@ You can create a project automation at these scopes:
...
+For **Run metrics change threshold met** events, you can configure:
+1. The current window of logged values to average across (defaults to the 10 most recent logged values).
+1. The prior window of logged values to average across (defaults to the 50 most recent logged values prior to the current window).
+1. Whether to evaluate the values as relative or absolute (defaults to **Relative**).
+1. How to compare the values with the threshold:
+      - Increases by at least
+      - Decreases by at least
+      - Increases or decreases by at least
+
+The current and prior window do not overlap. By default, the current window includes the 10 most recent logged values, and the prior window includes the 50 logged values prior to the current window.
```
ssisk: also @mdlinville I realized something I'm working on could affect these docs - not sure if you wanted to change them at all 

https://app.graphite.dev/github/pr/wandb/core/28397/feat(artifacts)-Simplified-alert-creation-UI-for-workspaces is the PR

basically, just in workspaces we have a simplified UI for creating slack notifications when you're making automations on a metric.

content/guides/core/automations/_index.md
mdlinville:
```suggestion
- When a metric for a run meets or exceeds a configured threshold.
- When a metric for a run changes by a configured threshold.

Each run is considered separately, even if it is part of a run set.
```

content/guides/core/automations/automation-events.md
mdlinville: I'm having trouble accessing that. If it is not already available, we should address it as a fast follow. These docs are intended to be published very very soon.

content/guides/core/automations/_index.md
mdlinville: I think I addressed this

tristan-spaulding2 (commented):
Looks good, except we need to note that this available only to Pro or Enterprise edition users on W&B Multi-tenant SaaS.  Unless I missed that.

content/guides/core/automations/_index.md
```diff
@@ -21,6 +21,10 @@ For an artifact in a [Registry]({{< relref "/guides/core/registry/">}}), you can
 For an artifact in a [project]({{< relref "/guides/models/track/project-page.md" >}}), you can configure an automation to run:
 - When a new version is added to an artifact. For example, start a training job when a new version of a dataset artifact is added to a given collection.
 - When an alias is added to an artifact version. For example, trigger a PII redaction workflow when the alias "redaction" is added to a dataset artifact.
+- When a metric for a run meets or exceeds a configured threshold.
+- When a metric for a run changes by a configured threshold.
+
+Each run is considered separately, even if it is part of a run set.
```
ssisk: Note that the term Run Set is meaningful in weights and biases - https://docs.wandb.ai/guides/reports/edit-a-report/#add-run-sets, and while this sentence isn't technically wrong, it's confusing to someone who knows what runsets are.

I think the point is important, but it's a detail we should move it down into the automation-events doc since it's a bit of an implementation detail. 

I would probably write it something like: 
`Each run is considered separately, even if the automation has a run filter that will match with multiple runs. The values for each run are put into their own window, and thus each run is compared to the threshold separately. If the automation fires for one run, it can still fire for other runs that match that automation.  A particular run metric automation will only fire once every 24hrs per run.`

mdlinville: Done!

content/guides/core/automations/automation-events.md
```diff
@@ -45,10 +45,46 @@ You can create a project automation at these scopes:
...
+- **Adding a new alias to a version of an artifact**: Trigger a special step of your workflow when a new artifact version in a project or collection has a specific label or alias applied. For example, run a series of downstream processing steps when an artifact has the `test-set-quality-check` alias applied.
+
+### Run metrics events
+An automation can watch for changes in a run metric:
+
+- **Run metrics threshold met**: Trigger a workflow when for a given metric, a single logged value or the average logged values meets the threshold you specify.
+- **Run metrics change threshold met**: Trigger a workflow when for a given metric, a single logged value or the average logged value changes by the threshold you specify.
+
+Each run is considered separately, even if the automation has a run filter that matches with multiple runs. Each run's values are put into a separate window and compared to the threshold separately. If an automation starts due to the result of one run, it can still start for another matching run. A particular run metric automation runs only once per run per 24-hour period.
```
ssisk: It feels a bit random having this here - we have sections for threshold, does it make sense to add a section for the run filter, and put this in it?

This would be fine to do in a follow on change since I know you're under time pressure.

content/guides/core/automations/automation-events.md
```diff
@@ -45,10 +45,46 @@ You can create a project automation at these scopes:
...
+
+{{% alert %}}
+You can create automations triggered by system metrics such as the `cpu` metric, which tracks the percentage of CPU utilization. W&B logs system metrics automatically every 15 seconds. Find each metric's name and details in [System metrics]({{< relref "/guides/models/app/settings-page/system-metrics.md" >}}).
```
ssisk: ditto that this feels random put in here - I don't see why it needs to have an alert. If we created a 'metric name' section, we could put this in there

This would be fine to do in a follow on change since I know you're under time pressure.

ssisk (approved):
2 non-blocking comments that could be follow on PRs

content/guides/core/automations/automation-events.md
mdlinville: Good idea, PTAL!

content/guides/core/automations/automation-events.md
mdlinville: I made an attempt -- PTAL!

content/guides/core/automations/create-automations/webhook.md
```diff
@@ -66,7 +66,7 @@ A W&B admin can create automations in a project.
 1. Log in to W&B and go to the project page.
 1. In the sidebar, click **Automations**.
 1. Click **Create automation**.
-1. Choose the [**Event**]({{< relref "/guides/core/automations/automation-events.md" >}}) to watch for.
+1. Choose the [**Event**]({{< relref "/guides/core/automations/automation-events.md" >}}) to watch for, such as when an artifact alias is added or when a run metric meets a given threshold.
```
mdlinville:
```suggestion
1. Choose the [event]({{< relref "/guides/core/automations/automation-events.md" >}}) to watch for, such as when an artifact alias is added or when a run metric meets a given threshold.
```

mdlinville:
> Looks good, except we need to note that this available only to Pro or Enterprise edition users on W&B Multi-tenant SaaS. Unless I missed that.

Done, in addition to the feedback you gave privately.

content/guides/core/automations/_index.md
```diff
@@ -35,6 +37,9 @@ To [create an automation]({{< relref "create-automations/" >}}), you:
     1. Define the [event]({{< relref "#automation-events" >}}) to watch for, such as when a new artifact version is added. 
     1. Define the action to take when the event occurs (posting to a Slack channel or running a webhook). For a webhook, specify a secret to use for the access token and/or a secret to send with the payload, if required.
 
+## Limitations
+[Run metrics automations]({{< relref "automation-events.md#run-metrics-events">}}) are currently supported only in [W&B Multi-tenant Cloud]({{< relref "/guides/hosting/#wb-multi-tenant-cloud" >}}).
```
ssisk, line 41: and we should also call out that it's a paid feature, so Pro/enterprise only  - this only calls out multi-tenant cloud

mdlinville: That's already covered by the prior Automations PR, up at the top of each page. It's not so obvious from the source here, but should be showing in the HTML previews.

content/guides/core/automations/automation-events.md
```diff
@@ -10,7 +10,7 @@ weight: 2
 {{< readfile file="/_includes/enterprise-cloud-only.md" >}}
 {{% /pageinfo %}}
 
-An automation can start when a specific event occurs within a project's or registrie's  scope. The *scope* of a project refers to [INSERT tech def of scope]. This page describes the events that can trigger an automation within each scope.
+An automation can start when a specific event occurs within a project's or registry. This page describes the events that can trigger an automation within each scope.
```
ssisk:
```suggestion
An automation can start when a specific event occurs within a project or registry. This page describes the events that can trigger an automation within each scope.
```

content/guides/core/automations/_index.md
ssisk: ah, great, thanks!

content/guides/core/automations/automation-events.md
```diff
@@ -86,6 +84,14 @@ For **Run metrics change threshold met** events, you can configure:
 
 The current and prior window do not overlap. By default, the current window includes the 10 most recent logged values, and the prior window includes the 50 logged values prior to the current window.
 
+#### Run filters
+This section describes how the automation selects runs to evaluate.
+
+- Each run is considered separately, and the automation can potentially start due to the result of multiple runs.
+- Each run's values are put into a separate window and compared to the threshold separately.
+- If you specify a run filter, only matching runs can start the automation.
+- A particular run metric automation runs at maximum once per run per 24-hour period.
```
ssisk: I think this is confusing because `run` is ambiguous, and eg could imply "a particular run metric can only evaluate metric values once per day".  We use the terms "fire" or "trigger" to refer to what happens when we execute the action that the user  has specified to avoid the confusion.

content/guides/core/automations/automation-events.md
```diff
@@ -86,6 +84,14 @@ For **Run metrics change threshold met** events, you can configure:
 
 The current and prior window do not overlap. By default, the current window includes the 10 most recent logged values, and the prior window includes the 50 logged values prior to the current window.
 
+#### Run filters
+This section describes how the automation selects runs to evaluate.
+
+- Each run is considered separately, and the automation can potentially start due to the result of multiple runs.
```
ssisk: I find this sentence a bit confusing and a bit repetitive with lines 91 and 92 - I wonder if we can just drop this line and move line 92 first, maybe tweaking it to say something like "if you specify a run filter, all matching run will be evaluated by the automation. If no run filter is specified all runs in a project will be evaluated."

ssisk (approved):

content/guides/core/automations/automation-events.md
```diff
@@ -45,10 +45,56 @@ You can create a project automation at these scopes:
...
 - Collection level: All collections in the project that match the filter you specify.
 
-### Events
-A project automation can watch for these events:
-- **A new version of an artifact is created in a collection**: Apply recurring actions to each version of an artifact. Specifying a collection is optional. For example, start a training job when a new dataset artifact version is created.
-- **An artifact alias is added**: Trigger a specific step of your workflow when a new artifact version in a project or collection has a specific alias applied. For example, run a series of downstream processing steps when an artifact has the `test-set-quality-check` alias applied.
+### Artifact events
+This section describes the events related to an artifact that can trigger an automation.
+
+- **Linking a new artifact**: Test and validate new models, datasets or dataset automatically.
```
ngrayluna:
```suggestion
- **Linking a new artifact**: Test and validate new models or datasets automatically.
```

content/guides/core/automations/automation-events.md
```diff
@@ -45,10 +45,56 @@ You can create a project automation at these scopes:
...
+- **Creating a new version of an artifact**: Apply recurring actions to each version of an artifact. For example, start a training job when a new dataset artifact version is created.
+- **Adding a new alias to a version of an artifact**: Trigger a special step of your workflow when a new artifact version in a project or collection has a specific label or alias applied. For example, run a series of downstream processing steps when an artifact has the `test-set-quality-check` alias applied.
```
ngrayluna:
```suggestion
- **Adding a new alias to a version of an artifact**: Trigger a specific step of your workflow when a new artifact version in a project or collection has a specific label or alias applied. For example, run a series of downstream processing steps when an artifact has the `test-set-quality-check` alias applied.
```

content/guides/core/automations/automation-events.md
```diff
@@ -45,10 +45,56 @@ You can create a project automation at these scopes:
...
+      - Decreases by at least
+      - Increases or decreases by at least
+
+The current and prior window do not overlap.
+
+#### Run filters
+This section describes how the automation selects runs to evaluate.
+
+- Each run is considered separately, and the automation can potentially start due to the result of multiple runs.
+- Each run's values are put into a separate window and compared to the threshold separately.
```
ngrayluna, line 102: What is meant by "windows"? Are windows like bins when it comes to histograms?

content/guides/core/automations/automation-events.md
```diff
@@ -45,10 +45,56 @@ You can create a project automation at these scopes:
...
+      - Increases by at least
+      - Decreases by at least
+      - Increases or decreases by at least
+
+The current and prior window do not overlap.
+
+#### Run filters
+This section describes how the automation selects runs to evaluate.
+
+- Each run is considered separately, and the automation can potentially start due to the result of multiple runs.
```
mdlinville:
```suggestion
- Each run is considered separately and can potentially trigger the automation.
```

content/guides/core/automations/automation-events.md
mdlinville: Not quite! Because this is about an amount of change, there is a window for "current" runs and a window for "historical" runs. The two windows run consecutively but don't overlap. By default, the most recent 10 runs are averaged and compared with the next most recent 50 runs. Do I have it right, @ssisk ?

content/guides/core/automations/automation-events.md
```diff
@@ -45,10 +45,56 @@ You can create a project automation at these scopes:
...
+- Each run's values are put into a separate window and compared to the threshold separately.
+- If you specify a run filter, only matching runs can start the automation.
+- A particular run metric automation runs at maximum once per run per 24-hour period.
```
mdlinville:
```suggestion
- In a 24 hour period, a particular automation can fire at most once per run.
```

content/guides/core/automations/automation-events.md
mdlinville: PTAL to see if this is better.

content/guides/core/automations/automation-events.md
mdlinville: PTAL

content/guides/core/automations/automation-events.md
ngrayluna: Ah. I see. Perhaps had a quick, short sentence that summarizes how windows work (and why it's important) before listing the window options (i.e. before the list)

ssisk: yes, matt's description is correct

content/guides/core/automations/automation-events.md
ssisk: thanks, much clearer! I assume that you're thinking that users will already be assuming that run metric automations without a run filter will match with all runs in the project?

tristan-spaulding2 (approved):

content/guides/core/automations/automation-events.md
mdlinville: Restructured a bit in the latest iteration, thank you both for helping me get this right.

content/guides/core/automations/automation-events.md
mdlinville: Nope, I would have missed this. Addressed.

PR #1221, opened 2025-03-27
[DOCS-1326] Fix support, tutorial, ref cards so their contents is not all one big <a> tag

mdlinville:
[DOCS-1326] Fix support, tutorial, ref cards so their contents is not all one big <a> tag

The cards in the top-level index are set up differently and I was not able to simplify them in the same way, so I skipped them in this change.

[DOCS-1326]: https://wandb.atlassian.net/browse/DOCS-1326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


johndmulhausen (changes requested):
@mdlinville Can you do this for `/ref/_index.md` as well? 🙏

PR #1229, opened 2025-04-02
Ko v3 refresh

tcapelle:
Korean V3 translation


johndmulhausen:
I'm going to pull this into the `translation-refresh` branch so we can assemble the big picture here. Thanks, @tcapelle !

PR #1232, opened 2025-04-02
[DOCS-1254] Update docs for Run Comparer

mdlinville:
[DOCS-1254] Update docs for Run Comparer

- Give more details about choosing what to compare
- Give more details about what Diff Only is about
- You can now filter values by key
- You can now copy any value
- You can now adjust the display density
- Remove the outdated image

Ready for technical review.

Preview: https://docs-1254.docodile.pages.dev/guides/app/features/panels/run-comparer/

[DOCS-1254]: https://wandb.atlassian.net/browse/DOCS-1254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ericakdiaz:
updated content looks good to me! i clicked on the preview link though, and saw the old gif, and couldn't see the new image - is that expected? i haven't used this docs preview before 😅 
image

ericakdiaz (approved):

mdlinville:
> updated content looks good to me! i clicked on the preview link though, and saw the old gif, and couldn't see the new image - is that expected? i haven't used this docs preview before 😅 image

I believe I may have fixed this now. What is happening is that in the preview, it is using `docs.wandb.ai` as the base URL for the image, instead of `/` inside the PR. (this would not look broken if the image already existed in production) We have noticed this before but haven't fixed it. Fingers crossed.

ericakdiaz (approved):
looks good now! thanks for fixing :)

mdlinville:
@johndmulhausen Can you help figure out why the image still is not working in the preview? It is pointing to `docs.wandb.ai` instead of https://docs-1254.docodile.pages.dev/images/general/run-comparer.png (which does work). It just means that the PR preview doesn't show the actual image. I tried updating the baseURL, which you suggested last time we ran into this, but doesn't seem to be working here.

content/guides/models/app/features/panels/run-comparer.md
```diff
@@ -8,17 +8,26 @@ title: Compare run metrics
...
+Run Comparer shows the configuration and logged metrics for the 10 first visible runs in the project, one column per run.
 
+- To change the runs to compare, you can search, filter, group, or sort the list of runs on the left-hand side. The Run Comparer updates automatically.
+- To filter or search for a configuration key, use the search field at the top of the Run Comparer.
+- To quickly see differences and hide identical values, toggle **Diff only** at the top of the panel.
+- To adjust the column width or row height, use the formatting buttons at the top of the panel.
+- To copy any configuration or metric's value, hover your mouse over the value, then click the copy button. The entire value is copied, even if it is too long to display on the screen.
 
+{{% alert %}}
+Run Comparer does not differentiate runs by type. This means that it is possible to compare runs that are not really comparable, like comparing an image run to an audio run. Search, filter, group, or sort the list of runs to limit it to the runs you want to analyze.
```
ngrayluna: Runs don't have "types". You can, however, specify a "job_type" when you create a run. See "job_type" in https://docs.wandb.ai/ref/python/init/

content/guides/models/app/features/panels/run-comparer.md
```diff
@@ -8,17 +8,26 @@ title: Compare run metrics
...
+{{% /alert %}}
 
-Toggle the **diff only** option to hide rows where the values are the same across runs.​​
+This image compares three image runs with **Diff only** enabled, and filters for configurations that match the string `run`.
```
ngrayluna: I don't think we should show this image, example. I think this confuses run names with filtering runs by what they do.  For a couple of reasons:

* Run names are normally used for describing the step in the ML. E.g. "training-image
* Normally you wouldn't create runs with the same name. At a minimum, you'd append the run.id to differentiate between runs. 
* You can sort runs based on the "job_type" specified when you create a run.

hugo.yaml
```diff
@@ -1,4 +1,4 @@
-baseURL: https://docs.wandb.ai
```
ngrayluna: Please move this to a separate PR for review.

scripts/production.sh
```diff
@@ -3,7 +3,7 @@
 # Build the English docs from this branch
 hugo mod get -u
 rm -rf public
-hugo
+hugo -b $CF_PAGES_URL
```
ngrayluna: Please move this to a separate PR for review.

ngrayluna (changes requested):

hugo.yaml
mdlinville: This was in this PR to work around the broken image in the PR preview. If I were not removing the image based on your feedback, it would make sense to keep it together.

scripts/production.sh
mdlinville: Ditto here, I'll make that change since I am also removing the image here.

hugo.yaml
mdlinville: Moved to https://github.com/wandb/docs/pull/1244

scripts/production.sh
mdlinville: Moved to https://github.com/wandb/docs/pull/1244

content/guides/models/app/features/panels/run-comparer.md
```diff
@@ -8,17 +8,22 @@ title: Compare run metrics
...
+- To change the runs to compare, you can search, filter, group, or sort the list of runs on the left-hand side. The Run Comparer updates automatically.
+- To filter or search for a configuration key, use the search field at the top of the Run Comparer.
+- To quickly see differences and hide identical values, toggle **Diff only** at the top of the panel.
+- To adjust the column width or row height, use the formatting buttons at the top of the panel.
+- To copy any configuration or metric's value, hover your mouse over the value, then click the copy button. The entire value is copied, even if it is too long to display on the screen.
 
-{{< img src="/images/general/run-comparer-2.gif" alt="" >}}
\ No newline at end of file
+{{% alert %}}
+Run Comparer does not differentiate runs with different values for [`job_type`]({{< relref "/ref/python/init.md" >}}). This means that it is possible to compare runs that are not really comparable, like comparing an image run to an audio run. Search, filter, group, or sort the list of runs to limit it to the runs you want to analyze. For example, filter or sort the list of runs by `job_type` to compare similar runs.
```
ngrayluna: Feel free to word smith:

```suggestion
By default, Run Comparer does not differentiate runs with different values for [`job_type`]({{< relref "/ref/python/init.md" >}}). This means that it is possible to compare runs that are not comparable within a project. For example, you could compare a training run to a model evaluation run. A training run could contain run logs, hyperparameters, training loss metrics, and the model itself. An evaluation run could use the model to check the model's performance on new training data.

When you search, filter, group, or sort the list of runs in the Runs Table, the Run Comparer automatically updates to compare the first 10 runs. Filter or search within the Runs Table to compare similar runs, such as by filtering or sorting the list by `job_type`. Learn more about [filtering runs]({{< relref "/guides/runs/filter-runs.md" >}}).
```

Link to filtering runs: https://docs.wandb.ai/guides/runs/filter-runs/

mdlinville: WDYT about my edit of your suggestion?

ngrayluna: Add "could" because, a user may decide to not do all the above and/or do more.

"A training run could contain run logs, hyperparameters, training loss metrics, and the model itself. While an evaluation run could use the model to check the model's performance on new training data."

content/guides/models/app/features/panels/run-comparer.md
```diff
@@ -8,17 +8,24 @@ title: Compare run metrics
...
 
-Toggle the **diff only** option to hide rows where the values are the same across runs.​​
+{{% alert %}}
+By default, Run Comparer does not differentiate runs with different values for [`job_type`]({{< relref "/ref/python/init.md" >}}). This means that it is possible to compare runs that are not comparable within a project. For example, you could compare a training run to a model evaluation run. A training run could contain run logs, hyperparameters, training loss metrics, and the model itself. An evaluation run could use the model to check the model's performance on new training data.
 
-{{< img src="/images/general/run-comparer-2.gif" alt="" >}}
\ No newline at end of file
+When you search, filter, group, or sort the list of runs in the Runs Table, the Run Comparer automatically updates to compare the first 10 runs. Filter or search within the Runs Table to compare similar runs, such as by filtering or sorting the list by `job_type`. Learn more about [filtering runs]({{< relref "/guides/runs/filter-runs.md" >}}).
```
mdlinville:
```suggestion
When you search, filter, group, or sort the list of runs in the Runs Table, the Run Comparer automatically updates to compare the first 10 runs. Filter or search within the Runs Table to compare similar runs, such as by filtering or sorting the list by `job_type`. Learn more about [filtering runs]({{< relref "/guides/models/track/runs/filter-runs.md" >}}).
```

ngrayluna (approved):

PR #1240, opened 2025-04-08
Create new guide for multiple-runs-per-process

noaleetz:
New "reinit" settings allow users to use multiple runs in a single active process. 

This is based on this ticket - https://wandb.atlassian.net/browse/WB-23108
This is based on this PR - [feat: reinit="allow" setting for multiple simultaneous runs in one process](https://github.com/wandb/wandb/pull/9562)


content/guides/models/track/runs/multiple-runs-per-process
```diff
@@ -0,0 +1,147 @@
+# Multiple Active Runs in One Process (Reinit Behavior)
```
timoffex: I believe we use "Sentence case" for all titles and headings.

How about,

> Simultaneous runs using `reinit="create_new"`

The title is shorter and informative enough that users who know the setting exists, but don't remember it, won't even have to click this entry in the navigation sidebar.

timoffex:
I can't find this page in the preview. Based on its path, I would expect it to get shown under "W&B Models > Experiments > What are runs?" in the Guides sidebar. Take a look at the other files in the same directory, you might need some additional metadata at the top of the file.

content/guides/models/track/runs/multiple-runs-per-process
```diff
@@ -0,0 +1,147 @@
...
+
+**Quick Reference**  
+- **`return_previous`**: Return the existing run; no new run is created.  
+- **`finish_previous`**: Finish (end) the old run, then create a new one.
```
timoffex: `(end)` is redundant and could be removed. I think the docs use "finish" consistently enough for users to know what it means, and it corresponds directly to the `run.finish()` method.

content/guides/models/track/runs/multiple-runs-per-process
```diff
@@ -0,0 +1,147 @@
+# Multiple Active Runs in One Process (Reinit Behavior)
+
+**Quick Reference**  
+- **`return_previous`**: Return the existing run; no new run is created.
```
timoffex: If there are already multiple runs, this returns the most recently created one that isn't finished.

See https://github.com/wandb/wandb/blob/0d5b23a564812dfd159edf3210d5a0aeda85df08/wandb/sdk/wandb_settings.py#L374-L399

content/guides/models/track/runs/multiple-runs-per-process
timoffex: If there are multiple runs, they're all finished. Maybe update to

> Finish all previous runs, then create a new one.

content/guides/models/track/runs/multiple-runs-per-process
```diff
@@ -0,0 +1,147 @@
...
+- Orchestrating sub-experiments in a single file.  
+- Logging from one “master” process to several runs that represent different tasks or time periods.
+
+W&B provides a **`reinit`** setting to control how `wandb.init()` behaves when there’s an existing active run. Below, we detail each `reinit` option and show how to maintain multiple concurrent runs—finishing them at different times without interference.
+
+---
+
+## Overview of `reinit` Options
+
+Use the `reinit` argument when calling `wandb.init()`, or set it via environment variable (e.g., `WANDB__INIT__REINIT`). The valid options:
```
timoffex: The environment variable is wrong. It is just `WANDB_REINIT`.

I think it's important to call out that `reinit` is a setting, and any way of configuring settings works.

> Set the `reinit` setting in any of the following ways:
>
> * Passing it as the `reinit` argument to `wandb.init()` or via the `settings` argument
> * Passing it via `settings` to `wandb.setup()` (applies to all `wandb.init()` calls)
> * Setting the `WANDB_REINIT` environment variable (applies to all `wandb.init()` calls)

There is another way, which is to set it in the config file, but I couldn't find any documentation about that to link out to.

noaleetz:
> I can't find this page in the preview. Based on its path, I would expect it to get shown under "W&B Models > Experiments > What are runs?" in the Guides sidebar. Take a look at the other files in the same directory, you might need some additional metadata at the top of the file.

Hey @ngrayluna can you take a look at this PR re any things that have to be done on the structure side of things? My goal was to get all the correct content in there to streamline the docs writing process

content/guides/models/track/runs/multiple-runs-per-process
```diff
@@ -0,0 +1,147 @@
...
+
+1. **`return_previous`**  
+   - If another run is active, return *that same* run.  
+   - *No new run is created.* This is effectively a “no-op” and is the legacy default.
```
timoffex:
```suggestion
   - *No new run is created.* This is effectively a no-op and is the legacy default when not running in a Jupyter/IPython notebook.
```

content/guides/models/track/runs/multiple-runs-per-process
```diff
@@ -0,0 +1,147 @@
...
+
+2. **`finish_previous`**  
+   - *Finish* (calls `wandb.finish()`) on any active run(s) before creating a new one.  
+   - Handy if you want a clear break between sequential sub-runs.
```
timoffex: Don't mention `wandb.finish()`, we are trying to deprecate global run functions.

```suggestion
   - *Finish* (calls `run.finish()`) on any active run(s) before creating a new one.  
   - Handy if you want a clear break between sequential sub-runs.
   - This is the default when running in a Jupyter/IPython notebook.
```

content/guides/models/track/runs/multiple-runs-per-process
```diff
@@ -0,0 +1,147 @@
...
+import wandb
+
+def train(name: str) -> None:
+    """
+    Perform one training iteration in its own W&B run.
+
+    Using a 'with wandb.init()' block with `reinit="create_new"` ensures that
+    this training sub-run can be created even if another run (like our primary
+    tracking run) is already active.
+    """
```
timoffex: The docstring style should ideally be consistent with the one we use in the SDK codebase. The summary line is attached to the opening triple-quote and separated from the rest of the description by a blank line.

```suggestion
    """Perform one training iteration in its own W&B run.

    Using a 'with wandb.init()' block with `reinit="create_new"` ensures that
    this training sub-run can be created even if another run (like our primary
    tracking run) is already active.
    """
```

content/guides/models/track/runs/multiple-runs-per-process
```diff
@@ -0,0 +1,147 @@
...
+    ) as run:
+        # Pretend there's a training loop here logging 'train_loss'.
+        # In a real script, you'd run your training steps inside this block.
+        run.log({"train_loss": 0.42})  # Replace with your real metric(s)
+
+def evaluate_loss_accuracy() -> (float, float):
+    """
+    Returns the current model's loss and accuracy.
+    Replace this placeholder with your real evaluation logic.
+    """
```
timoffex: Likewise here

```suggestion
    """Returns the current model's loss and accuracy.
    
    Replace this placeholder with your real evaluation logic.
    """
```

content/guides/models/track/runs/multiple-runs-per-process
```diff
@@ -0,0 +1,147 @@
...
+    tracking_run.log({"eval_loss": loss, "eval_accuracy": accuracy})
+    
+    # The 'tracking_run' finishes automatically when this 'with' block ends.
+
+---
+
+### Key Takeaways
+
+1. **`reinit="create_new"`**  
+   Forces a new run each time you call `wandb.init()`.
```
timoffex: The word `forces` implies some struggle, like the natural state for `wandb.init()` isn't to create a new run. I would soften this to "Creates a new run ...".

content/guides/models/track/runs/multiple-runs-per-process
```diff
@@ -0,0 +1,147 @@
...
+
+2. **Keep references to each run**  
+   Since `wandb.run` won’t automatically point to the newest run in `create_new` mode, store them in variables like `run_a`, `run_b1`, etc., and call `.log()` or `.finish()` on those objects.
+
+3. **Order of finishing**  
+   You can finish sub-runs whenever you want while keeping the primary run open until the very end.
+
+4.  **Always Finish Your Runs**
```
timoffex: Sentence case

content/guides/models/track/runs/multiple-runs-per-process
```diff
@@ -0,0 +1,147 @@
...
+   Forces a new run each time you call `wandb.init()`.
+
+2. **Keep references to each run**  
+   Since `wandb.run` won’t automatically point to the newest run in `create_new` mode, store them in variables like `run_a`, `run_b1`, etc., and call `.log()` or `.finish()` on those objects.
+
+3. **Order of finishing**  
+   You can finish sub-runs whenever you want while keeping the primary run open until the very end.
+
+4.  **Always Finish Your Runs**  
+    If you don’t explicitly call `run.finish()`, your script exit may be slower, or your data may not be fully uploaded.
```
timoffex: Another reason it is important is that `create_new` will eventually become the default (though likely not soon since `wandb.run` is so embedded), and scripts that don't finish their runs explicitly will not work correctly.

timoffex (commented):
Generally looks good to me! Just some wording and formatting suggestions.

content/guides/models/track/runs/multiple-runs-per-process
mdlinville: This page won't show up until you:
- Rename the file to `multiple-runs-per-process.md` (currently missing the extension)
- Add some front-matter in YAML format. Copy it from a sibling page. Look for the `---` lines. It shows up under the `parent` you specify. See https://github.com/wandb/docs/blob/main/content/guides/models/track/runs/alert.md?plain=1#L6.

To influence how the pages in a section are sorted, you can use `weight`. It defaults to `1` if it's not set. You can leave space between `weight` values to ease adding more pages in the future.

mdlinville:
> > I can't find this page in the preview. Based on its path, I would expect it to get shown under "W&B Models > Experiments > What are runs?" in the Guides sidebar. Take a look at the other files in the same directory, you might need some additional metadata at the top of the file.
> 
> Hey @ngrayluna can you take a look at this PR re any things that have to be done on the structure side of things? My goal was to get all the correct content in there to streamline the docs writing process

https://github.com/wandb/docs/pull/1240/files#r2053153855 I explain what needs to happen for the page to show up in the TOC. :)

ngrayluna:
Thanks @mdlinville for taking a look. I'll go ahead and push some edits in to get this closer to over the line.

ngrayluna:
Closing this PR in favor of this one: https://github.com/wandb/docs/pull/1279

PR #1244, opened 2025-04-08
Fix broken images in Cloudflare preview builds

mdlinville:
Moved from #1232 

This change updates production.sh to call hugo
differently in Cloudflare environments than otherwise.
- In Cloudflare, the variable $CF_PAGES_URL tracks the
  base URL of a PR build or the production deploy and needs
  to be part of the hugo command.
- Locally, this variable is not set so don't try to use it.

To help with debugging, log the environment we are building in.

## Tests
- [x] Local build works, image shows up correctly
- [x] PR build works, image shows up correctly


hugo.yaml
```diff
@@ -1,4 +1,4 @@
-baseURL: https://docs.wandb.ai
+baseURL: /
```
johndmulhausen: When running locally, this makes it so that the link provided in the terminal when doing `hugo serve` does not work, as it removes the `https://` protocol from the string. Must we do it this way?

mdlinville: No, I can work around this. Thanks for noticing it. I'll ping you when it's ready for another look.

ngrayluna (approved):
Approved. Don't forget to remove rocket image first!

PR #1279, opened 2025-04-25
Experiments: Multiple runs in one process

ngrayluna:
## Docs preview:
https://noah-multiple-runs-one-proce.docodile.pages.dev/guides/runs/multiple-runs-per-process/

## Description
Adds doc on `reinint` with emphasis on multi-process tracking. This is a rewrite of what Noa S. proposed: https://github.com/wandb/docs/pull/1240/files. This PR condenses that PR (redundant info) and adds some additional context when deemed necessary. 

Note: This doc is neither here nor there...I definitely feel it can be improved. E.g. 

* We only show code example for one of the options. 
* Why do we mention `return_previous` since it's not part of the multi process run workflow we are highlighting?
* Similarly, we mention `finish_previous` but never show how to use it.
* etc.

## Ticket
Jira: https://wandb.atlassian.net/browse/DOCS-866


content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -0,0 +1,168 @@
+---
+description: Manage multiple runs in a single Python process using W&B’s `reinit` functionality
```
mdlinville: I don't think Markdown is supported in the front matter.

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -0,0 +1,168 @@
...
+title: Create and manage multiple runs in a single process
+---
+
+Manage multiple runs in a single Python process. This is useful for workflows where you want to keep a primary process active while creating short-lived secondary processes for sub-tasks. Some use cases include:
+
+- Keeping a single “primary” run active throughout a script while spinning up short-lived “secondary” runs for evaluations or sub-tasks.  
+- Orchestrating sub-experiments in a single file.  
+- Logging from one “main” process to several runs that represent different tasks or time periods.
+
+By default, W&B assumes each Python process has only one active run at a time when you call `wandb.init()`. If you call `wandb.init()` again, W&B will either return the same run or finish the old run before starting a new one—depending on the configuration. The content in this guide explains how to use `reinit` to modify the `wandb.init()` behavior to enable multiple runs in a single Python process.
```
mdlinville:
```suggestion
By default, W&B assumes each Python process has only one active run at a time when you call `wandb.init()`. If you call `wandb.init()` again, W&B will either return the same run or finish the old run before starting a new one, depending on the configuration. The content in this guide explains how to use `reinit` to modify the `wandb.init()` behavior to enable multiple runs in a single Python process.
```

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+## `reinit` options
+
+Use the `reinit` parameter to determine how W&B handles multiple calls to `wandb.init()`. `reinit` accepts:
```
mdlinville:
```suggestion
Use the `reinit` parameter to configure how W&B handles multiple calls to `wandb.init()`. Valid values are:
```

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -0,0 +1,168 @@
...
+
+- `create_new`: Create a new run (`wandb.init()`) without finishing existing, active runs.
```
mdlinville: This would be easier to read as a table, I think, or as H3 headings where you can give the specifics about each one. There is some beefy stuff here that might get missed in the current presentation.

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -0,0 +1,168 @@
...
+   * W&B does not automatically switch the global `wandb.run` to new runs. You must hold onto each run object yourself. See the [example]({{< relref "multiple-runs-per-process/#example-multiple-runs-in-one-process" >}}) below for details. 
+- `finish_previous`: Finish all active runs with `run.finish()` before creating a new one run with `wandb.init()`. 
+   * Ideal when you want to break sequential sub-processes into separate individual runs.
+   * Default behaviour for non notebook environments.
+- `return_previous`: Return the most recent, unfinished run.
+   * This option does not create a new run.
+   * Default behaviour for notebook environments.
+
+{{% alert  %}}
+W&B does not support `create_new` mode for integrations that assume a single global run, such as Hugging Face Trainer, Keras callbacks, and PyTorch Lightning. If you use these integrations, you should run each sub-experiment in a separate process.
```
mdlinville: Is this the full set? If not, how can they know whether their integration has that limitation?

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -0,0 +1,168 @@
...
+   ```
+
+- Use `wandb.setup()` to set the `reinit` option globally for all runs in the current process. This is useful if you want to configure the behavior once and have it apply to all subsequent `wandb.init()` calls in that process.
+
+   ```python
+   import wandb
+   wandb.setup(wandb.Settings(reinit="<create_new|finish_previous|return_previous>"))
+   ```
+
+- Specify the desired value for `reinit` in an environment variable `WANDB_REINIT`. Defining an environment variable applies the reinit option to `wandb.init()` calls.
```
mdlinville:
```suggestion
- Specify the desired value for `reinit` in the environment variable `WANDB_REINIT`. Defining an environment variable applies the reinit option to `wandb.init()` calls.
```

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -0,0 +1,168 @@
...
+```python
+import wandb
+
+wandb.setup(wandb.Settings(reinit="create_new"))
+
+with wandb.init() as experiment_results_run:
+   for ...:
+      with wandb.init() as run:
+         # The do_experiment() function logs fine-grained metrics
+         # to the given run and then returns result metrics that
```
mdlinville:
```suggestion
         # to the given run and returns result metrics that
```

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -0,0 +1,168 @@
...
+         # you want to track separately.
+         experiment_results = do_experiment(run)
+
+         # After each experiment, log its results to a parent
+         # run. Each point in the parent run's charts corresponds
+         # to one experiment's results.
+         experiment_results_run.log(experiment_results)
+```
+## Example: Concurrent processes
```
mdlinville, line 86:
```suggestion

## Example: Concurrent processes
```
This may not matter in the output but Github isn't doing a good job with the syntax highlighting in the diff.

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -0,0 +1,168 @@
...
+
+Suppose you want to create a primary process that remains open for the script's entire lifespan, while periodically spawning short-lived secondary processes without finishing the primary process. This may occur if you want to keep a primary run (like training a model) active while computing evaluations or other tasks in separate runs.
```
mdlinville:
```suggestion
Suppose you want to create a primary process that remains open for the script's entire lifespan, while periodically spawning short-lived secondary processes without finishing the primary process. This may occur if you want to keep a primary run such as a training run active, while computing evaluations or other tasks in separate runs.
```

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -0,0 +1,168 @@
...
+
+Note three key points from the previous example:
+
+1. `reinit="create_new"` creates a new run each time you call `wandb.init()`.
+2. You keep references of each run. `wandb.run` does not automatically point to the new run created with `reinit="create_new"`. Store new runs in variables like `run_a`, `run_b1`, etc., and call `.log()` or `.finish()` on those objects as needed.
+3. You can finish sub-runs whenever you want while keeping the primary run open until.
+4. Finish your runs with `run.finish()` when you are done logging to them. This ensures that all data is uploaded and the run is properly closed.
+
+ 
+### Considerations
```
mdlinville: Is this section specific to the example it is in, or did you mean for it to be an H2?

I'm not crazy about how this section looks / is structured in the output. The "if" in line 157 is doing a lot of work and could be missed. I wonder if this is actually a list of antipatterns. WDYT?

mdlinville (changes requested):
Some small suggestions and a question or two.

content/guides/models/track/runs/multiple-runs-per-process.md
noaleetz: cc @timoffex to help answer here - more generally, would be good to understand what exactly we mean here by integration - is this any workflow that calls wandb.init() for the user?

noaleetz (approved):

noaleetz:
Nit but might be good to mention that this only available in the most recent SDK release - 0.19.10

content/guides/models/track/runs/multiple-runs-per-process.md
ngrayluna: This might refer frameworks that we have an integration for? https://docs.wandb.ai/guides/integrations/   Summoning @timoffex for confirmation.

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -13,25 +13,24 @@ Manage multiple runs in a single Python process. This is useful for workflows wh
...
+| `finish_previous` | Finish all active runs with `run.finish()` before creating a new one run with `wandb.init()`. Default behavior for non notebook environments. | Yes | Ideal when you want to break sequential sub-processes into separate individual runs. |
+| `return_previous` |  Return the most recent, unfinished run. Default behavior for notebook environments. | No | |
 
 {{% alert  %}}
-W&B does not support `create_new` mode for integrations that assume a single global run, such as Hugging Face Trainer, Keras callbacks, and PyTorch Lightning. If you use these integrations, you should run each sub-experiment in a separate process.
+W&B does not support `create_new` mode for [W&B Integrations]({{< relref "/guides/integrations/_index.md" >}}) that assume a single global run, such as Hugging Face Trainer, Keras callbacks, and PyTorch Lightning. If you use these integrations, you should run each sub-experiment in a separate process.
```
mdlinville:
```suggestion
W&B does not support `create_new` mode for [W&B Integrations]({{< relref "/guides/integrations/" >}}) that assume a single global run, such as Hugging Face Trainer, Keras callbacks, and PyTorch Lightning. If you use these integrations, you should run each sub-experiment in a separate process.
```
I believe that you can't link to the underbar index, only to the directory it is in. It's the one wrinkle to `relref`.

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -57,7 +56,7 @@ W&B does not support `create_new` mode for integrations that assume a single glo
    wandb.setup(wandb.Settings(reinit="<create_new|finish_previous|return_previous>"))
    ```
 
-- Specify the desired value for `reinit` in an environment variable `WANDB_REINIT`. Defining an environment variable applies the reinit option to `wandb.init()` calls.
+- Specify the desired value for `reinit` in the environment variable `WANDB_REINIT`. Defining an environment variable applies the `reinit` option to `wandb.init()` calls.
```
mdlinville, line 59:
```suggestion
- Specify the desired value for `reinit` in the environment variable `WANDB_REINIT`. Each  `wandb.init()` call reads the environment variable if it is set.
```

content/guides/models/track/runs/multiple-runs-per-process.md
```diff
@@ -83,9 +82,10 @@ with wandb.init() as experiment_results_run:
          # to one experiment's results.
          experiment_results_run.log(experiment_results)
 ```
+
 ## Example: Concurrent processes
 
-Suppose you want to create a primary process that remains open for the script's entire lifespan, while periodically spawning short-lived secondary processes without finishing the primary process. This may occur if you want to keep a primary run (like training a model) active while computing evaluations or other tasks in separate runs.
+Suppose you want to create a primary process that remains open for the script's entire lifespan, while periodically spawning short-lived secondary processes without finishing the primary process. This may occur if you want to keep a primary run active, such as training a model, while computing evaluations or other tasks in separate runs.
```
mdlinville:
```suggestion
Suppose you want to create a primary process that remains open for the script's entire lifespan, while periodically spawning short-lived secondary processes without finishing the primary process. For example, this pattern can be useful if you want to train a model in the primary run, but compute evaluations or do other work in separate runs.
```

mdlinville (approved):
A few more suggestion, skip if they are incorrect.

content/guides/models/track/runs/multiple-runs-per-process.md
ngrayluna: Good question, I am unsure if this is true....

ngrayluna: (Leaving out for now)

content/guides/models/track/runs/multiple-runs-per-process.md
timoffex: Sorry, missed the notification here!

> more generally, would be good to understand what exactly we mean here by integration - is this any workflow that calls wandb.init() for the user?

It's any workflow that uses `wandb` at all for the user. Some of these require the user to call `wandb.init()` and then use the global `wandb.run` by calling global functions like `wandb.log()`---these integrations will not work, and their error messages will confusingly suggest calling `wandb.init()` before using them even if the user already did.

It's not possible to have a full list of integrations because customers may have their own, private integrations with `wandb` and their own infra. The list @ngrayluna linked is a good resource though.

PR #1280, opened 2025-04-25
Update production script

mdlinville:
The previous fix for broken images in PR previews was overly broad and has resulted in incorrect domains in the English sitemap. This was because I made some incorrect assumptions about the [Cloudflare environment variables](https://developers.cloudflare.com/pages/configuration/build-configuration/#environment-variables). This PR aims to correct the problems introduced by that fix.

- Improve the check for whether we are in Cloudflare
- Determine if we are building main or a PR preview
- Echo the Cloudflare variables for debugging
- Only set a specific Hugo build context if we are in a PR preview
- Improve how the sitemap is moved around
- More code comments

To test:
- Review the Cloudflare build logs for this PR. The output should show all the variables and make the correct statements. ([Example](https://dash.cloudflare.com/832d6d41f129452d5f4797d566bd6cc8/pages/view/docs/df2b8f9a-bd8d-4388-9902-955f8c5d233f))
- Check out the PR branch locally and run `scripts/production.sh`. The build output should correctly detect that we are building locally.
- To test the `main` deploy with this fix, I think we need to merge this PR.


scripts/production.sh
```diff
@@ -1,17 +1,36 @@
...
     echo "Building locally"
     hugo
 fi
+
+# Move English sitemap into root
 rm public/sitemap.xml
-mv public/en/sitemap.xml public/sitemap.xml
+mv public/en/sitemap.xml public/
+
+# Clear Hugo translated content, which we don't yet use
```
mdlinville, line 33: Should we also remove the `en/` directory?

mdlinville:
I'm going to go ahead and merge this to hopefully fix the sitemap. If it all goes wrong, we can revert or iterate.

PR #1283, opened 2025-04-28
Add Segment CDP to docs site

johndmulhausen:
What it says on the tin. Per discussions with Cassie, Taylor, Dave, etc. 

We currently don't have full-fat telemetry and are only logging Google Analytics events at the moment. This should fix it 👍

### DIFF
--- a/layouts/partials/hooks/body-end.html
+++ b/layouts/partials/hooks/body-end.html
@@ -35,3 +35,7 @@
   });
   
 </script>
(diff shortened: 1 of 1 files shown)


hhuangwb (approved):
looks like a good v0, I'll circle back later with additional adjustments for both this and the weave-docs site to add proper consent management

johndmulhausen:
Thanks, @hhuangwb !

PR #1288, opened 2025-04-30
Remove production.sh script, use CloudFlare-managed Hugo environments

johndmulhausen:
Now that the translations have finally shipped, let's just use CloudFlare's off-the-shelf Hugo framework/environment and keep it simple. The maintenance benefits are obvious here; who knows how long the hacky overrides we've been doing will hold up. 

Note: I did a test deploy in prod and confirmed that this uses the correct domain in the sitemap URLs. See: docs.wandb.ai/en/sitemap.xml

Out of the box, this also will create the sitemap-of-sitemap approach for multiple languages, e.g. a top-level tree at the root that links into language-specific sitemaps, as supported by Google: https://developers.google.com/search/blog/2012/05/multilingual-and-multinational-site


johndmulhausen:
Oh, this actually is site-breaking. Low-pri anyway to get out of having to maintain this, so let's close this.

PR #1294, opened 2025-05-02
Fix multiple languages showing up in one support section

johndmulhausen:
Currently, the support section shows articles in all languages. While this is at least usable, it's definitely wrong and a bug that needs to be fixed. Let's narrow it down to the one you've already chosen.


johndmulhausen:
Closing; the real fix was to move the content into `en`

PR #1308, opened 2025-05-07
[DOCS-1479] Remove mentions of System tab

mdlinville:
[DOCS-1479] Remove mentions of System tab

- Intentionally updating some localized pages

Ready for peer review

[DOCS-1479]: https://wandb.atlassian.net/browse/DOCS-1479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/tutorials/integration-tutorials/monai_3d_segmentation.md
+++ b/content/tutorials/integration-tutorials/monai_3d_segmentation.md
@@ -581,7 +581,7 @@ for epoch in epoch_progress_bar:
 artifact.wait()
 ```
 
-Instrumenting the code with `wandb.log` not only enables tracking all metrics associated with the training and validation process, but also the all system metrics (our CPU and GPU in this case) on the W&B dashboard.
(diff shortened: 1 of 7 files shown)


ktmathews89 (approved):
thank you!

content/tutorials/integration-tutorials/monai_3d_segmentation.md
```diff
@@ -581,7 +581,7 @@ for epoch in epoch_progress_bar:
 artifact.wait()
 ```
 
-Instrumenting the code with `wandb.log` not only enables tracking all metrics associated with the training and validation process, but also the all system metrics (our CPU and GPU in this case) on the W&B dashboard.
+Instrumenting the code with `wandb.log` not only enables tracking all metrics associated with the training and validation process, but also logs all system metrics (our CPU and GPU in this case) on the W&B dashboard.
```
ngrayluna, line 584: Just a heads up that these docs are autogenerated (manual and have not been updated in ages) so if we were to regenerate these, this change will be overwritten.

ngrayluna (approved):

PR #1331, opened 2025-05-16
Runs: Add beta parameter and argument to code sample

ngrayluna:
Add newly added run setting available to make it easier to restrict which devices are collecting metrics. 

Note: Ideally we'd link to the Reference Docs. However, this feature is in beta and Settings Class is not currently available in public docs. For the latter, it'll get updated when we switch to lazydocs. 

Jira ticket: https://wandb.atlassian.net/browse/DOCS-1489

### DIFF
--- a/content/guides/models/track/log/distributed-training.md
+++ b/content/guides/models/track/log/distributed-training.md
@@ -118,7 +118,7 @@ Within the primary node, initialize a W&B run with [`wandb.init`]({{< relref "/r
 1. The `mode` parameter set to `"shared"` to enable shared mode.
 2. A unique label for [`x_label`](https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_settings.py#L638). You use the value you specify for `x_label` to identify which node the data is coming from in logs and system metrics in the W&B App UI. If left unspecified, W&B creates a label for you using the hostname and a random hash.
 3. Set the [`x_primary`](https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_settings.py#L660) parameter to `True` to indicate that this is the primary node.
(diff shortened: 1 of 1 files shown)


content/guides/models/track/log/distributed-training.md
```diff
@@ -113,11 +113,12 @@ To track multiple processes to a single run, you must have:
...
+Within the primary node, initialize a W&B run with [`wandb.init`]({{< relref "/ref/python/init.md" >}}). Pass in a `wandb.Settings` object to the `settings` parameter (`wandb.init(settings=wandb.Settings()`) with the following:
 
 1. The `mode` parameter set to `"shared"` to enable shared mode.
 2. A unique label for [`x_label`](https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_settings.py#L638). You use the value you specify for `x_label` to identify which node the data is coming from in logs and system metrics in the W&B App UI. If left unspecified, W&B creates a label for you using the hostname and a random hash.
 3. Set the [`x_primary`](https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_settings.py#L660) parameter to `True` to indicate that this is the primary node.
+4. Optionally provide a list of GPU indices ([0,1,2]) to `x_stats_gpu_device_ids` to specify which GPUs W&B tracks metrics for. If you do not provide a list, W&B tracks metrics for all GPUs on the machine.
```
mdlinville: I'd prefer "indexes" personally. Are we sure you can't provide a range? Is there a use case for logging metrics only for selective GPUs?

mdlinville (approved):
I'll go ahead and approve with a suggestion and a question.

content/guides/models/track/log/distributed-training.md
ngrayluna: Sadly it expected a list of integers. 

I'm guessing there could be a case where (e.g.) 2/3 GPUs are doing the heavy lifting and the third GPU is running not as compute heavy work.  Unsure how common this is.

PR #1336, opened 2025-05-20
[DOCS-1532] Document run metrics min and max thresholds

mdlinville:
[DOCS-1532] Document run metrics min and max thresholds

Ready for technical review

[DOCS-1532]: https://wandb.atlassian.net/browse/DOCS-1532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


content/guides/core/automations/automation-events.md
```diff
@@ -60,7 +60,7 @@ From the project's **Automations** tab or directly from a line plot panel, you c
 - A [system metric]({{< relref "/guides/models/app/settings-page/system-metrics.md" >}}) such as `cpu`, which tracks the percentage of CPU utilization. W&B logs system metrics automatically every 15 seconds.
 
 The notification can watch the metric for these events:
-- **Run metrics threshold met**: Trigger a workflow when for a given metric, a single logged value or the average logged values meets the threshold you specify.
+- **Run metrics threshold met**: Trigger a workflow when a single logged value for a metric, the average logged values for the metric over a number of runs, or the min or max value for the metric over a number of runs, meets the threshold you specify.
```
ssisk: I definitely think that introducing the phrase "over a number of runs" makes things more confusing. The average/min/max is calculated on the values from a single run. 

You use the phrase "logged value" for 2 of the 3, but just "value" for the 3rd, I assume that's a mistake. 

Would it be simpler to just say `Trigger a workflow when for a given metric, a single logged value or the average/min/max logged values meets the threshold you specify.`? Regardless, there's no reason to keep average and "min or max" in separate clauses of the sentence, it implies there's something different about them.

content/guides/core/automations/automation-events.md
```diff
@@ -60,7 +60,7 @@ From the project's **Automations** tab or directly from a line plot panel, you c
...
 - **Run metrics change threshold met**: Trigger a workflow when the average logged values of a run change by the absolute or relative threshold you specify.
```
ssisk: min/max also applies to the change threshold metrics.

content/guides/core/automations/automation-events.md
```diff
@@ -70,14 +75,23 @@ Run metric automations are currently available only in [W&B Multi-tenant Cloud](
...
-For **Run metrics threshold met** events, you configure:
+A **Run metrics threshold met** event can be triggered either when a metric's value meets a defined numeric value or when the value is the minimium or maximum value over a defined number of values.
+
+To specify a numeric threshold, you configure:
 1. The number of logged values to average across (defaults to 5).
 1. How to compare the values with the threshold.
-
 For example, trigger an automation when `accuracy` exceeds `.6`.
 
+To trigger the automation when a value is the minimum or maximum value, you configure:
```
mdlinville:
```suggestion
To trigger the automation when a value is the minimum or maximum value over a defined number of runs, you configure:
```

content/guides/core/automations/automation-events.md
mdlinville: PTAL, I restructured this area a bit more, to hopefully address this concern.

content/en/guides/core/automations/automation-events.md
```diff
@@ -70,14 +75,23 @@ Run metric automations are currently available only in [W&B Multi-tenant Cloud](
 {{% /alert %}}
 
 #### Threshold
-For **Run metrics threshold met** events, you configure:
+A **Run metrics threshold met** event can be triggered either when a metric's value meets a defined numeric value or when the value is the minimium or maximum value over a defined number of values.
+
+To specify a numeric threshold, you configure:
 1. The number of logged values to average across (defaults to 5).
```
ssisk: This still says it will be averaged, which is wrong. 

There's basically a new step where the user will configure the aggregation we will use, which will be either min max or average.

content/en/guides/core/automations/automation-events.md
```diff
@@ -59,9 +59,14 @@ From the project's **Automations** tab or directly from a line plot panel, you c
...
 - A [system metric]({{< relref "/guides/models/app/settings-page/system-metrics.md" >}}) such as `cpu`, which tracks the percentage of CPU utilization. W&B logs system metrics automatically every 15 seconds.
 
-The notification can watch the metric for these events:
-- **Run metrics threshold met**: Trigger a workflow when for a given metric, a single logged value or the average logged values meets the threshold you specify.
-- **Run metrics change threshold met**: Trigger a workflow when the average logged values of a run change by the absolute or relative threshold you specify.
+The notification can trigger a workflow when your criteria are met. Options include:
+
+- **Run metrics threshold met**:
+  - Trigger the automation when a logged metric's value meets a defined [threshold]({{< relref "#threshold" >}}), optionally averaged over a defined number of logged values. For example, trigger the automation when `accuracy` exceeds `.6`.
+  - Trigger the automation when a single logged value is the minimum or maximum value over a defined number of recently logged values.
```
ssisk: (continuing our discussion from https://github.com/wandb/docs/pull/1336#discussion_r2103411313) 

taking a look at the restructured version - I don't understand why this is broken out into a separate step, the relationship between the two bullet points here is confusing - are they an OR or an AND? 

It's basically that we're going to take the min, max or average of the values in the window and compare that to the threshold. 

(same below)

content/en/guides/core/automations/automation-events.md
```diff
@@ -70,14 +75,23 @@ Run metric automations are currently available only in [W&B Multi-tenant Cloud](
...
-For **Run metrics threshold met** events, you configure:
+A **Run metrics threshold met** event can be triggered either when a metric's value meets a defined numeric value or when the value is the minimium or maximum value over a defined number of values.
+
+To specify a numeric threshold, you configure:
 1. The number of logged values to average across (defaults to 5).
 1. How to compare the values with the threshold.
-
 For example, trigger an automation when `accuracy` exceeds `.6`.
 
+To trigger the automation when a value is the minimum or maximum value, you configure:
```
ssisk: I think it's confusing to treat min max as a separate way of configuring the trigger vs the rest

content/en/guides/core/automations/automation-events.md
```diff
@@ -87,11 +101,14 @@ The windows are consecutive and do not overlap.
...
-      - Decreases by at least
-      - Increases or decreases by at least
+1. The change threshold, one of the following:
+   - For a numeric threshold, you define:
+      1. Whether to evaluate the values as relative or absolute (defaults to **Relative**).
+      1. How to compare the values with the threshold:
+          - Increases by at least
+          - Decreases by at least
+          - Increases or decreases by at least
+    - For a **Min** or **Max** threshold, you define whether to check for the **Min** or **Max** value.
```
ssisk: it's not a min/max threshold, it's that you're choosing between average, min and max.

ssisk (changes requested):

content/en/guides/core/automations/automation-events.md
```diff
@@ -71,24 +71,31 @@ Run metric automations are currently available only in [W&B Multi-tenant Cloud](
...
+1. Whether to evaluate the **Average**, **Min**, or **Max** value within the window.
+1. The comparison to make:
+      - Above
+      - Above or equal to
+      - Below
+      - Below or equal to
+      - Not equal to
+      - Equal to
+
+For example, trigger an automation when average `accuracy` exceeds `.6`.
```
ngrayluna: For the last line, should "exceeds" be replaced with one of the values in the aforementioned list?

content/en/guides/core/automations/automation-events.md
```diff
@@ -71,24 +71,31 @@ Run metric automations are currently available only in [W&B Multi-tenant Cloud](
...
 
 #### Change threshold
 For **Run metrics change threshold met** events, the automation uses two "windows" of values to check whether to start:
```
ngrayluna, line 87: Might be useful to elaborate what windows means for Run metrics threshold met (above) too.

content/en/guides/core/automations/automation-events.md
```diff
@@ -71,24 +71,31 @@ Run metric automations are currently available only in [W&B Multi-tenant Cloud](
...
-1. The current window (defaults to 10).
-1. The prior window (defaults to 50).
+1. The current window of logged values (defaults to 10).
+1. The prior window of logged values (defaults to 50).
 1. Whether to evaluate the values as relative or absolute (defaults to **Relative**).
-1. How to compare the values with the threshold:
+1. The comparison to make:
       - Increases by at least
       - Decreases by at least
       - Increases or decreases by at least
```
ngrayluna, line 101: Would be nice to have an example here too (like what you have in the previous section).

ngrayluna (commented):
For future updates, maybe we can add a line or two in [project section](https://docs-1532.docodile.pages.dev/guides/core/automations/automation-events/#project) about scope vs event automations. (To tie the section together more).

content/en/guides/core/automations/automation-events.md
mdlinville:
```suggestion
For example, trigger an automation when average `accuracy` is above `.6`.
```

content/en/guides/core/automations/automation-events.md
mdlinville, line 101:
```suggestion
      - Increases or decreases by at least
      
      For example, trigger an automation when `loss` decreases by at least `.25`.
```

content/en/guides/core/automations/automation-events.md
mdlinville: I don't think I can make it work quite as well. I don't think I need to say more about the window than I already am, up there. I think there is more detail here because there are two windows to consider. WDYT?

ngrayluna: Should be fine. Wasn't sure if folks new what "window" meant in this context but maybe I was overthinking it

ssisk (approved):

ngrayluna (approved):

PR #1348, opened 2025-05-27
Upgrade to Docsy 0.12 (from 0.11) and Hugo 0.147.5 (from 0.140.2) and Node 19 (from 18)

johndmulhausen:
What it says on the tin

- Upgrades to Hugo 0.147.5
- Gets rid of customizations around `ja` UI strings and `render-heading` that are no longer necessary.
- All the upgrades here: https://github.com/google/docsy/releases/tag/v0.12.0
- All the upgrades from 0.140.2 to 0.147.5 listed here: https://github.com/gohugoio/hugo/releases
- Requires a Node version upgrade to run locally (I bumped via `nvm install 19`) 
- Updates header self-link fix to point to new partial location


mdlinville:
In the HTML preview, I do not see any heading self-links: https://docsy-upgrade.docodile.pages.dev/guides/

mdlinville (approved):

PR #1371, opened 2025-06-05
Create /en subdir in /content for English docs (fixes /support)

johndmulhausen:
The Support portal has an issue where the English content being at the root of /content is scoping all the taxonomy entries from /ko and /ja at build time and producing a support portal that has English, as well as Japanese and Korean support articles all rendering together on the English site. Example: https://docs.wandb.ai/support/anonymous/ ([Fixed version from this PR's deploy preview](https://support-en-only.docodile.pages.dev/support/anonymous/))

In addition we are doing a lot of directory filtering to produce the right list of stuff when translating the docs, and in general the content architecture of "ja" and "ko" living among "guides," "tutorials," et al is awkward. 

This PR fixes all of the above issues. 

**Note: Public-facing URLs are not impacted. This change only impacts docs contributors.**


hugo.yaml
```diff
@@ -5,7 +5,6 @@ buildFuture: true
 # cSpell:ignore goldmark github hugo readingtime docsy subdir lastmod pygments linenos catmullrom norsk gu
 
 # Language settings
-contentDir: content
 defaultContentLanguage: en
 defaultContentLanguageInSubdir: false
```
mdlinville, line 9: I think if this is `false` then we don't need to put the English content in `en/`. I would prefer not to have to remember to put the English content in `en/`, so what if we try not moving it and leaving this as `false`? Let me try that in a separate branch and see what happens. https://gohugo.io/configuration/languages/#base-settings

mdlinville: So, this is interesting. When I leave `defaultContentLanguageInSubdir` as `false`, move `content/en/*` back to `content/`, and set:
```yaml
languages:
  en:
    contentDir: content
```
it _almost_ works. But when I look in the English "Anonymous" article, a couple of Japanese support articles show up. I suspect the way to work around this would be set to set the language keys of all the English articles to English explicitly. I suppose this approach is better than that.

There are trade-offs with relying on the content directory to signal the page's content. For example, the pages must have exactly the same file name in each directory to be automatically linked together. Also, we still need to educate translators that they can set a localized page title, slug, URL in the translated page's front matte, and can set up customized menu too in the `_index.md`.

For this PR to be correct, I think I need to try setting `defaultContentLanguageInSubdir: true` and see what happens in a local build.

mdlinville: Well I never would have predicted this, but when you set `defaultContentLanguageInSubdir: true`, then `en/` gets added to the English URLs! For example, locally the home page is http://localhost:1313/en/. So I guess I am OK with approving this PR for now.

mdlinville (changes requested):
I'm confused about something and don't want to approve before I understand. I notice now that there is `content/guides/` and `content/en/guides/` and I believe they have duplicate content. Was the intent to remove the `content/guides/`? Same question for the `content/launch/` and other directories that are also in `content/en/`? For example, both of these local `diff` commands show that the directories have identical contents:

```shell
$ diff -r guides en/guides
$ diff -r launch en/launch
```

ngrayluna:
Before merging we'll need to make a few updates to docugen:
1. Here, add `en` to this path: https://github.com/wandb/docugen/blob/71015b0ea908ee198f9c3bc552ab4be4a3b33e86/action.yml#L101C47-L101C54
2. Make a new release tag for docugen 
3. Update the release tag number in the wandb/ github action https://github.com/wandb/wandb/blob/ef6074148a981eb0fb5688723ec34db70ed039e8/.github/workflows/generate-docodile-documentation.yml#L32C15-L32C35

I can approve and merge the first two, we'll need someone from the SDK Team to approve the PR off of wandb

johndmulhausen:
> I'm confused about something and don't want to approve before I understand. I notice now that there is `content/guides/` and `content/en/guides/` and I believe they have duplicate content. Was the intent to remove the `content/guides/`? Same question for the `content/launch/` and other directories that are also in `content/en/`? For example, both of these local `diff` commands show that the directories have identical contents:
> 
> ```shell
> $ diff -r guides en/guides
> $ diff -r launch en/launch
> ```

I am not seeing `content/guides` in this branch?

hugo.yaml
johndmulhausen: Yep, that's what that setting does, which is why it's false; I think `en/` on all the URLs is not what we want.

johndmulhausen:
> Before merging we'll need to make a few updates to docugen:
> 
> 1. Here, add `en` to this path: https://github.com/wandb/docugen/blob/71015b0ea908ee198f9c3bc552ab4be4a3b33e86/action.yml#L101C47-L101C54
> 2. Make a new release tag for docugen
> 3. Update the release tag number in the wandb/ github action https://github.com/wandb/wandb/blob/ef6074148a981eb0fb5688723ec34db70ed039e8/.github/workflows/generate-docodile-documentation.yml#L32C15-L32C35
> 
> I can approve and merge the first two, we'll need someone from the SDK Team to approve the PR off of wandb

@ngrayluna Added `.../en` in this PR: https://github.com/wandb/docugen/pull/60

mdlinville:
> > I'm confused about something and don't want to approve before I understand. I notice now that there is `content/guides/` and `content/en/guides/` and I believe they have duplicate content. Was the intent to remove the `content/guides/`? Same question for the `content/launch/` and other directories that are also in `content/en/`? For example, both of these local `diff` commands show that the directories have identical contents:
> > ```shell
> > $ diff -r guides en/guides
> > $ diff -r launch en/launch
> > ```
> 
> I am not seeing `content/guides` in this branch?

I can confirm that I'm not seeing these now, not sure what was happening. I did notice that `search.md` is still in the root. Is that page yet to be translated?

johndmulhausen:
> > > I'm confused about something and don't want to approve before I understand. I notice now that there is `content/guides/` and `content/en/guides/` and I believe they have duplicate content. Was the intent to remove the `content/guides/`? Same question for the `content/launch/` and other directories that are also in `content/en/`? For example, both of these local `diff` commands show that the directories have identical contents:
> > > ```shell
> > > $ diff -r guides en/guides
> > > $ diff -r launch en/launch
> > > ```
> > 
> > 
> > I am not seeing `content/guides` in this branch?
> 
> I can confirm that I'm not seeing these now, not sure what was happening. I did notice that `search.md` is still in the root. Is that page yet to be translated?

Ah, good catch; Search indeed needed to be moved.

mdlinville:
> Fixed version from this PR's deploy preview

It works in English, but when I try to go to, for example, `https://ec745638.docodile.pages.dev/ja/search/?q=Foo` I get a 404, and if I switch my language to Japanese and then search, the page is in English. Same in Korean. Is that expected? I noticed they don't have a `search.md`.

mdlinville:
To resolve the conflicts, you may have better luck with an interactive rebase than a merge. Let me know if you want help with it. Noah pushed a new API version in the meantime.

johndmulhausen:
> > Fixed version from this PR's deploy preview
> 
> It works in English, but when I try to go to, for example, `https://ec745638.docodile.pages.dev/ja/search/?q=Foo` I get a 404, and if I switch my language to Japanese and then search, the page is in English. Same in Korean. Is that expected? I noticed they don't have a `search.md`.

https://ec745638.docodile.pages.dev/search/?q=Foo pulls up a results page for me, and that page is where you end up if you search from the Japanese docs. It didn't get translated, no. (Not sure that one, or the 404 page, can be translated?)

> To resolve the conflicts, you may have better luck with an interactive rebase than a merge. Let me know if you want help with it. Noah pushed a new API version in the meantime.

That would be amazing to get your help

mdlinville (approved):

PR #1373, opened 2025-06-05
Launch: Move, hide Launch APIs

ngrayluna:
Per Slack thread, instructed to hide Launch related objects for now. 

Ticket: https://wandb.atlassian.net/browse/DOCS-1565


static/_redirects
```diff
@@ -183,6 +183,10 @@
 /guides/launch/run-agent/ /guides/launch/setup-agent-advanced/ 301
 /guides/launch/sagemaker/ /guides/launch/setup-launch-sagemaker/ 301
```
ngrayluna: @johndmulhausen Should "guides" be removed in these old Launch docs?

static/_redirects
```diff
@@ -183,6 +183,10 @@
...
 /guides/launch/vertex/ /guides/launch/setup-vertex/ 301
+/ref/python/launch-library/ /launch-library/ 301
```
ngrayluna: Based on the above comment, might need to update redirect path to include "guides"

mdlinville: If all of the pages still exist but just under the new path, then use a splat in place of all of these rules, down in the bottom section.

```
/guides/launch/* /launch-library/:splat 301
```
Or to redirect all the old pages to a single page:
```
/guides/launch/* /launch-library/ 301
```
Looks like you should use splats for the ref redirects too.

ngrayluna: Makes sense. Quick question: When do we use splats vs manually specifying each page? Is the former because it is more efficient + less error prone?

mdlinville: You should always use splats if possible, with specific rules that should override the splat (because the source or destination page is different, for example) coming first. There are some times when we should have used splats before (like these) but that I wasn't sure about when I converted most of the redundant rules to splats. Cloudflare has a pretty conservative limit on the # of individual rules (I think it may be 100 or 1000).

mdlinville: In https://github.com/wandb/docs/pull/1373#discussion_r2130577037 there are two different kinds of splats. In the first example, any request to a page in `/guides/launch/` will redirect to the same page name under `/launch-library/`. In the second example, any request to a page in `/guides/launch/` will redirect to exactly the page `/launch-library/` (the landing page, in this case).

content/launch-library/launch.md
```diff
@@ -1,5 +1,6 @@
 ---
-title: launch
+title: launch api
+url: guides/launch-library/launch
```
mdlinville: The content itself is not in `guides/` any longer, right? Is intentional to manually add it to the URLs? I believe if you don't want `guides/` to be part of the URL, you should not set the `url` in the page front matter.

mdlinville (changes requested):

content/launch-library/launch.md
ngrayluna: OK. So I think I understand why I was confused. 

The Launch docs (that are hidden) are here: https://docs.wandb.ai/guides/launch/ (you can't access them unless you know the path).

Those docs (listed in `content/launch/`) do contain the `url` key.

In my case, if I want them to show up in https://docs.wandb.ai/launch-library, change the value for the `url`  key to just `launch-library`

ngrayluna: (which I just did in the latest commit)

ngrayluna: It seems, at least based on what was done in the past, that we can set the url key in the frontmatter.

static/_redirects
ngrayluna: Update: I believe these are in place because the `url` keys in the respective pages contain "guides/launch/...."

mdlinville (approved):
We synced offline about the `url` keys. :)

mdlinville (approved):

PR #1374, opened 2025-06-06
Fix table of contents, consolidate fonts

johndmulhausen:
1. The Table of Contents has uneven spacing and is very oversized; CoreWeave sets their TOC `font-size` to a bit below 13px; it's set to 13px here.
2. There is a lot of surface area where the default Docsy font, Open Sans, pokes through -- even in the main text. Source Sans Pro is used in other places -- in fact, every time we make an active choice. We shouldn't whipsaw between two reading fonts, in any case, and Source Sans Pro seems to be the preference (and, to my eyes, looks more modern/readable). So, consolidating the reading surfaces to Source Sans Pro, here.
3. The CoreWeave docs don't need an "On This Page" header above the Table of Contents and neither do we.

Pay special attention to the TOC on the right in these comparisons, but also note the more modern-looking font in the main text body: 

Before: https://docs.wandb.ai/guides/artifacts/construct-an-artifact/
After: https://styles.docodile.pages.dev/guides/artifacts/construct-an-artifact/

### DIFF
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -794,6 +794,14 @@ img, iframe {
   display: block;
 }
 
+.section-index {
+    max-width: 80%;
+}
+
+.section-index .entry p {
+    margin-top: 5px;
+}
+
 @media (min-width: $breakpoint-lg) {
     img, iframe {
         max-width: 80% !important;
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -117,6 +117,15 @@ body {
         background-color: $dark-bg-primary;
     }
 }
+p, td, div
+{
+    font-family: "Source Sans Pro", sans-serif;
+}
+.td-sidebar-nav__section-title, .td-sidebar-nav__section without-child 
+{
+    font-size: 16px; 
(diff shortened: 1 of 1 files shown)


mdlinville (commented):
Nit: In the section TOC at the bottom of the footer, like [here](https://styles.docodile.pages.dev/guides/app/), I think there should be a bit more vertical space between the title and description (where there is a description). I also wonder whether the content should wrap a bit more tightly. to align with the main content column. I'll hold off on approving in case you'd like to address it before auto-merge kicks in. [Here's](https://styles.docodile.pages.dev/guides/app/features/panels/) another example with a bit more text to play with down in that bottom section.

johndmulhausen:
> Nit: In the section TOC at the bottom of the footer, like [here](https://styles.docodile.pages.dev/guides/app/), I think there should be a bit more vertical space between the title and description (where there is a description). I also wonder whether the content should wrap a bit more tightly. to align with the main content column. I'll hold off on approving in case you'd like to address it before auto-merge kicks in. [Here's](https://styles.docodile.pages.dev/guides/app/features/panels/) another example with a bit more text to play with down in that bottom section.

These are fixed 👍

mdlinville (approved):

PR #1394, opened 2025-06-16
Use Deepnote for embedded notebook functionality

johndmulhausen:
Creates a new Quickstart that takes the classic MNIST-on-wandb tutorial notebook, converts it into Deepnote (a notebook platform that supports embedding notebooks in arbitrary webpages), and embeds it in a new page that offers an in-docs/live alternative to our current Quickstart.


mdlinville:
I reckon you can close this PR now in favor of the hackathon implementation, right?

mdlinville:
This PR is toast, right? We will need to re-implement this after the migration (but I think we know how already).

johndmulhausen:
I think Weave and Inference are good products for this maybe but Models is not. Closing for now.

PR #1397, opened 2025-06-18
Launch: Move launch ref (again)

ngrayluna:
This was reverted in the [PR when en/ was added](https://github.com/wandb/docs/pull/1371/files#diff-4471657b712abce5cd73781b7619ff5577762a44234801ddc3e747f93edc2120) .

Original PR: https://github.com/wandb/docs/pull/1373/files


mdlinville (approved):
There is a chance that the link checker will still report a couple false positives after this merges. If so, some entries will need to be added to `.lycheeignore` at the root of the repo.

PR #1408, opened 2025-06-23
remove launch apis

ngrayluna:
Manually removes Launch APIs (`wandb.apis.public`): https://wandb.atlassian.net/browse/DOCS-1595

The next SDK release won’t generate them anymore: 
* Ticket: https://github.com/wandb/docugen/commit/7787238b02098c631f837c5453a568d0b0dea04b
* Jira: https://wandb.atlassian.net/browse/DOCS-1594


static/_redirects
```diff
@@ -360,6 +360,9 @@
 /ref/python/inittorch.view_as_real/ /ref/python/init/ 301
 /ref/python/integrations/sklearn/ /guides/integrations/scikit/ 301
 /ref/python/join/ /ref/python/ 301
+/ref/python/public-api/queuedrun/ /ref/python/init/ 301
```
mdlinville, line 363: When I look at https://redirect-launch-apis-public.docodile.pages.dev/ref/python/init/, I see no reference to "launch", so if I actually encountered one of these redirects as a reader, I would be very confused. Is there anything that can be added to the target page to help?

mdlinville (changes requested):

static/_redirects
ngrayluna: Launch Team wants to remove all mentions of Launch APIs. No info or guidance as to where to redirect.

mdlinville (approved):

PR #1414, opened 2025-06-24
Removed keras integration ref docs

ngrayluna:
Per discussion with SDK Team, remove Integration docs from public ref guide. 

Jira ticket: https://wandb.atlassian.net/browse/DOCS-1597


content/en/guides/integrations/keras.md
```diff
@@ -258,7 +258,7 @@ See our [example repo](https://github.com/wandb/examples) for scripts, including
 
 The `WandbCallback` class supports a wide variety of logging configuration options: specifying a metric to monitor, tracking of weights and gradients, logging of predictions on training_data and validation_data, and more.
 
-Check out [the reference documentation for the `keras.WandbCallback`]({{< relref "/ref/python/integrations/keras/wandbcallback.md" >}}) for full details.
+Check out the reference documentation for the `keras.WandbCallback` for full details.
```
mdlinville, line 261: How do they get to the reference documentation now to check it out?

mdlinville (changes requested):

content/en/guides/integrations/keras.md
ngrayluna: They won't. They'll need to use __doc__ (dunder method) to see it.  This was per request of the SDK Team because these APIs are some combination of broken and/or not maintained.

mdlinville: Got it, as long as they will understand the new call to action here.

mdlinville (approved):

PR #1457, opened 2025-07-21
Artifacts: Add docs on downloading reference artifacts  with checkum

ngrayluna:
Customer is requesting for documentation to be added to the SDK site about not being able to download artifacts created with "artifact.add_reference" with checksum disabled, and suggesting that setting "checksum=true" will solve the issue. This request has been made before and has a +2 impact.

Jira ticket: https://wandb.atlassian.net/browse/DOCS-1014

Preview: https://claude-code-docs-1014.docodile.pages.dev/guides/artifacts/download-and-use-an-artifact/
Preview: https://claude-code-docs-1014.docodile.pages.dev/guides/artifacts/construct-an-artifact/#add-a-uri-reference

Status: Ready for tech review/

📄 **[View preview links for changed pages](https://github.com/wandb/docs/pull/1457#issuecomment-3202090694)**


content/en/guides/core/artifacts/download-and-use-an-artifact.md
```diff
@@ -14,6 +14,20 @@ Download and use an artifact that is already stored on the W&B server or constru
 Team members with view-only seats cannot download artifacts.
 {{% /alert %}}
 
+{{% alert color="warning" %}}
+Artifacts created with reference URIs using `checksum=False` cannot be downloaded. When you add a reference to an artifact using `artifact.add_reference(..., checksum=False)`, W&B only stores the URI reference without the actual file contents or checksums. To ensure your artifacts can be downloaded, use `checksum=True` (the default) when adding references:
```
amusipatla-wandb:
```suggestion
Artifacts created with reference URIs using `checksum=False` cannot be downloaded. When you add a reference to an artifact using `artifact.add_reference(..., checksum=False)`, W&B only stores the URI reference without the checksums. To ensure your artifacts can be downloaded, use `checksum=True` (the default) when adding references:
```
for reference artifacts we never store the file contents, so I think keeping this in might be misleading

content/en/guides/core/artifacts/download-and-use-an-artifact.md
```diff
@@ -14,6 +14,20 @@ Download and use an artifact that is already stored on the W&B server or constru
 Team members with view-only seats cannot download artifacts.
 {{% /alert %}}
 
+{{% alert color="warning" %}}
```
mdlinville, line 17: Unless I am misunderstanding something, I'd align this with the note in the previous file. If you take the bold Important out of that one, then you can move the admonition to an include and invoke it both places with `readfile`.

content/en/guides/core/artifacts/download-and-use-an-artifact.md
```diff
@@ -14,6 +14,18 @@ Download and use an artifact that is already stored on the W&B server or constru
...
+Artifacts created with reference URIs using `checksum=False` cannot be downloaded. To ensure your artifacts can be downloaded, use `checksum=True` (the default) when adding references:
```
mdlinville, line 18:
```suggestion
If you set `checksum=False` when adding a reference, the artifact cannot be downloaded later. Only use `checksum=False` when you need to track reference URIs without downloading capability. For downloadable artifacts, always use `checksum=True` (the default):
```
This way the notes at least share most of the text. But if checksum=True is the default, then line 22 seems not to need the `checksum=True` and it might be more clear to leave it off.

mdlinville (approved):
LGTM, with a non-blocking suggested edit to align the text in the notes and a followup question.

ngrayluna:
Closing. (see ticket thread)

PR #1424, opened 2025-06-27
Lazydocs: API Reference doc refresh

ngrayluna:
Updates API Reference docs using `lazydocs`. This PR replaces content from `docugen` (/ref/python/*).  Adds redirects to APIs that were moved. 

This is a manual deploy. Follow up PRs will use [GitHub action (see this PR)](https://github.com/wandb/wandb/pull/10066) to update ref docs (like we do w/ wandb/docugen). 

Preview: https://lazydocs-preview.docodile.pages.dev/ref/python/

Latest Lychee report: https://github.com/wandb/docs/actions/runs/16179819281

PR #1501, opened 2025-08-07
Update reference docs to sdk@v0.21.1, core@master

davidwallacejackson:
Used https://github.com/wandb/wandb/commit/83e1dbfc87b33c21dfda6cc660f0acce153ea275


content/en/ref/python/.md
mdlinville, line 1: What's this file about?

content/en/ref/python/_index.md
mdlinville, line 1: Are the changes in this file correct? Removal of the cards and the SDK version look like regressions, at a glance.

content/en/ref/python/artifact.md
mdlinville, line 1: How can the Artifact reference be new?

mdlinville (changes requested):
I'm going to stop reviewing here because I have concerns about this PR. I think Noah might have [done the last update manually](https://github.com/wandb/docs/pull/1424) and this PR might be effectively undoing that. Can you confirm that this PR is basically correct? Since Noah has way more context than I do, it might be best to wait until they are back from PTO.

PR #1503, opened 2025-08-08
UI: Breadcrumb home icon + left-nav animation/spacings

johndmulhausen:
This PR adds a bunch of CSS tweaks and also a couple of features:

- Breadcrumb: render a Home icon as the first item via the Hugo partial (no JS injection). 
- Left nav: adds smooth expand/collapse animation, rotating caret, fixes caret clipping, and normalizes spacing before/after child groups.

Partially inspired by me wanting to try GPT-5 :)

### DIFF
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -560,7 +560,6 @@ ul.foldable, .card {
 
 // Table of Contents Component
 #TableOfContents {
-    font-family: $base-font;
     a{
         margin: 0px !important;
         padding: 8px !important;
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -129,6 +129,10 @@ p, td, div, li, .lead
     }
 }
 
+#content-mobile {
+    max-height: 65px;
+}
+
 .td-sidebar-nav__section-title, .td-sidebar-nav__section without-child 
 {
     font-size: 14px; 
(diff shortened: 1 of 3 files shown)


assets/scss/_variables_project.scss
```diff
@@ -117,34 +118,56 @@ body {
...
+}
+h3, .td-content h3 { 
+    font-size: 20px !important;
+    font-weight: $font-weight-bold !important;
+}
+h4, .td-content h4 { 
+    font-size: 16px !important;
+    font-weight: $font-weight-bold !important;
+}
+h5, .td-content h5 {
```
mdlinville, line 163: This is now the same as the base font, which means that H6 is smaller than the base font. WDYT?

assets/scss/_variables_project.scss
```diff
@@ -6,6 +6,8 @@ Add styles or override variables from the theme here.
 
 @import url('https://fonts.googleapis.com/css2?family=Source+Serif+4:ital,opsz,wght@0,8..60,200..900;1,8..60,200..900&display=swap');
 
+$base-font: system-ui, -apple-system, Segoe UI, Roboto, Ubuntu, Cantarell, Noto Sans, sans-serif, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
```
mdlinville, line 9: This is quite a few font face changes. Are we intentionally going away from Source Sans Pro?

assets/scss/_variables_project.scss
```diff
@@ -488,20 +560,31 @@ ol.breadcrumb li:not(:first-child):before {
...
 #TableOfContents {
-    
+    font-family: $base-font;
     font-size: 13px !important;
-    font-family: "Source Sans Pro", sans-serif;
     a{
         margin: 0px !important;
         padding: 8px !important;
         font-weight: 500 !important;
+        font-size: 13px !important;
```
mdlinville: I wonder if this should also be 12 to align with the breadcrumbs, which are also a navigation element. Also not sure you need line 564 and 569 both. It would be bad if they went out of sync.

assets/scss/_variables_project.scss
```diff
@@ -488,20 +560,31 @@ ol.breadcrumb li:not(:first-child):before {
...
     }
     // List styles
     ul { 
         padding: 0px !important;
         li {
             padding: 0px !important; 
             line-height: 14px;
+            font-size: 13px !important;
```
mdlinville: Same feedback here about the font size. Probably all of the links in the navigation are in `<ul><li>` hierarchy.

assets/scss/_variables_project.scss
```diff
@@ -1083,18 +1173,21 @@ div.td-page-meta__lastmod {
 
 // Update the sidebar navigation styles
 .td-sidebar-nav {
-    font-family: "Source Sans Pro", sans-serif;
+    font-family: $base-font;
     padding-left: 15px !important;
+    font-size: 14px !important; // Slightly larger than right nav
```
mdlinville, line 1187: This one is 14 as opposed to 13 above, and these are still navigation `<ul><li>`, right?

mdlinville (changes requested):
Feedback from skimming the CSS changes. Next I will spot check some pages in case there is more feedback there.

mdlinville (changes requested):
Screenshot 2025-08-08 at 3 32 54 PM

Maybe my eyes are playing tricks, but the terminal breadcrumb looks less bold than the others and the black on gray is harder to read than the teal with no background color.

Screenshot 2025-08-08 at 3 34 46 PM

To my eyes, these headings look way too heavy.

Screenshot 2025-08-08 at 3 35 39 PM

These titles also look heavy and large.

Screenshot 2025-08-08 at 3 36 48 PM

Code font in paragraph text looks quite large to me.

mdlinville:
One more: In dark mode, the left nav looks too light / not bold enough. I think it is an optical illusion, but it may point to a need to make the navigation a little heavier in both modes.
Screenshot 2025-08-08 at 3 39 47 PM

assets/scss/_variables_project.scss
johndmulhausen: Oh totally; looks way better and is the font of choice on the Weave docs.

assets/scss/_variables_project.scss
johndmulhausen: Good catch; fixed.

assets/scss/_variables_project.scss
johndmulhausen: Good calls on both things. Fixed.

assets/scss/_variables_project.scss
johndmulhausen: Right, removed this definition; everything should be styled by just the font-size definition in `a` now

assets/scss/_variables_project.scss
johndmulhausen: Right-nav is now 12, so I think having the left-nav be 14 makes sense; page titles are shorter than headings generally and we frequently shorten them even more with the `linkTitle` front-matter for pure "scan-ability." This one I'm calling good. However, we'll address font-weight below. :)

johndmulhausen:
> Maybe my eyes are playing tricks, but the terminal breadcrumb looks less bold than the others and the black on gray is harder to read than the teal with no background color.
> 
> To my eyes, these headings look way too heavy.
> 
> These titles also look heavy and large.
> 
> Code font in paragraph text looks quite large to me.

Good call on the breadcrumbs. I am matching the font weight on Docusaurus re: headings and sizes so I'm going to leave those as-is, I think they look great.  Re: inline code, though, I think what was making them look off was the background color on the pills was nigh transparent. I think properly framed by a visible pill it makes a lot more sense now.

> One more: In dark mode, the left nav looks too light / not bold enough. I think it is an optical illusion, but it may point to a need to make the navigation a little heavier in both modes. 

Bolded up the left-nav a bit. 

Responded to the other bits inline.

assets/scss/_variables_project.scss
```diff
@@ -488,20 +560,31 @@ ol.breadcrumb li:not(:first-child):before {
 
 // Table of Contents Component
 #TableOfContents {
-    
+    font-family: $base-font;
     font-size: 13px !important;
```
mdlinville: Is this one still redundant?

johndmulhausen: It is, zapped

johndmulhausen:
I meant to just close this; as a thought experiment it was fun but given that we're moving to Mintlify it's a pointless exercise. After merging, instantly reverted. This can stay dead.

PR #1547, opened 2025-08-20
fix: add error parameters to Lychee action to catch blocks for Node.js compatibility

mdlinville:
## Description

This PR fixes a potential syntax error in the link checker script that was causing issues in [GitHub Actions run #17112743303](https://github.com/wandb/docs/actions/runs/17112743303/job/48537533886#step:7:32).

## Problem

The `fix_broken_links.mjs` script was using optional catch binding syntax (`catch {}` without an error parameter), which might not be supported in all Node.js environments or could be causing parsing issues in the GitHub Actions runner.

## Solution

Updated all catch blocks to include the error parameter:
- `catch {}` → `catch (e) {}`

This ensures compatibility across different Node.js versions and environments.

## Changes

- Updated 4 catch blocks in `fix_broken_links.mjs`:
  - Line 32: `normalizeUrl` function
  - Line 115: Controller cleanup in `probeRedirect`
  - Line 146: URL parsing in `candidateVariants`
  - Line 231: URL manipulation in `generateFromCandidates`

## Testing

The updated code maintains the same functionality while being more compatible with different JavaScript environments. The error parameter is added but not used (as was the original intent with optional catch binding).

### DIFF
--- a/.github/scripts/fix_broken_links.mjs
+++ b/.github/scripts/fix_broken_links.mjs
@@ -29,7 +29,7 @@ function normalizeUrl(u) {
 	try {
 		const url = new URL(u);
 		return url.toString();
-	} catch {
+	} catch (e) {
 		return u;
 	}
 }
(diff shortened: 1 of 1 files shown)


mdlinville:
I've rebased this PR to properly base it on main. The PR now only contains the catch syntax fix and no longer includes the base URL changes from PR #1546. 

The branch has been force-pushed with just the single commit that adds error parameters to the catch blocks in `fix_broken_links.mjs`.

PR #1560, opened 2025-08-22
Add diagrams to artifact and run deletion documentation

johndmulhausen:
- Added a comprehensive diagram illustrating the artifact garbage collection process, clarifying the soft-delete mechanism and subsequent file deletion.
- Introduced a new section detailing the run deletion workflow, including options for deleting associated artifacts and handling Model Registry links.

Resolves DOCS-1005

📄 **[View preview links for changed pages](https://github.com/wandb/docs/pull/1560#issuecomment-3215351439)**


ngrayluna (commented):
@johndmulhausen Do you have someone lined up from the Artifacts team to review this? If not, I'll message them.

johndmulhausen:
> @johndmulhausen Do you have someone lined up from the Artifacts team to review this? If not, I'll message them.

Pinged #artifacts here; also took note of some names per our discussion in triage in the Jira

estellazx:
@johndmulhausen This looks good as is!
I have 2 things that might be also good to include
1. Should we mention that deleting an entity, project, run (if also delete artifacts), or artifact collection will also result in `Delete an artifact` page process?
2. [Protected aliases](https://docs.wandb.ai/guides/core/registry/model_registry/access_controls/) might be worth mentioning. Protected aliases are aliases in the Registry that registry admins can set, if an artifact has a protected alias, it can't be deleted by non-registry admins

amusipatla-wandb:
> [Protected aliases](https://docs.wandb.ai/guides/core/registry/model_registry/access_controls/) might be worth mentioning. Protected aliases are aliases in the Registry that registry admins can set, if an artifact has a protected alias, it can't be deleted by non-registry admins

just to clarify here, within a registry, registry admins can unlink the protected artifact versions and delete collections/registries with protected aliases

however, for the source artifacts: a source artifact that is linked to a registry with a protected alias cannot be deleted by any user, **but** the registry admins can remove the protected aliases and then delete the source artifact

johndmulhausen:
@estellazx @amusipatla-wandb Added some color based on your suggestions. What do you think?

estellazx (approved):

PR #1572, opened 2025-08-27
chore(weave): Add weave integration docs

andrewtruong:
https://wandb.atlassian.net/browse/WB-27235

📄 **[View preview links for changed pages](https://github.com/wandb/docs/pull/1572#issuecomment-3228174953)**


content/en/guides/integrations/weave.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Auto-initialize Weave with W&B
+
+Initialize a W&B run and import Weave. No extra configuration is required—Weave will auto-init for you.
+
+```python
+import wandb
+import weave
+
+wandb.init(project="weave-demo")
```
ngrayluna: So the usage here is to initialize a run as a background process, but not explicitly use the returned run object? Once the run finishes, it'll provide a link to the W&B dashboard, etc.?

andrewtruong: That's correct, although you can use the run object if you want

dbrian57 (approved):

PR #1609, opened 2025-09-08
[WB-27552] Update docs for registry env var for server installs

ibindlish:
Description
-----------

[WB-27552](https://wandb.atlassian.net/browse/WB-27552)

W&B Registry was made generally available in Server v0.70.0. This PR updates the docs to reflect that. The `ENABLE_REGISTRY_UI` env var is still around in case an organization wish to disable the Registry UI altogether. 

Related issues
-----------
- Fixes DOCS-12345 
- Fixes #12345 

📄 **[View preview links for changed pages](https://github.com/wandb/docs/pull/1609#issuecomment-3268358180)**

[WB-27552]: https://wandb.atlassian.net/browse/WB-27552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


content/en/guides/core/registry/_index.md
```diff
@@ -106,7 +106,7 @@ Based on your deployment type, satisfy the following conditions to enable W&B Re
 | ----- | ----- |
 | Multi-tenant Cloud | No action required. W&B Registry is available on the W&B App. |
 | Dedicated Cloud | Contact your account team to enable W&B Registry for your deployment. |
-| Self-Managed | Set the environment variable `ENABLE_REGISTRY_UI` to `true`. Refer to [Configure environment variables]({{< relref "/guides/hosting/env-vars.md" >}}). Requires Server v0.59.2 or newer. |
+| Self-Managed | For Server v0.70.0 or newer, no action required. For Server v0.59.2 or newer, set the environment variable `ENABLE_REGISTRY_UI` to `true`. Refer to [Configure environment variables]({{< relref "/guides/hosting/env-vars.md" >}}). |
```
ibindlish: What's the process around updating the Japanese and Korean translations of these guides?

mdlinville: It is out of your scope. :) 

For the Server version, don't mention versions that old. We are trying to EOL versions in a rolling fashion, and currently the oldest version is 0.60, set to EOL with v0.74. Giving specifics will quickly become tech debt. So just say "For older supported Server versions, set....."

mdlinville (changes requested):

mdlinville:
Please update the PR title to remove the conventional commit prefix. We don't use that in Docs PRs. If you have a JIRA where you are tracking this work, it would be wonderful if you mention it in the title and description.

content/en/guides/hosting/env-vars.md
```diff
@@ -33,7 +33,7 @@ In addition to configuring instance level settings via the System Settings admin
 | `GORILLA_CUSTOMER_SECRET_STORE_SOURCE` | Sets the secret manager for storing team secrets used by W&B Weave. These secret managers are supported: <ul><li><b>Internal secret manager</b> (default): <code>k8s-secretmanager://wandb-secret</code></li><li><b>AWS Secret Manager</b>: <code>aws-secretmanager</code></li><li><b>GCP Secret Manager</b>: <code>gcp-secretmanager</code></li><li><b>Azure</b>: <code>az-secretmanger</code></li><ul>  |
 | `GORILLA_DATA_RETENTION_PERIOD`    | How long to retain deleted data from runs in hours. Deleted run data is unrecoverable. Append an `h` to the input value. For example, `"24h"`. |
 | `GORILLA_DISABLE_PERSONAL_ENTITY`  | When set to true, turns off [personal entities]({{< relref "/support/kb-articles/difference_team_entity_user_entity_mean_me.md" >}}). Prevents creation of new personal projects in their personal entities and prevents writing to existing personal projects. |
-| `ENABLE_REGISTRY_UI`               | When set to true, enables the new W&B Registry UI.            |
+| `ENABLE_REGISTRY_UI`               | When set to true, enables the new W&B Registry UI. (Deprecated in Server v0.74.0)           |
```
mdlinville: Do you mean deprecated (the variable still exists) or removed (the variable no longer exists or no longer works)?

content/en/guides/hosting/hosting-options/self-managed/kubernetes-operator/operator-airgapped.md
```diff
@@ -197,8 +197,6 @@ spec:
         password: password
         port: 3306
         user: wandb
-      extraEnv:
```
mdlinville: If we are actually removing it now, then we had no deprecation period and the other change I flagged is probably incorrect. PTAL.

mdlinville (changes requested):
Thanks for the quick edit. I could use some clarification about the removal of the migration environment variable.

content/en/guides/hosting/env-vars.md
ibindlish: Ah good catch. `Removed` would be correct here.

content/en/guides/hosting/hosting-options/self-managed/kubernetes-operator/operator-airgapped.md
ibindlish: Yeah, we'll be removing the environment variable. Does that mean I should just remove it from the list of environment variables in `env-vars.md`?

mdlinville: Maybe you need to leave it, but say that the environment variable is ignored in v0.74.x and above because it is the default. Something like that. WDYT?

mdlinville:
Adde a label because we can't merge this until v0.74.0.

content/en/guides/hosting/hosting-options/self-managed/kubernetes-operator/operator-airgapped.md
ibindlish: Got it. I misunderstood your previous comment, but it makes sense to me now. Users on an older server version might still refer to these docs and need a working config. Added it back! And apologies for the confusion

content/en/guides/hosting/env-vars.md
ibindlish: Sorry - I missed your comment about needing some clarification around this. I'd added it to the PR description, but lmk if you have any follow up questions!

mdlinville (approved):

PR #1617, opened 2025-09-10
Fix Pydantic formatting: Settings & Automations

ngrayluna:
Fixes formatting for objects defined with Pydantic. 

Note[1]: This was manually copied over from generated docs. 
Note[2]: There are some new args that are being surface, need to check with Dima and Tony if they should be hidden or exposed (and are just missing descriptions).

📄 **[View preview links for changed pages](https://github.com/wandb/docs/pull/1617#issuecomment-3273151543)**


johndmulhausen (changes requested):
So much improved, but we need to swap the places of the code blocks and the descriptive text to lead with the code blocks. Then I think we're good here 👍

johndmulhausen (approved):
Great work here @ngrayluna !

PR #1637, opened 2025-09-12
Fix "Too many redirects" in some pages

mdlinville:
Fix some specific issues in the Cloudflare `_redirects` file that emerged when we fixed a much more pervasive bug with the file

- Removed circular redirects
- Removed redundant redirects
- Removed redirects that are already handled by Hugo frontmatter like `url`, `alias`, and cascade keys (including some that were in conflict with each other or the `_redirects` file)
- Small fixes to the ordering of rules

**Note**: The trailing slash 404 issue is not fixed. Expect URLs like `/guides/tables` to result in a 404.

To test:
- In the PR review, confirm that https://fix-too-many-redirects.docodile.pages.dev/guides/tables/ and https://fix-too-many-redirects.docodile.pages.dev/guides/registry/ URLs no longer result in "too many redirects"
- In the PR preview, spot check other pages and other redirects you know about, like https://fix-too-many-redirects.docodile.pages.dev/guides/core/automations/

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -85,7 +85,6 @@
 /frameworks/pytorch/ /guides/integrations/pytorch/ 301
 /getting-started/ /quickstart/ 301
 /guides/app/features/ /guides/models/ 301
-/guides/app/features/app/features/panels/parallel-coordinates/ /guides/app/features/app/features/panels/parallel-coordinates/ 301
 /guides/app/features/panels/compare-metrics/ /guides/app/features/panels/line-plot#compare-two-metrics-on-one-chart/ 301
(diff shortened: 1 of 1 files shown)


static/_redirects
```diff
@@ -85,7 +85,6 @@
 /frameworks/pytorch/ /guides/integrations/pytorch/ 301
 /getting-started/ /quickstart/ 301
 /guides/app/features/ /guides/models/ 301
-/guides/app/features/app/features/panels/parallel-coordinates/ /guides/app/features/app/features/panels/parallel-coordinates/ 301
```
ngrayluna, line 88: How did these repeated ones get added in the first place? 🫨

ngrayluna (approved):

PR #1658, opened 2025-09-17
Update experiments.md

knisar:
## Description

Another customer feedback - Removing the need to explicitly call wandb-core as the new sdk automatically uses it

📄 **[View preview links for changed pages](https://github.com/wandb/docs/pull/1658#issuecomment-3304341386)**


knisar:
The corresponding colab seem to be already updated.

ngrayluna (approved):

PR #1659, opened 2025-09-17
Add "Capabilities" section to Inference guide

jamie-rasmussen:
## Description

📄 **[View preview links for changed pages](https://github.com/wandb/docs/pull/1659#issuecomment-3304460063)**


dbrian57 (approved):
Pushed an update to this. I felt "Capabilities" was kinda vague, so I changed it to "Response Settings" which I feel is more descriptive of what we're documenting here.

I also rewrote descriptions to align with our style guide. Let me know if any of this feels misguided. Otherwise, I'm good with this.

Thanks, Jamie!

jamie-rasmussen:
> Pushed an update to this. I felt "Capabilities" was kinda vague, so I changed it to "Response Settings" which I feel is more descriptive of what we're documenting here.

It doesn't look like you actually renamed it in the UI. My only concern with "Response Settings" would be that it isn't generic enough if we want to add other things (like audio input perhaps) to the same section - but if you're OK with revisiting that later (and those things could e.g. go under a separate "Input Settings" section) it's fine with me.

PR #1668, opened 2025-09-20
BlueHawk: Adds bluehawk code snippets

ngrayluna:
## Description

Uses [BlueHawk CLI](https://mongodb-university.github.io/Bluehawk/code-snippets/#use-the-cli-to-extract-snippets) to parse a script into snippets using code snippets. Those snippets are then injected into markdown files using Hugo shortcode.

📄 **[View preview links for changed pages](https://github.com/wandb/docs/pull/1668#issuecomment-3314527689)**


johndmulhausen (changes requested):
This is great and has huge ramifications if we follow through on it. Great work!

One issue, though: I don't think we want the `source` directory to be here; rather, these snippet comments should be in the external example code repo, and we should script the cloning of that repo and generation of the snippets

PR #1673, opened 2025-09-23
Python SDK: Adds code example to box3d() function

ngrayluna:
## Description

Follow up to Python SDK revamp. While we wait for another solution, we are adding a code example to show how to use `box3d()` function.

## Status:

PR made on `wandb/wandb` needs to get approved first: https://github.com/wandb/wandb/pull/10555

## Related issues

- Jira ticket https://wandb.atlassian.net/browse/DOCS-1769

📄 **[View preview links for changed pages](https://github.com/wandb/docs/pull/1673#issuecomment-3322070449)**


johndmulhausen:
- Better docstring than "Returns a Box3D" please (the example code returns an Object3D)
- Maybe add a `score` to the example, as an attempt to get rid of the `null`s all over the generated boxes

PR #1694, opened 2025-10-06
Start publishing SDK release notes on docs.wandb.ai

mdlinville:
## Description

Closed in favor of #1854 

Start publishing SDK release notes on docs.wandb.ai.

1. Created a Python script to read the changelog upstream in wandb/wandb and create a new page with release notes for a given SDK release with some minor post-processing.
2. Ran the script to backfill all historical SDK release notes, and verified that the Hugo build succeeds and that the pages look good.
3. Created a Github action that parses the upstream SDK CHANGELOG.md to create release notes for the requested SDK version.
    1. Takes a SDK version string as its argument
    2. Fails early if the generated file already exists or the SDK version doesn't exist in the changelog
    3. Creates or reuses an existing branch using a naming pattern
    4. In the branch, generates and commits the new or updated page, pushes the branch or pushes into it
    5. If a PR exists for the branch, reuse it. Otherwise create a draft PR
4. Adjusted landing page IDs and cascading metadata
 
Ready for technical revieew

## Preview links
The HTML preview comment includes a lot of pages:
- Moved Server release notes, including archive
- New SDK release notes hierarchy
- Updates to the main landing page and Server release policy page

## Out of scope
- EOL policy for the SDK is in discussion
- Mapping between Server and SDK releases is out of scope

## Testing
I tested these changes:
- [x] Local build succeeds with all of the generated pages in place without errors or broken internal links (`hugo server`)
- [x] Python script works locally and correctly generates release notes for the given version 
- [x] Test the action in my fork
  - [x] Action successfully generates release notes for SDK 0.6 through 0.21
  - [x] Action successfully pushes a branch and creates a new PR
  - [x] In the PR created by the action, all checks succeed
  - [x] The action correctly fails if the generated page already exists
  - [x] The action correctly fails if the SDK version doesn't exist

You can review these changes:
- Check SDK release note Markdown source and HTML previews.
- Check Server release notes at their new home
- In the HTML preview, test that redirects from the old Server release notes location to the new one redirects correctly
- Verify that release notes in a given section and a given page sort correctly
- Verify the RSS feeds in the Release Notes landing page, which should include releases for both Server and SDK
- Verify the RSS feeds for Server and SDK landing pages

## Related issues
- Fixes DOCS-1709

📄 **[View preview links for changed pages](https://github.com/wandb/docs/pull/1694#issuecomment-3374515961)**

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -1111,8 +1111,19 @@
             "icon": "/icons/cropped-reports.svg",
             "pages": [
               "release-notes",
-              "release-notes/server-releases",
-              "release-notes/server-releases-archived",
+              {
+                "group": "W&B Server",
+                "pages": [
+                  "release-notes/server-releases",
+                  "release-notes/server-releases-archived"
+                ]
+              },
+              {
+                "group": "W&B SDK",
+                "pages": [
+                  "release-notes/sdk-releases"
+                ]
+              },
               "release-notes/release-policies"
             ]
           },
--- a/release-notes.mdx
+++ b/release-notes.mdx
@@ -3,20 +3,24 @@ title: Release Notes
 mode: wide
 ---
 
-We have two channels for release notes, each with their own RSS feeds.
+This section includes release notes for W&B Server and W&B SDK.
 
 <Columns cols={2}>
-<Card 
-  title="Release notes for supported W&B Server releases" 
-  href="./release-notes/server-releases"
+<Card
+  title="W&B Server"
   icon="rss"
-/>
+  href="/release-notes/server-releases"
+>
+  - [Supported releases](/release-notes/server-releases)
+  - [Archived releases](/release-notes/server-releases-archived)
+</Card>
 
 <Card 
-  title="Historical release notes for archived W&B Server releases" 
-  href="./release-notes/server-releases-archived"
-  icon="rss"
-/>
+  title="W&B SDK" 
+  href="./release-notes/sdk-releases"
+  icon="rss">
+  - [All releases](/release-notes/sdk-releases)
+</Card>
 </Columns>
 
-In addition, you can view [details about Weights & Biases' release policies](./release-notes/release-policies).
+<Tip>Learn about [W&B releases policies and processes](./release-notes/release-policies).</Tip>
--- a/release-notes/sdk-releases.mdx
+++ b/release-notes/sdk-releases.mdx
@@ -1,0 +1,6007 @@
+---
+title: "W&B SDK releases"
+sidebarTitle: "All releases"
+rss: true
+---
+
+This page includes release notes for the W&B Python SDK (`wandb` package). For W&B Server, see [W&B Server release notes](/release-notes/server-releases).
+
+<Update label="v0.22.2" description="October 07, 2025">
+
+View the [v0.22.2 changelog](https://github.com/wandb/wandb/releases/tag/v0.22.2) on GitHub.
+
+## Fixed
+
+- Possibly fixed some cases where the `output.log` file was not being uploaded.
+- Fixed excessive data uploads when calling `run.save()` repeatedly on unchanged files.
+
+</Update>
+
+<Update label="v0.22.1" description="September 29, 2025">
+
+View the [v0.22.1 changelog](https://github.com/wandb/wandb/releases/tag/v0.22.1) on GitHub.
+
+## Added
+
+- Optimize artifacts downloads re-verification with checksum caching.
+- Lazy loading support for `Api().runs()` to improve performance when listing runs. The new `lazy=True` parameter (default) loads only essential metadata initially, with automatic on-demand loading of heavy fields like config and summary when accessed.
+- Add `storage_region` option when creating artifacts. Users can use [CoreWeave AI Object Storage](https://docs.coreweave.com/docs/products/storage/object-storage) by specifying `wandb.Artifact(storage_region="coreweave-us")` when using wandb.ai for faster artifact upload/download on CoreWeave's infrastructure.
+
+
+## Fixed
+
+- `Api.artifact_exists()` and `Api.artifact_collection_exists()` now raise on encountering timeout errors, rather than (potentially erroneously) returning `False`.
+
+</Update>
+
+<Update label="v0.22.0" description="September 18, 2025">
+
+View the [v0.22.0 changelog](https://github.com/wandb/wandb/releases/tag/v0.22.0) on GitHub.
+
+## Notable Changes
+
+This version removes support of MacOS 10.
+
+## Removed
+
+- Remove a build targeting MacOS 10.x due to multiple security and supply chain considerations.
+
+## Fixed
+
+- Resuming a run with a different active run will now raise an error unless you call `run.finish()` first, or call `wandb.init()` with the parameter `reinit='create_new'`.
+- Fix `Api().runs()` for wandb server &lt;  0.51.0 (when `project.internalId` was added to gql API).
+- Sweeps: `command` run scripts that `import readline` whether directly or indirectly (e.g. `import torch` on Python 3.13) should no longer deadlock.
+
+</Update>
+
+{/* 
+<Update label="v0.21.4" description="September 11, 2025">
+
+View the [v0.21.4 changelog](https://github.com/wandb/wandb/releases/tag/v0.21.4) on GitHub.
+
+## Added
+
+- Add DSPy integration: track evaluation metrics over time, log predictions and program signature evolution to W&B Tables, and save DSPy programs as W&B Artifacts (complete program or state as JSON/PKL).
+
+</Update>
+
+<Update label="v0.21.3" description="August 30, 2025">
+
+View the [v0.21.3 changelog](https://github.com/wandb/wandb/releases/tag/v0.21.3) on GitHub.
+
+## Changed
+
+- Updated `click` dependency constraint from `>=7.1` to `>=8.0.1`.
+
+## Fixed
+
+- The message "Changes to your wandb environment variables will be ignored" is no longer printed when nothing changed.
+
+</Update>
+
+<Update label="v0.21.2" description="August 28, 2025">
+
+View the [v0.21.2 changelog](https://github.com/wandb/wandb/releases/tag/v0.21.2) on GitHub.
+
+## Notable Changes
+
+This version raises errors that would previously have been suppressed during calls to `Artifact.link()` or `Run.link_artifact()`. While this prevents undetected failures in those methods, it is also a breaking change.
+
+## Added
+
+- New settings for `max_end_of_run_history_metrics` and `max_end_of_run_summary_metrics`.
+- New `wandb.integration.weave` module for automatically initializing Weave when a W&B run is active and `weave` is imported.
+
+## Changed
+
+- Errors encountered while linking an artifact are no longer suppressed/silenced, and `Artifact.link()` and `Run.link_artifact()` no longer return `None`.
+- The "Run history" and "Run summary" printed at the end of a run are now limited to 10 metrics each.
+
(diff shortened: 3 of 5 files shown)


scripts/sdk-changelog-to-hugo-combined.py
timoffex, line 1: This file is quite large; you should probably factor parts of it out into multiple more focused files.

Although this is a script, I recommend using type annotations so that IDEs can give you suggestions and warnings during development.

timoffex:
I've only ever seen release notes sorted latest-first; it's generally a more useful order. So in https://docs-1709.docodile.pages.dev/ref/release-notes/sdk/0.21/ for instance, 0.21.4 should be at the top, followed by 0.21.3 and so on.

The text transformation code looks brittle and doesn't catch everything, like `Add pydantic>=2.6,<3 as a dependency (@dmitryduev in https://github.com/wandb/wandb/pull/8649 & estellazx. in https://github.com/wandb/wandb/pull/8905)` in 0.19.0 and a lot of examples in 0.17.x where we didn't use parentheses around the `by <user> in <pr>` part. Without a format linter in the SDK codebase, you'll probably have to keep updating the script, and it'll always require human review to make sure the extracted markdown is correct.

I wonder if all this is worth the complexity and maintenance effort. Why not link to `CHANGELOG.md`? Or you could include it but apply very minimal transformations (just extract the sections, maybe change heading levels, but don't parse the bullet points) and fix any issues directly in the source file.

mdlinville:
> I've only ever seen release notes sorted latest-first; it's generally a more useful order. So in https://docs-1709.docodile.pages.dev/ref/release-notes/sdk/0.21/ for instance, 0.21.4 should be at the top, followed by 0.21.3 and so on.
> 

It's an interesting point about the sort order. I wondered what folks would think. In the Server release notes, we don't do it that way, because the major release has major stuff in it, and the patch is much smaller. Even if you are skipping straight to the patch, you still need to read the .0 release notes to understand what changed in this release line. With the SDK having so many releases, it's overwhelming already, and I thought the extra cognitive load of needing to read it effectively from bottom to top to understand the full scope of your upgrade, was a lot. But I suspect in Mintlify, it will automatically be sorted this way. FWIW, the RSS feed is always sorted the way you want, because they are sorted by the `date` field in the front matter, which gets updated for each patch.

This page suffers from a heading misnumbering bug too. I'll get that fixed.

> The text transformation code looks brittle and doesn't catch everything, like `Add pydantic>=2.6,<3 as a dependency (@dmitryduev in https://github.com/wandb/wandb/pull/8649 & estellazx. in https://github.com/wandb/wandb/pull/8905)` in 0.19.0 and a lot of examples in 0.17.x where we didn't use parentheses around the `by <user> in <pr>` part. Without a format linter in the SDK codebase, you'll probably have to keep updating the script, and it'll always require human review to make sure the extracted markdown is correct.
> 
> I wonder if all this is worth the complexity and maintenance effort. Why not link to `CHANGELOG.md`? Or you could include it but apply very minimal transformations (just extract the sections, maybe change heading levels, but don't parse the bullet points) and fix any issues directly in the source file.

First, we never need to generate the old release notes again after we do it once. Y'all don't edit them, and we are committing the generated pages to the repo. So if the transforms are brittle for those old releases, I can fix it manually. Thanks for pointing them out. But for the last many many releases, y'all have used a pretty standard format, and the docs PR author (me) gets a chance to make adjustments to the generated content before pushing Publish.

Next, I don't think it's appropriate to expose W&B employee details (or in general, PR links) in the published release notes. I can trim them off completely, or we could consider trimming them from the upstream changelog format. If you have the PR link in the changelog, curious folks can follow it to discover the author. And I can find and trim it off much more easily.

Also, I want the contents of the release notes to be in docs.wandb.ai, not only upstream in Github. This adds them to the search corpus and has all of the impact of being on our own site, rather than on Github. When we send docs.wandb.ai readers somewhere else, including Github, we may not get them back. And I don't think readers will find the commit prefixes useful, nor the convention of lowercasing everything. I can make the transform less brittle, and we can also agree to a format upstream to make this more repeatable downstream. For example, it would be useful for me if the upstream changelog didn't include the committer and PR link.

mdlinville:
I addressed this feedback in the script and the Github action, then regenerated all of the pages using the current logic. PTAL.

> I've only ever seen release notes sorted latest-first; it's generally a more useful order. So in https://docs-1709.docodile.pages.dev/ref/release-notes/sdk/0.21/ for instance, 0.21.4 should be at the top, followed by 0.21.3 and so on.
> 

Done! This actually allowed me to remove the latest-patch notice at the top and the description field from the frontmatter. Adjusted the script and the Github action.

> The text transformation code looks brittle and doesn't catch everything, like `Add pydantic>=2.6,<3 as a dependency (@dmitryduev in https://github.com/wandb/wandb/pull/8649 & estellazx. in https://github.com/wandb/wandb/pull/8905)` in 0.19.0 and a lot of examples in 0.17.x where we didn't use parentheses around the `by <user> in <pr>` part. Without a format linter in the SDK codebase, you'll probably have to keep updating the script, and it'll always require human review to make sure the extracted markdown is correct.
> 

The specific note is in http://localhost:1313/ref/release-notes/sdk/0.19/#v0190 (and I think it was before, also). Simplified the logic and now just drop the contributor details instead of trying to comment them.

Any minor defects in the transform will be caught and corrected when the tech writer reviews the PR that results from running the Github action to generate the release notes. This is a one-time action, they are not all regenerated each time, and the results of the action are not blindly committed. If the defects persist, the script can be adjusted.

timoffex:
> For example, it would be useful for me if the upstream changelog didn't include the committer and PR link.

CC @wandb/sdk-team 

The PR links often include important context like additional details, links to other PRs or issues, repro scripts, etc. They come in very handy when trying to understand why an issue might not have been fixed.

The committer we could maybe remove. This was there to give credit to contributors (especially external ones), but it is visible in other ways on GitHub, and one can find this info by clicking through on a PR.

timoffex:
Oof, I didn't publish my review comments yesterday 🤦‍♂️ 🤦‍♂️  I reviewed the non-combined file at the time. I'll move the comments to the combined file and post them.

scripts/sdk-changelog-to-hugo-combined.py
```diff
@@ -0,0 +1,612 @@
...
+    r'^docs\([^)]+\):\s*',
+    r'^style\([^)]+\):\s*',
+    r'^refactor\([^)]+\):\s*',
+    r'^perf\([^)]+\):\s*',
+    r'^test\([^)]+\):\s*',
+    r'^build\([^)]+\):\s*',
+    r'^ci\([^)]+\):\s*',
+    r'^chore\([^)]+\):\s*',
+    r'^revert\([^)]+\):\s*',
+    # Without scope
```
timoffex: You don't need to have a scoped and unscoped entry for each prefix because regex can handle this with an optional group `(...)?`. For maintainability, it would be best to define just the conventional commit tags in a list and then construct the regular expressions from them:

```python
CONVENTIONAL_COMMIT_TYPES = (
    "feat",
    "fix",
    ...,
    "revert",
)

CONVENTIONAL_COMMIT_PREFIX_RE = re.compile(
    "|".join(rf"^{type}(\([^)]+\))?:\s*" for type in CONVENTIONAL_COMMIT_TYPES),
)
```

That last expression is still pretty gross (the closing parenthesis inside the brackets confuses the eyes) so I'd just put the whole thing in a function and use some variables instead:

```python
def _remove_conventional_commit_prefix(title: str) -> str:
    types = (
        "feat",
        "fix",
        ..., 
    )

    scope_re = "\([^)]+\)"
    prefix_re = "|".join(rf"^{type}({scope_re})?:\s*" for type in types)

    return re.sub(prefix_re, '', title, flags=re.IGNORECASE)
```

scripts/sdk-changelog-to-hugo-combined.py
```diff
@@ -0,0 +1,612 @@
...
+    Returns:
+        list: List of release dictionaries.
+    """
+    with open(filepath, 'r') as f:
+        content = f.read()
+    
+    # Three patterns: new format (## [version] - date), old format (# version (date)), and older format (## version (date))
+    new_release_pattern = r'^## \[([^\]]+)\] - (.+?)$'
+    old_release_pattern = r'^# ([\d\.]+) \((.+?)\)$'
+    older_release_pattern = r'^## ([\d\.]+) \((.+?)\)$'
```
timoffex: For the semver version, a less general but more readable regex could work: `\[\d+\.\d+\.\d+\]`. Or even `\[\d+(\.\d+)*\]` (one or more groups of one or more digits separated by periods).

I don't like `[^\]]`

scripts/sdk-changelog-to-hugo-combined.py
```diff
@@ -0,0 +1,612 @@
...
+                formatted_date = date_str
+                description = date_str
+            
+            current_release = {
+                'version': version,
+                'title': f'[{version}] - {date_str}',
+                'date': formatted_date,
+                'description': description,
+                'content': ''
+            }
```
timoffex: Use a dataclass for maintainability:

```python
import dataclasses

@dataclasses.dataclass()
class ReleaseEntry:
    lines: list[str]
    version: str
    date: str
    title: str  # not sure you need this one, could be a property


current_release = ReleaseEntry(
    lines=[],
    version=version,
    date=date_str,
)

current_release.content = ...
```

scripts/sdk-changelog-to-hugo-combined.py
```diff
@@ -0,0 +1,612 @@
...
+            # Look for patterns like:
+            # - (@username in https://...)
+            # - (by @username in https://...)
+            # - (@user1, @user2 in https://...)
+            # - by @username in https://...
+            # And variations with 'and' between users
+            
+            # First, extract the base content (everything before the PR info)
+            # PR info typically starts with (@, (by @, or by @
+            pr_pattern = r'[\s]*(?:\((?:by\s+)?@[^)]+\)|by\s+@[^\s]+\s+(?:and\s+@[^\s]+\s+)?in\s+https://[^\s\)]+)[\s\.]*$'
```
timoffex: You should use the [re.VERBOSE](https://docs.python.org/3/library/re.html#re.VERBOSE) flag and add inline comments. Maybe even break this up into parts.

I can't read this regex, but here's my from-scratch attempt:

```python
user_re = r"@[\w\.-]+\b"
pr_link_re = r"\bhttps://.+\b"
attribution_re = rf"""
    \(? # optional '('
      (                          # One or more of:
        (\bby\b \s+)? {user_re}  # '@user' or 'by @user'
        | (in \s+)? {pr_link_re} # 'in URL' or 'URL'
        | \band\b | &            # 'and' or '&'
        | \s+                    # a sequence of spaces
      )+
    \)? # optional ')'
    $   # end of line
"""
```

scripts/sdk-changelog-to-hugo-combined.py
```diff
@@ -0,0 +1,612 @@
...
+    
+    # Save last release
+    if current_release:
+        current_release['content'] = '\n'.join(current_content).strip()
+        releases.append(current_release)
+    
+    return releases
+
+
+def transform_content(content):
```
timoffex: This function will be hard to maintain. It's long and has a high cyclomatic complexity because of the nested conditionals (25, I checked using [radon](https://pypi.org/project/radon/)), even though it's well organized and has helpful comments.

A small improvement you could make is factor out the "Transform list items" and "Adjust heading levels" portions into helper functions. They could return a list of lines to append to `transformed_lines`; "Transform list items" would return an empty list for "docs(sdk)" items.

timoffex:
But overall LGTM to the general idea. I'll let someone from the docs team stamp since maintenance will fall on your team.

PR #1695, opened 2025-10-07
Update reference docs to sdk@v0.22.2, core@master

davidwallacejackson:
Used https://github.com/wandb/wandb/commit/9e9276d04a6e525631c3fbcb566eace3a7fe742c

📄 **[View preview links for changed pages](https://github.com/wandb/docs/pull/1695#issuecomment-3378521390)**


johndmulhausen:
Thanks for this PR! But we have to close it because we switched docs platforms. @ngrayluna will be following up with a new script that generates 0.22.2 docs for Mintlify.

PR #1707, opened 2025-10-13
FREEZE EXCEPTION: Inference and Serverless RL Model name corrections

dbrian57:
## Description
Updates:
* Inference and Serverless RL model pages with proper Qwen3 model name 
* Removes Qwen2.5 from Serverless RL page


mdlinville (approved):
LGTM, please also replicate in `mintlifytest`. :)

PR #1716, opened 2025-10-16
Fix import issues

mdlinville:
## Description
Fix some one-off issues after the migration

- Fix Models quickstart location
- Fix inadvertent lowercasing of Python API and Weave API refs by the vendor during the conversion
- Bring in missed Weave reference generation scripts, update them to preserve the casing in the Weave repo

Verified that all other added, renamed, updated, deleted files in `main` commits since 9/17 look correct, both for file location and content changes.

The commit history in this PR is a bit of a mess. It's difficult to do case-sensitive Git commands on macOS, which has a case-insensitive filesystem.


ngrayluna:
Should the `models_quickstart.mdx` be moved to the root of the directory? Not in `/guides` since we are no longer using that?

ngrayluna:
Did the Mintlify folks change the file name casing?

mdlinville:
> Should the `models_quickstart.mdx` be moved to the root of the directory? Not in `/guides` since we are no longer using that?

We still are, but maybe that's a mistake also. This content came in through `mintlifytest`, I think. In this PR's branch: https://github.com/wandb/docs/tree/4a224d736762310ed5c70f207a42e892f18a4289/guides 

Maybe @johndmulhausen can help answer this.

scripts/reference-generation/generate_python_sdk_minimal.py
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env python3
+"""
+Minimal Python SDK documentation generator for Mintlify.
+This avoids lazydocs and its dependencies to prevent license policy violations.
```
ngrayluna: Curious as to what license policy violations these are referring to.

mdlinville:
> > Should the `models_quickstart.mdx` be moved to the root of the directory? Not in `/guides` since we are no longer using that?
> 
> We still are, but maybe that's a mistake also. This content came in through `mintlifytest`, I think. In this PR's branch: https://github.com/wandb/docs/tree/4a224d736762310ed5c70f207a42e892f18a4289/guides
> 
> Maybe @johndmulhausen can help answer this.

Addressed in #1719

PR #1730, opened 2025-10-20
Fix typos

jk4e:


mdlinville:
Thank you for this PR, but I'm afraid that it has several conflicts that arose after our migration to Mintlify from Hugo last week. Because of the magnitude of the conflicts when compared with the size of the changes, I'll close this for now. Please open a new PR if you would still like to pursue these typos.

PR #1752, opened 2025-10-21
Fix broken Weave reference links

mdlinville:
## 
- Fixed Models_and_Weave_Integration_Demo links to point to /weave/cookbooks/ instead of non-existent /weave/reference/gen_notebooks/
- Fix Weave Service API reference links to use correct OpenAPI paths
  - Updated links to use the kebab-case operationId pattern that Mintlify generates from the OpenAPI spec (e.g., /weave/reference/service-api/call-start-call-start-post)
  - These links work on the live site (docs.wandb.ai) where Mintlify generates the API pages from the OpenAPI spec
  - Links will show as broken in 'mint broken-links' locally because the OpenAPI pages are dynamically generated at build time, not stored as static files
  - Verified the URLs are correct by checking against the live docs site

## Testing
- [x] Local build succeeds without errors or broken internal links
- [ ] PR tests succeed

### DIFF
--- a/weave/cookbooks/online_monitoring.mdx
+++ b/weave/cookbooks/online_monitoring.mdx
@@ -232,9 +232,9 @@ In this cookbook, we demonstrated how to create a custom production monitoring d
 
 - **Data Input:**
   - Framework-agnostic tracing with [@weave-op()](/weave/quickstart#2-log-a-trace-to-a-new-project) decorator and the possibility to import calls from CSV (see related [import cookbook](/weave/cookbooks/import_from_csv))
-  - Service API endpoints to log to Weave from for various programming frameworks and languages, see [here](/api-reference/calls/call-start) for more details.
+  - Service API endpoints to log to Weave from for various programming frameworks and languages, see [here](/weave/reference/service-api/call-start-call-start-post) for more details.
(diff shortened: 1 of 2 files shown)


weave/cookbooks/online_monitoring.mdx
```diff
@@ -232,9 +232,9 @@ In this cookbook, we demonstrated how to create a custom production monitoring d
 
 - **Data Input:**
   - Framework-agnostic tracing with [@weave-op()](/weave/quickstart#2-log-a-trace-to-a-new-project) decorator and the possibility to import calls from CSV (see related [import cookbook](/weave/cookbooks/import_from_csv))
-  - Service API endpoints to log to Weave from for various programming frameworks and languages, see [here](/api-reference/calls/call-start) for more details.
+  - Service API endpoints to log to Weave from for various programming frameworks and languages, see [here](/weave/reference/service-api/call-start-call-start-post) for more details.
```
mdlinville:
```suggestion
  - Service API endpoints to log to Weave from for various programming frameworks and languages. See [`https://trace.wandb.ai/call/start`](/weave/reference/service-api/call-start) for more details.
```

PR #1775, opened 2025-10-24
Updates models for Training with new LLama model

dbrian57:
## Description
Adds a new Llama model to Training's available models page.


mdlinville (approved):

dbrian57:
Apparently this was not needed.

PR #1780, opened 2025-10-24
[WEAVE] OTEL attribute mapping table and threads example

dbrian57:
## Description
Resolves [DOCS-1825](https://wandb.atlassian.net/browse/DOCS-1825), [DOCS-1899](https://wandb.atlassian.net/browse/DOCS-1899), and [DOCS-1914](https://wandb.atlassian.net/browse/DOCS-1914).

This PR adds:

* A recommendation to use the `BatchSpanProcessor()` instead of the `SimpleBathcProcessor()`
* A multi-turn thread example (created and vetted by Z. Birenbaum)
* A table of OTEL spans

Adds a new table that explains shows how we normalize OTEL data in Weave.

[DOCS-1825]: https://wandb.atlassian.net/browse/DOCS-1825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DOCS-1899]: https://wandb.atlassian.net/browse/DOCS-1899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DOCS-1914]: https://wandb.atlassian.net/browse/DOCS-1914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zbirenbaum:
This table looks pretty good. We should have a guide on how to use threads with OTEL though. I'm fine with working it in here or including in a followup

I think that including the vendors for each key is unnecessary. Maybe list the ones we support (Vercel, OpenInference, etc) in a separate block

weave/guides/tracking/otel.mdx
```diff
@@ -46,14 +46,18 @@ First, install the required dependencies:
 pip install openai openinference-instrumentation-openai opentelemetry-exporter-otlp-proto-http
 ```
 
+<Warning>
+**Performance Recommendation**: Always use `BatchSpanProcessor` instead of `SimpleSpanProcessor` when sending traces to Weave. `SimpleSpanProcessor` exports spans synchronously and can cause performance issues. `BatchSpanProcessor` batches spans asynchronously, which is more efficient and recommended for production use.
```
mdlinville:
```suggestion
**Performance Recommendation**: Always use `BatchSpanProcessor` instead of `SimpleSpanProcessor` when sending traces to Weave. `SimpleSpanProcessor` exports spans synchronously, potentially impacting the performance of other workloads. These examples illustrate `BatchSpanProcessor`, which is recommended in production because it batches spans asynchronously and efficiently.
```
Totally optional and maybe not yet perfect. I wanted to tighten up the recommendation a bit. I'm tempted to remove "efficiently" entirely, because who knows?

weave/guides/tracking/otel.mdx
```diff
@@ -272,3 +276,291 @@ python opentelemetry_example.py
 ```
 
 The span attribute prefixes `gen_ai` and `openinference` are used to determine which convention to use, if any, when interpreting the trace. If neither key is detected, then all span attributes are visible in the trace view. The full span is available in the side panel when you select a trace.
+
+## Organizing OTEL traces into threads
```
mdlinville:
```suggestion
## Organize OTEL traces into threads
```
I noticed that this page uses Title Case for many section headings. Maybe suggests a later refactor project across the docs.

weave/guides/tracking/otel.mdx
```diff
@@ -272,3 +276,291 @@ python opentelemetry_example.py
...
+
+You can organize your OpenTelemetry traces into [Weave threads](/weave/guides/tracking/threads) by adding specific span attributes. This allows you to group related operations (like multi-turn conversations or user sessions) and visualize them in Weave's thread UI.
```
mdlinville:
```suggestion
Add specific span attributes to organize your OpenTelemetry traces into [Weave threads](/weave/guides/tracking/threads), then use Weave's Thread UI to analyze related operations like multi-turn conversations or user sessions in Weave's thread UI.
```
Maybe link to the thread UI? Do the Weave docs say UI or App?

weave/guides/tracking/otel.mdx
```diff
@@ -272,3 +276,291 @@ python opentelemetry_example.py
...
+
+Add the following attributes to your OTEL spans to enable thread grouping:
+
+- `wandb.thread_id`: Groups spans into a specific thread
+- `wandb.is_turn`: Marks a span as a conversation turn (appears as a row in the thread view)
+
+The following code shows four examples of how to organize OTEL traces into Weave threads using `wandb.thread_id` (to group related operations) and `wandb.is_turn` (to mark high-level operations that appear as rows in the thread view). The examples perform the following:
```
mdlinville:
```suggestion
The following code shows several examples of organizing OTEL traces into Weave threads. They use `wandb.thread_id` to group related operations,  and  use `wandb.is_turn` to view high level operations as rows in the thread view. Each example performs the followingmark high-level operations that appear as rows in the thread view). 
```

weave/guides/tracking/otel.mdx
```diff
@@ -272,3 +276,291 @@ python opentelemetry_example.py
...
+
+* **Example 1**: Traces a basic single-turn thread
```
mdlinville: Super optional but I think this would look better as a little table, then maybe you could get away without "Traces" in each one They need periods.

weave/guides/tracking/otel.mdx
```diff
@@ -272,3 +276,291 @@ python opentelemetry_example.py
...
+* **Example 2**: Traces a multi-turn conversation sharing one thread ID
+* **Example 3**: Traces deeply nested operations where only the outermost span is marked as a turn
```
mdlinville:
```suggestion
* **Example 3**: Traces deeply nested operations and marks only the outermost span as a turn.
```

weave/guides/tracking/otel.mdx
```diff
@@ -272,3 +276,291 @@ python opentelemetry_example.py
...
+* **Example 4**: Traces background operations that belong to a thread but aren't turns
+
+<Accordion title="Click to open the code example">
```
mdlinville: Would it be cute to accordion each example separately? Then you could link the above.

weave/guides/tracking/otel.mdx
```diff
@@ -272,3 +276,291 @@ python opentelemetry_example.py
...
+```
+</Accordion>
+
+After sending these traces, you can view them in the Weave UI under the **Threads** tab, where they'll be grouped by `thread_id` and each turn will appear as a separate row.
+
+## Attribute Mappings
+
+Weave automatically maps OpenTelemetry span attributes from various instrumentation frameworks to its internal data model. When multiple attribute names map to the same field, Weave applies them in priority order, allowing frameworks to coexist in the same traces.
+
+### Supported Frameworks
```
mdlinville, line 507: I think if this and the next one really belong in this page, it might be good to have an H2 called Appendix or something to contain them. It might be better to eventually split them out to a reference page.

mdlinville (approved):
Some small suggestions and ideas for future followup. Nice work.

weave/guides/tracking/otel.mdx
dbrian57: oh yeah, totally. I should have caught that.

PR #1784, opened 2025-10-25
Runs: Remove mention of immutable run IDs

ngrayluna:
## Description

From Jira ticket:

"""
We shouldn’t be using the “immutable run ID” reference anymore in the application. 

Currently, this is referenced in forking a run documentation here: https://docs.wandb.ai/models/runs/forking#using-an-immutable-run-id 

We could just remove the entire section “Using an immutable run ID” section.
"""

## Related issues

- Fixes https://wandb.atlassian.net/browse/DOCS-1855

### DIFF
--- a/models/runs/forking.mdx
+++ b/models/runs/forking.mdx
@@ -37,16 +37,6 @@ forked_run = wandb.init(
 )
 ```
 
-### Using an immutable run ID
-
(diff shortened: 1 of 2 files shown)


ktmathews89 (approved):
thank you noah!

PR #1798, opened 2025-10-29
[WEAVE] New limits page

dbrian57:
Resolves [DOCS-1244](https://wandb.atlassian.net/browse/DOCS-1244). Documents Weave retry behavior, but also documents a few other behaviors and limitation in a new Limits and Expected Behaviors doc. Eventually this doc will become part of a new "Details" section, but I've placed it in the Tools and Resources section for now.

[DOCS-1244]: https://wandb.atlassian.net/browse/DOCS-1244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/guides/tools/limits.mdx
+++ b/weave/guides/tools/limits.mdx
@@ -18,19 +18,37 @@ description: A list of Weave's limitations, known issues, and expected behaviors
 
     ```python
     import weave
-
+    
     class MyObj:
+        """An object with a large string attribute."""
         def __init__(self, x: int):
             self.x = x
-
+    
         def __repr__(self):
             return f"MyObj(x={self.x})"
-
+    
         def to_dict(self):
             return {"x": self.x}
-
+    
     @weave.op()
     def make_my_obj():
-        x = "s" * 10_000
+        x = "a" * 10_000
         return MyObj(x)
+    
+    def main():
+        weave.init("<entity/project>")
+    
+        # Define a traceable operation that processes MyObj
+        @weave.op()
+        def process_obj(obj: MyObj) -> int:
+            return len(obj.x)
+    
+        # Create and process a large MyObj instance
+        large_obj = make_my_obj()
+        length = process_obj(large_obj)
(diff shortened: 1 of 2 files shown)


weave/guides/tools/limits.mdx
```diff
@@ -0,0 +1,36 @@
...
+    ```python showLineNumbers
+    # This raises exceptions
+    result, call = foo.call(__should_raise=True)
+    ```
+
+* Dedicated Weave instances use a different OpenTelemetry ingress URL. See [Send OpenTelemetry Traces](/weave/guides/tracking/otel) for authoritative endpoint information.
+
+* Weave sometimes truncates large trace data objects. This occurs because default trace output is a raw, custom Python object that Weave doesn’t know how to serialize. To return all of your trace data, define a dictionary of strings, like this: 
+
+    ```python
```
ngrayluna, line 19: Suggestion: Adding more to this code might make it even more clear how this dictionary behaves in action. Not a blocker so approving.

```python
import weave

class MyObj:
    """An object with a large string attribute."""
    def __init__(self, x: int):
        self.x = x

    def __repr__(self):
        return f"MyObj(x={self.x})"

    def to_dict(self):
        return {"x": self.x}

@weave.op()
def make_my_obj():
    x = "a" * 10_000
    return MyObj(x)

def main():
    weave.init("<entity/project>")

    # Define a traceable operation that processes MyObj
    @weave.op()
    def process_obj(obj: MyObj) -> int:
        return len(obj.x)

    # Create and process a large MyObj instance
    large_obj = make_my_obj()
    length = process_obj(large_obj)
    print(f"Length of x in MyObj: {length}")
    print("dict:", large_obj.to_dict())

if __name__ == "__main__":
    main()
```

ngrayluna (approved):

PR #1806, opened 2025-10-30
[WEAVE] Update titles to be task-based

dbrian57:
## Description
Updates the titles in Weaves documentation to use task-based descriptive language.

### DIFF
--- a/weave/guides/core-types/models.mdx
+++ b/weave/guides/core-types/models.mdx
@@ -1,6 +1,6 @@
 ---
-title: "Version your application with models"
-description: "Version your application with structured models that combine data and code."
+title: "Track application versions with models"
+description: "Track versions of your application with structured models that combine data and code."
 ---
 
 A `Model` is a combination of data (which can include configuration, trained model weights, or other information) and code that defines how the model operates. By structuring your code to be compatible with this API, you benefit from a structured way to version your application so you can more systematically keep track of your experiments.
--- a/weave/guides/core-types/models.mdx
+++ b/weave/guides/core-types/models.mdx
@@ -1,6 +1,6 @@
 ---
-title: "Models"
-description: "Version your application with structured models that combine data and code"
+title: "Version your application with models"
+description: "Version your application with structured models that combine data and code."
 ---
 
 A `Model` is a combination of data (which can include configuration, trained model weights, or other information) and code that defines how the model operates. By structuring your code to be compatible with this API, you benefit from a structured way to version your application so you can more systematically keep track of your experiments.
--- a/weave/guides/core-types/prompts.mdx
+++ b/weave/guides/core-types/prompts.mdx
@@ -1,6 +1,6 @@
 ---
-title: "Store and version prompts"
-description: "Store, version, and organize prompts for LLM applications"
+title: "Store and track versions of prompts"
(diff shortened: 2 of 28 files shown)


weave/guides/core-types/models.mdx
```diff
@@ -1,6 +1,6 @@
 ---
-title: "Models"
-description: "Version your application with structured models that combine data and code"
+title: "Version your application with models"
```
mdlinville: I don't love Version as a verb

weave/guides/core-types/prompts.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: Prompts
+title: "Store and version prompts"
```
mdlinville: Ditto

weave/guides/platform/weave-self-managed.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: "W&B Weave Self-Managed"
+title: "Set up a self-managed Weave instance"
 description: "Deploy and manage Weave on your own infrastructure"
```
mdlinville, line 3: I suspect that you need W&B in the description

weave/guides/tools/deploy.mdx
```diff
@@ -9,18 +9,14 @@ description: "Deploy your Weave models to cloud platforms"
...
 
-Given a Weave ref to any Weave Model you can run:
+To deploy a Weave model to Google Cloud, run:
 
 ```
 weave deploy gcp <ref>
 ```
 
-to deploy a gcp cloud function that serves your model. The last line of the deployment will look like `Service URL: <PATH_TO_MODEL>`. Visit `<PATH_TO_MODEL>/docs` to interact with your model!
+Once deployed, the terminal returns a URL using the following schema:
```
mdlinville, line 18: Are we expecting the user to parse the URLs for the model and the docs from the output? This seems like it could be brittle.

weave/guides/tools/weave-in-workspaces.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: Weave in Workspaces
+title: Use Weave in workspaces
```
mdlinville: Would it be better to say "in the W&B App" or something like that?

weave/guides/tracking/video.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Video Support"
+title: "Log videos"
```
mdlinville, line 2: I note that some titles are quoted and some are not

mdlinville (approved):
Some minor feedback and questions to consider but I will go ahead and approve.

weave/guides/core-types/models.mdx
dbrian57: Totally fair. I'll go with something else

weave/guides/tools/deploy.mdx
dbrian57: yeah, this is a good question. I rewrote this as just a minor wording cleanup because it looked unprofessional, but the workflow needs work, too. Ultimately, I assume people are gonna want to pull this info out of the resulting JSON object and we should note that here.

PR #1813, opened 2025-10-31
Update outdated screenshots

johndmulhausen:
What it says on the tin. 

Fixes DOCS-1362


images/app_ui/delete_project.png
mdlinville, line 1: I'd prefer this to be cropped to only the modal with maybe a small border of the background to clarify that it's a modal. It's hard to actually read the text in the screenshot

images/app_ui/metadata_panel.png
mdlinville, line 1: WHy so much larger?

mdlinville (approved):

mdlinville:
FYI, there are some broken links in the localized content from some images that this PR removes.

PR #1821, opened 2025-11-03
Sweeps: Update warning nesting admonition 

### LINKED ISSUE #1491: Incorrect example for nested parameters
URL: https://docs.wandb.ai/guides/sweeps/define-sweep-configuration/

The example (which is meant to illustrate overridding nested parameters) shows a sweep configuration dictionary being defined as:
```
sweep_configuration = {
    "top_level_param": 0,
    "nested_param": {
        "learning_rate": 0.01,
        "double_nested_param": {"x": 0.9, "y": 0.8},
    },
}
```

However that isn't a valid sweep configuration! That is the format that could be given as a *run* configuration.

I believe (from trying it in my own code) the correct *sweep* configuration (that would in turn create a run configuration looking like the above) is:
```
sweep_configuration = {
    # I've skipped other keys like "method", "metric", etc.
    "parameters": {
        "top_level_param": {"value": 0},
        "nested_param": {
            "parameters": {
                "learning_rate": {"value": 0.01},
                "double_nested_param": {
                    "parameters": {
                        "x": {"value": 0.9}, 
                        "y": {"value": 0.8},
                    },
                },
            },
        },
    },
}
```

I also think the instructions above the example are misleading, as they appear to say you can have a "distribution" or even a "value" for the "nested_param" itself, whereas I have found you can only have distributions and values within a named parameter (within the "parameters" within "nested_param").

ngrayluna:
## Description

Nested sweep config block code example is incorrect. This PR fixes it.

## Related issues

- Fixes https://wandb.atlassian.net/browse/DOCS-1696
- Fixes https://github.com/wandb/docs/issues/1491

### DIFF
--- a/models/sweeps/define-sweep-configuration.mdx
+++ b/models/sweeps/define-sweep-configuration.mdx
@@ -228,9 +228,8 @@ parameters:
 {/* For example, the proceeding code snippets show a sweep config both in a YAML config file and a Python script. */}
 
 
-
-When a sweep sets a nested parameter, W&B applies that subtree to the run config rather than deep-merging it with your defaults. As
-a result, keys present under the same parent in `wandb.init(config=...)` may not be preserved.
+<Warning>
+Nested parameters defined in sweep configuration overwrite keys specified in a W&B run configuration.
 
 As an example, suppose you have `train.py` script that initializes a run with a nested default:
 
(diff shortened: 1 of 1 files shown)


models/sweeps/define-sweep-configuration.mdx
```diff
@@ -74,52 +74,197 @@ Within the top level `parameters` key, the following keys are nested: `learning_
 
 ## Double nested parameters
 
-Sweep configurations support nested parameters. To delineate a nested parameter, use an additional `parameters` key under the top level parameter name. Sweep configs support multi-level nesting.
+Sweep configurations support nested parameters. To define a nested parameter, include an additional `parameters` key under the top-level parameter name.
```
mdlinville, line 77: In the code snippets, it looks like `parameters` is a top-level key. Maybe that's my misunderstanding.

mdlinville (approved):

PR #1828, opened 2025-11-04
Remove cards add buttons

ngrayluna:
## Description

Replaces `<Card>` with custom buttons.

## Related issues

- Fixes https://wandb.atlassian.net/browse/DOCS-1828


mdlinville (commented):
LGTM for the most part. When it's one of multiple buttons, the line wrapping seems inconsistent with other buttons. See https://wb-21fd5541-remove-cards-add-buttons.mintlify.app/models/artifacts 
Screenshot 2025-11-04 at 8 42 54 AM Do you think we need to make sure it doesn't wrap or do some other thing?

mdlinville (approved):

ngrayluna:
If you collapse it enough, they both do the same thing. 🤔

PR #1838, opened 2025-11-06
[DOCS-1887] Update link checker for Mintlify

mdlinville:
## Description
Fixes DOCS-1887

In Hugo, the link checker checked files on disk to discover links to check. This won't work in Mintlify, where content is served from memory only. This gives us an opportunity to simplify the link checking:

1. Rename existing action and use it to check the live site only. It will run monthly or on demand and creates a Github issue with the report only if broken links are found.
2. Add a separate action that will run in a PR. It will run when the PR branch updates, and will start only after the Mintlify preview deployment succeeds, similar to how the HTML preview works. It will fail the PR check if broken links are found and can be made required for a PR to merge.
3. Updates the .lycheeignore:
    - Consolidate ignored W&B prod URLs using regular expressions
    - Ignores URLs that start with `/` or do not start with `http`, which are internal links checked by `mint broken-links`
    - Reorganizes and categorizes remaining URLs for maintenance
4. Revert the logic I had previously added to try to auto-fix broken links and cut a PR, because it never worked correctly

To test the `.lycheeignore` and the basics locally, you could:
1. `brew install lychee`
2. In the `docs` clone, run `mint dev`.
3. Test the PR command. In the `docs` clone, run this command after adjusting the port if necessary.
    ```shell
    lychee --threads 5 --max-retries 5 --retry-wait-time 2 -f markdown --output=./lychee-report.md --include '^https?://' --include '^http?://' --base-url=http://localhost:3000 http://localhost:3000
    ```
    Replace the port if `mint dev` started on a different port. This is the same command that the PR will run, and will also pick up the `.lycheeignore`.

    You're done with `mint dev` and can kill it. Take a look at `lychee-report.md`.
4. Test the `main` command.  In the `docs` clone, run this command to check the live site:
    ```shell
    lychee --threads 5 --max-retries 5 --retry-wait-time 2 -f markdown --include '^https?://' --include '^http?://' --base-url=https://docs.wandb.ai https://docs.wandb.ai
    ```
    This doesn't save the output to a file.

## Testing
- [x] Locally, `lychee` works with the suggested test commands

Github actions are difficult to test ahead of merging.
- The simplified logic of the production check was working before I had updated it
- The PR check is even more simple because it doesn't need to save the report or write a comment in the PR.


.github/workflows/linkcheck-pr.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: "--threads 5 --max-retries 5 --retry-wait-time 2 -f markdown --output=./lychee-report.md --base-url='$PREVIEW_URL' '$PREVIEW_URL'"
+          fail: false
+        env:
+          # to be used in case rate limits are surpassed
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
dbrian57, line 32: Do we need to specifically call out the token here and then again above? I thought Git secret were automatically loaded and available for workflows.

dbrian57 (approved):
This is awesome! One question inline, but otherwise looks good.

.github/workflows/linkcheck-pr.yml
mdlinville: It looks like we can skip the `with` but not the `env`. I think the Lychee action does not assume that the token is in your environment. https://github.com/lycheeverse/lychee-action/blob/master/action.yml#L42

.github/workflows/linkcheck-pr.yml
```diff
@@ -0,0 +1,35 @@
...
+      contains(github.event.deployment_status.environment_url, 'mintlify'))
+    steps:
+      # check URLs with Lychee
+      - uses: actions/checkout@v4
+
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
mdlinville:
```suggestion
```

.github/workflows/linkcheck-prod.yml
```diff
@@ -0,0 +1,34 @@
...
+      issues: write # required for peter-evans/create-issue-from-file
+    steps:
+      # check URLs with Lychee
+      - uses: actions/checkout@v4
+
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
mdlinville:
```suggestion
```

PR #1875, opened 2025-11-13
Updated Weave onboarding notebooks

m-rgba:
## Description

Updated Weave's onboarding ipynb's for Traces and Evals to reflect the same examples as our Quickstart screens when users first create a new project.

These will be linked from our Quickstart page, as well as the empty spaces within the Traces and Evals pages.


weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -23,12 +12,11 @@
     "\n",
     "Weave is a toolkit for developing AI-powered applications.\n",
     "\n",
-    "You can use Weave to:\n",
-    "- Log and debug language model inputs, outputs, and traces.\n",
-    "- Build rigorous, apples-to-apples evaluations for language model use cases.\n",
-    "- Organize all the information generated across the LLM workflow, from experimentation to evaluations to production.\n",
+    "This notebook demonstrates how to evaluate a model or function using Weave’s Evaluation API. \n",
+    "\n",
+    "Evaluation is a core concept in Weave that helps you measure and iterate on your application by running it against a dataset of examples and scoring the outputs using custom-defined functions. \n",
```
mdlinville:
```suggestion
    "In Weave, you evaluate your application by running it against a dataset of examples and scoring the outputs using custom-defined functions. This helps you to measure and improve your application's performance.\n",
```

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -23,12 +12,11 @@
...
     "\n",
-    "This notebook demonstrates how to evaluate a model or function using Weave’s Evaluation API. Evaluation is a core concept in Weave that helps you measure and iterate on your application by running it against a dataset of examples and scoring the outputs using custom-defined functions. You'll define a simple model, create a labeled dataset, track scoring functions with `@weave.op`, and run an evaluation that automatically tracks results in the Weave UI. This forms the foundation for more advanced workflows like LLM fine-tuning, regression testing, or model comparison.\n",
+    "You'll define a simple model, create a labeled dataset, track scoring functions with `@weave.op`, and run an evaluation that automatically tracks results in the Weave UI. This forms the foundation for more advanced workflows like LLM fine-tuning, regression testing, or model comparison.\n",
```
mdlinville:
```suggestion
    "In this notebook, you define a simple model, create a labeled dataset, track scoring functions with `@weave.op`, run an evaluation, and review the results in the Weave UI.\nThis workflow forms the foundation for more advanced workflows like fine tuning an LLM model, detecting regressions, and comparing models.\n",
```

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -39,13 +27,22 @@
     "id": "McE7cuqSxMiP"
    },
    "source": [
-    "## 🔑 Prerequisites\n",
+    "## 🐝 Running your first evaluation\n",
```
mdlinville:
```suggestion
    "## Run your first evaluation\n",
```
We are trying hard to get rid of emojis in the docs

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -39,13 +27,22 @@
...
     "\n",
-    "Before you can run a Weave evaluation, complete the following prerequisites.\n",
+    "In this example, we're using W&B Inference. [Learn more](https://docs.wandb.ai/inference) about our inference API.\\\n",
```
mdlinville:
```suggestion
    "This example uses the [W&B Inference](https://docs.wandb.ai/inference)  API.\\\n",
```
Is the `\\\n` right? Looks funny.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -39,13 +27,22 @@
...
+    "Using another provider? [We support all major clients and frameworks](https://docs.wandb.ai/weave/guides/integrations).\n",
```
mdlinville:
```suggestion
    "Learn more about [Weave's supported client and framework integrations](https://docs.wandb.ai/weave/guides/integrations).\n",
```
It seems like this should be joined with the previous line.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -39,13 +27,22 @@
...
     "\n",
-    "1. Install the W&B Weave SDK and log in with your [API key](https://wandb.ai/settings#api).\n",
-    "2. Install the OpenAI SDK and log in with your [API key](https://platform.openai.com/api-keys).\n",
-    "3. Initialize your W&B project."
+    "We also have an OpenAI version of this example below."
```
mdlinville:
```suggestion
    "To use OpenAI, run [link to block](link-to-block) instead of this cell.
```
I'd suggest updating the title of this cell and the OpenAI cell to enforce the either/or, or maybe cells support tabs and we could show Inference and OpenAI tabs. I think that generally, we do the authorization and imports in a separate cell so that it only needs to be run once, then the user could try both Inference and OpenAI paths more easily, and more easily understand the specifics the example is trying to demonstrate.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -56,124 +53,187 @@
...
+    "# Find your wandb API key at: https://wandb.ai/authorize\n",
+    "# Ensure that your wandb API key is available at:\n",
+    "# os.environ['WANDB_API_KEY'] = \"<your_wandb_api_key>\"\n",
     "\n",
     "import os\n",
-    "from getpass import getpass\n",
+    "import re\n",
+    "from textwrap import dedent\n",
     "\n",
     "from openai import OpenAI\n",
```
mdlinville: Wait, is this the Inference example of the OpenAI example? Now I'm confused.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -56,124 +53,187 @@
...
-    "os.environ[\"WANDB_API_KEY\"] = getpass(\"Enter your Weights and Biases API key: \")\n",
-    "print(\"---\")\n",
-    "print(\"You can generate your OpenAI API key here: https://platform.openai.com/api-keys\")\n",
-    "os.environ[\"OPENAI_API_KEY\"] = getpass(\"Enter your OpenAI API key: \")\n",
-    "print(\"---\")\n",
-    "#####\n",
-    "\n",
-    "# 🏠 Enter your W&B project name\n",
-    "weave_client = weave.init(\"MY_PROJECT_NAME\")  # 🐝 Your W&B project name"
+    "# Setup your wandb project (team name / project name) 🐝\n",
```
mdlinville:
```suggestion
    "# Set up your W&B entity and project: \n",
```
This could be in the authorization / importing block too. The format is shown in the next line already so I removed it from the comment.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -56,124 +53,187 @@
...
+    "WANDB_ENTITY_PROJECT = \"my-team-name/my-project-name\"\n",
+    "\n",
+    "\n",
```
mdlinville: This looks odd, maybe should be joined

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -56,124 +53,187 @@
...
+    "You are an assistant that answers questions about JSON data provided by the user. The JSON data represents structured information of various kinds, and may be deeply nested. In the first user message, you will receive the JSON data under a label called 'context', and a question under a label called 'question'. Your job is to answer the question with as much accuracy and brevity as possible. Give only the answer with no preamble. You must output the answer in XML format, between <answer> and </answer> tags.\n",
+    "\"\"\")\n",
+    "    )\n",
+    "    model: str = \"OpenPipe/Qwen3-14B-Instruct\"\n",
+    "    _client: OpenAI\n",
+    "\n",
+    "    def __init__(self):\n",
+    "        super().__init__()\n",
+    "        self._client = OpenAI(\n",
+    "            base_url=\"https://api.inference.wandb.ai/v1\",\n",
```
mdlinville: Or is it the Inference example after all? (Maybe this is me not understanding the example, apologies if so)

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -56,124 +53,187 @@
...
+    "\n",
+    "\n",
+    "if __name__ == \"__main__\":\n",
+    "    if not os.environ.get(\"WANDB_API_KEY\"):\n",
+    "        print(\n",
+    "            \"WANDB_API_KEY is not set - make sure to export it in your environment or assign it in this script\"\n",
+    "        )\n",
+    "        exit(1)\n",
+    "\n",
+    "    # Find your wandb API key at: https://wandb.ai/authorize\n",
```
mdlinville: I don't think we should need to do this twice in the same notebook. Every time we handle sensitive data we increase the potential security risk.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -56,124 +53,187 @@
...
   },
   {
    "cell_type": "markdown",
-   "metadata": {
-    "id": "0mmzfm_cxr6Z"
-   },
+   "metadata": {},
    "source": [
-    "## 🐝 Run your first evaluation\n",
+    "## 🐝 Running your first evaluation (using OpenAI)\n",
```
mdlinville:
```suggestion
    "## Run your first evaluation (using OpenAI)\n",
```
Hmm, I guess the other example was the Inference example after all.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -56,124 +53,187 @@
    },
    "outputs": [],
    "source": [
-    "# Install dependancies and imports\n",
-    "!pip install wandb weave openai -q\n",
+    "# Find your wandb API key at: https://wandb.ai/authorize\n",
+    "# Ensure that your wandb API key is available at:\n",
```
mdlinville: A lot of the notebooks handle the setup differently in a way that allows you to paste it into the notebook instead of needing to set it in the environment. See https://github.com/wandb/examples/blob/7323cc40bec88074ad7dcc8bbf27948166571739/weave/docs/quickstart_dspy.ipynb#L136. This might be useful.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -56,124 +53,187 @@
...
-   "metadata": {
-    "id": "0mmzfm_cxr6Z"
-   },
+   "metadata": {},
    "source": [
-    "## 🐝 Run your first evaluation\n",
+    "## 🐝 Running your first evaluation (using OpenAI)\n",
     "\n",
-    "The following code sample shows how to evaluate an LLM using Weave’s `Model` and `Evaluation` APIs. First, define a Weave model by subclassing `weave.Model`, specifying the model name and prompt format, and tracking a `predict` method with `@weave.op`. The `predict` method sends a prompt to OpenAI and parses the response into a structured output using a Pydantic schema (`FruitExtract`). Then, create a small evaluation dataset consisting of input sentences and expected targets. Next, define a custom scoring function (also tracked using `@weave.op`) that compares the model’s output to the target label. Finally,  wrap everything in a `weave.Evaluation`, specifying your dataset and scorers, and call `evaluate()` to run the evaluation pipeline asynchronously."
+    "Your can find your OpenAI API keys here: https://platform.openai.com/api-keys"
```
mdlinville:
```suggestion
    "Find your [OpenAI API key](https://platform.openai.com/api-keys):"
```
But we don't need it both here AND  in line 162, surely. So maybe apply this edit there.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -56,124 +53,187 @@
...
+    "from textwrap import dedent\n",
+    "\n",
+    "import openai\n",
+    "\n",
+    "import weave\n",
     "\n",
-    "class ExtractFruitsModel(weave.Model):\n",
-    "    model_name: str\n",
-    "    prompt_template: str\n",
+    "# Setup your wandb project (team name / project name) 🐝\n",
```
mdlinville: Same suggestion as in the earlier cell

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -56,124 +53,187 @@
...
-    "    @weave.op()\n",
-    "    def predict(self, sentence: str) -> dict:\n",
-    "        client = OpenAI()\n",
     "\n",
-    "        response = client.beta.chat.completions.parse(\n",
-    "            model=self.model_name,\n",
+    "class JsonModel(weave.Model):\n",
+    "    prompt: weave.Prompt = weave.StringPrompt(\n",
+    "        dedent(\"\"\"\n",
+    "You are an assistant that answers questions about JSON data provided by the user. The JSON data represents structured information of various kinds, and may be deeply nested. In the first user message, you will receive the JSON data under a label called 'context', and a question under a label called 'question'. Your job is to answer the question with as much accuracy and brevity as possible. Give only the answer with no preamble. You must output the answer in XML format, between <answer> and </answer> tags.\n",
```
mdlinville: I wonder if it makes sense to set up the prompts in a different cell so you don't need to duplicate it.

mdlinville (changes requested):
I reviewed only the Evaluation notebook so far. I saw some room for streamlining / restructuring, using some techniques I'm used to seeing in some of the Models colabs, especially around handling API keys. Since I don't tend to review or run Weave notebooks, my suggestions may be off base for Weave norms. I'd appreciate feedback on my feedback before I review the other tracing notebook.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
m-rgba: So the Inference examples still use the OpenAI client - it just uses our `https://api.inference.wandb.ai/v1` endpoint as the endpoint. 

It's OpenAI-compatible, pretty similar to other compatible-based endpoints.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
m-rgba: As above

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
```diff
@@ -1,266 +1,164 @@
...
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "-OjlLbir286L"
+      },
+      "outputs": [],
+      "source": [
+        "# Ensure your dependencies are installed with:\n",
+        "!pip install openai pandas weave"
```
mdlinville: I get a ton of "requirement already satisfied" from this, and then I get bounced from Colab and need to restart the session because the latest Pandas version is not compatible with Colab, which requires v2.2.2. This helped, but still has the "Requirement already met" messages.

```
!pip install --upgrade openai pandas==2.2.2 weave
```

weave/cookbooks/source/Intro_to_Weave_Hello_Trace.ipynb
```diff
@@ -1,156 +1,111 @@
...
+      ],
+      "metadata": {
+        "id": "8ZmKBnUE8bGR"
+      }
+    },
+    {
+      "cell_type": "code",
+      "source": [
+        "# Ensure your dependencies are installed with:\n",
+        "!pip install openai weave"
```
mdlinville: I needed `jedi` too. I used:
```
!pip install --upgrade --quiet openai jedi weave
```

weave/cookbooks/source/Intro_to_Weave_Hello_Trace.ipynb
```diff
@@ -1,156 +1,111 @@
...
+    {
+      "cell_type": "markdown",
+      "source": [
+        "# Introduction to Traces\n",
+        "\n",
+        "<img src=\"http://wandb.me/logo-im-png\" width=\"400\" alt=\"Weights & Biases\" />\n",
+        "\n",
+        "Weave is a toolkit for developing AI-powered applications.\n",
+        "\n",
+        "Weave traces let you automatically capture the inputs, outputs, and internal structure of your Python functions—especially useful when working with LLMs.\n",
```
mdlinville:
```suggestion
        "Use Weave traces to capture the inputs, outputs, and internal structure of your Python function automatically to observe and debug LLM applications.\n",
```
This is a bit more action-oriented. Is this a good summary about the helpfulness?

weave/cookbooks/source/Intro_to_Weave_Hello_Trace.ipynb
```diff
@@ -1,156 +1,111 @@
...
+        "\n",
+        "By decorating a function with `@weave.op`, Weave records a rich trace of how your function runs, including any nested operations or external API calls. This makes it easy to debug, understand, and visualize how your code is interacting with language models, all from within your notebook.\n",
```
mdlinville:
```suggestion
        "When you decorate a function with `@weave.op`, Weave records a rich trace of the function while it runs, including any nested operations or external API calls. Use the trace to to debug, understand, and visualize interactions between your code and LLM models, without leaving your notebook.\n",
```

weave/cookbooks/source/Intro_to_Weave_Hello_Trace.ipynb
```diff
@@ -1,156 +1,111 @@
...
+        "\n",
+        "To get started, complete the prerequisites. Then, define a function with the `@weave.op` decorator to track LLM calls, run it on an example input, and Weave will automatically capture and visualize the trace."
```
mdlinville:
```suggestion
        "To get started, complete the prerequisites. Then, define a function decorated with `@weave.op` decorator and run it on an example input to track LLM calls. Weave captures and visualizes the trace automatically."
```

weave/cookbooks/source/Intro_to_Weave_Hello_Trace.ipynb
```diff
@@ -1,156 +1,111 @@
...
+      "execution_count": null,
+      "outputs": []
+    },
+    {
+      "cell_type": "code",
+      "source": [
+        "import os\n",
+        "import getpass\n",
+        "\n",
+        "#@title Setup your credentials\n",
```
mdlinville:
```suggestion
        "#@title Set up your credentials\n",
```
This is in a few places in each of these notebooks. "Set up" is the verb, "Setup" is the noun.

weave/cookbooks/source/Intro_to_Weave_Hello_Trace.ipynb
```diff
@@ -1,156 +1,111 @@
...
+        "inference_provider = \"W&B Inference\" #@param [\"W&B Inference\", \"OpenAI\"]\n",
+        "\n",
+        "# Setup your W&B project and credentials\n",
+        "os.environ[\"WANDB_ENTITY_PROJECT\"] = input(\"Setup your W&B project (team name/project name): \")\n",
```
mdlinville: I initially left off the `wandb/` prefix from my entity and the next cell crashed hard, and could not recover even with me fixing it and running this cell again. Had to restart the runtime to recover. Maybe a quirk of Colab.

mdlinville (changes requested):
This is much improved. Here is my feedback on the second notebook after running it in Colab.

weave/cookbooks/source/Intro_to_Weave_Hello_Eval.ipynb
m-rgba: So the `--upgrade` is causing the issue since colab expects a certain version of a few packages which is why you need the pinned version in the install here.

Instead we could do something like:

```
!pip install --quiet openai pandas weave
```

For the `jedi` warning below - that shouldn't matter at all cause it's an unused dependency for ipython that colab doesn't include by default. 

If we want to suppress it, we can either run `--no-warn-conflicts` or install it in the `pip install...` command. I think just installing it is fine, cause it looks less scary - so we can update to the final install command like:

```
!pip install --quiet jedi openai pandas weave
```

Ends up feeling like the following:

image

weave/cookbooks/source/Intro_to_Weave_Hello_Trace.ipynb
m-rgba: I tried a couple things to replicate but wasn't able to on my side:
- Errored cell 2, re-ran the credentials and then ran cell 2.
- Updated the entity/project and posted to separate projects.

May have been the crash in the colab environment you had running.

weave/cookbooks/source/Intro_to_Weave_Hello_Trace.ipynb
m-rgba: Updated with above: https://github.com/wandb/docs/pull/1875#discussion_r2529496882

mdlinville (approved):

PR #1936, opened 2025-12-01
Fix 404 errors: Add redirect rules for broken documentation links, add link-checking to validate-mdx

johndmulhausen:
## Summary

This PR adds 5 new redirect rules to `docs.json` to fix 404 errors for commonly accessed documentation paths.

## Changes

Added the following redirect rules:
- `/integrations/jupyter.html` → `/models/track/jupyter` - Redirects old Jupyter integration path to new location
- `/models/track/launch/*` → `/platform/launch` - Redirects launch docs from models to platform section
- `/library/init` → `/models/ref/python/functions/init` - Redirects library init reference to proper API docs
- `/models/app/settings-page/:slug*` → `/platform/app/settings-page/:slug*` - Fixes settings page paths
- `/_print/*` → `/` & `/docs/*` to `/models` - Handles the dozens of `/_print`and `/docs` requests that we can no longer serve
- Adds link checking to validate-mdx CI and ensures that it only runs when an mdx file has changed

## Impact

These redirects will prevent users from encountering 404 errors when following old links or bookmarks to these documentation pages, improving the overall user experience.


platform/hosting.mdx
```diff
@@ -27,7 +27,7 @@ Some features and functionality require an [Enterprise](https://wandb.ai/site/pr
   <a id="wb-Self-Managed"></a><Card title="W&B Self-Managed">
     Deploy and manage W&B Server on your own managed infrastructure. W&B Server is a self-contained packaged mechanism to run the W&B Platform & its supported W&B products. W&B recommends this option if all your existing infrastructure is on-prem, or your organization has strict regulatory needs that are not satisfied by W&B Dedicated Cloud. With this option, you are fully responsible to manage the provisioning, and continuous maintenance & upgrades of the infrastructure required to support W&B Server.
     
-    See **[W&B Self-Managed](/platform/hosting/hosting-options/self-managed/)** for more information.
+    See **[W&B Self-Managed](/platform/hosting/hostsdfsdsdfsddfsing-options/self-managed/)** for more information.
```
johndmulhausen:
```suggestion
    See **[W&B Self-Managed](/platform/hosting/hosting-options/self-managed/)** for more information.
```

.github/workflows/validate-mdx.yml
```diff
@@ -1,25 +1,79 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Need full history to compare with base branch
+      
+      - name: Check for MDX changes
+        id: mdx-check
+        run: |
+          # Get the list of changed files between the base and head branches
+          echo "Checking for MDX file changes..."
+          MDX_CHANGED=$(git diff --name-only origin/${{ github.base_ref }}...HEAD -- '*.mdx' | head -1)
```
mdlinville, line 20: Nit: This is a bit complicated, and line 26 below is probably redundant. If you felt like it, you can just store the list of MDX files found in the variable and then check its length to determine whether there are any, and if so, list them from the variable. 

I'm curious what bugged you about the `pull_request.path` way because I use it elsewhere too and if it causing problems I might need to update that.

mdlinville (commented):
One nit, but before I approve, I'm wondering how you tested the action and how I can test it. Did you push it to a fork?

.github/workflows/validate-mdx.yml
johndmulhausen: Ah, you're right about line 26, I'll have a look. 

Yes, the way this triggers got a bit of an update because PRs that don't touch .mdx files would just hang forever, with `validate-mdx` sitting in yellow status indefinitely.

PR #1941, opened 2025-12-02
Fix Weave reference generation script's handling of Typescript landing page

mdlinville:
## Description
Fix Weave reference generation script's handling of Typescript landing page

There's a bug where the generation scripts are creating  but then they try to move it to the correct spot for the landing page, but it's looking for README.mdx so it fails to update the landing page. These fixes are applied in #1940 , which no longer has broken links, has a new function, and has an actually updated Typescript landing page.

## Testing
- [x] In #1940, verify that moving the readme to overwrite the Typescript SDK landing page fixes the issue
- [x] PR tests succeed


dbrian57 (approved):
basing this review on what I saw in https://github.com/wandb/docs/pull/1940 and the code update here. Looks good!

PR #1950, opened 2025-12-02
Add forward-signals flag to sweeps agent doc

domphan-wandb:
## Description

This PR adds docs for the new  `--forward-signals` flag for the sweeps agent.


ngrayluna:
Thanks @domphan-wandb! I believe that you need to update this file as well: https://github.com/wandb/wandb/blob/main/wandb/cli/cli.py#L1735 since these docs (`wandb/docs/models/ref/*`) are autogenerated :)

domphan-wandb:
> Thanks @domphan-wandb! I believe that you need to update this file as well: https://github.com/wandb/wandb/blob/main/wandb/cli/cli.py#L1735 since these docs (`wandb/docs/models/ref/*`) are autogenerated :)

Thanks Noah! Is this the correct way to do it? https://github.com/wandb/wandb/pull/9651/files#diff-7a44f7c1d012d75ecf9ae33152ba567bdf056ba3998cc86948298cd34508b564R1733-R1739

Also, how does the ordering work when it comes to merging in docs? Like I wouldn't want these docs to show on our docs page until the feature is released.

mdlinville (changes requested):
Thank you! Unfortunately, this is generated content that will get blown away the next time we regenerate the CLI refs. We last generated them for SDK 0.25.1. Is this flag added in the upcoming version?

mdlinville:
> Thank you! Unfortunately, this is generated content that will get blown away the next time we regenerate the CLI refs. We last generated them for SDK 0.25.1. Is this flag added in the upcoming version?

Update to say we actually already have this change. :) https://docs.wandb.ai/models/ref/cli/wandb-agent

PR #1977, opened 2025-12-09
winter hackathon: canonical code snippets

ngrayluna:
## Description

First round of basic code examples for winter hackathon. The idea is to use these for the agent to reference (as a [skill](https://www.anthropic.com/engineering/equipping-agents-for-the-real-world-with-agent-skills), or for the judge, etc.).

Will do a better job of naming the Py scripts once I am done with a first pass.


ngrayluna:
Closing. Pushed the source of truth, and accessory script to https://github.com/wandb/docs-code-eval

PR #1979, opened 2025-12-10
Create redirect to solve /guides/models 404s, fix validate-mdx

johndmulhausen:
The top 404 URL is /guides/models right now; this is because the general `/guides/:slug*` rule comes before all the other more specific rules, when it should run as the final catch-all rule when the more specific rules have not matched.

Meanwhile, this PR will not merge unless I remove the `path: **.mdx` filter on the `validate-mdx` since it is rqeuired, so that has been removed. Further, I can see re: the broken link check that it was scheduled to run but it never actually fails if a link is broken, so that is fixed too. 

TO MERGE: Upload a commit that fixes the broken link in `integration.mdx`, see that the broken link check now passes. And approve this PR. It will auto-merge once approved & the link is fixed.

### DIFF
--- a/models/integrations.mdx
+++ b/models/integrations.mdx
@@ -15,7 +15,7 @@ If you're looking to integrate W&B with a library not listed here, see [Add W&B 
 
 Integrate W&B with popular machine learning frameworks:
 
-- [Catalyst](/models/integrations/catalyst/testsoetijsoietjsioetjiosetj)
+- [Catalyst](/models/integrations/catalyst)
 - [DeepChem](/models/integrations/deepchem)
 - [DSPy](/models/integrations/dspy)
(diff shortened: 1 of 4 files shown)


models/integrations.mdx
```diff
@@ -15,7 +15,7 @@ If you're looking to integrate W&B with a library not listed here, see [Add W&B
 
 Integrate W&B with popular machine learning frameworks:
 
-- [Catalyst](/models/integrations/catalyst)
+- [Catalyst](/models/integrations/catalyst/testsoetijsoietjsioetjiosetj)
```
johndmulhausen:
```suggestion
- [Catalyst](/models/integrations/catalyst)
```

PR #1981, opened 2025-12-10
Update ttl.mdx

aninahpets:
Change model registry -> registry and clarify Artifact TTL removal vs just Model TTL removal


ngrayluna (approved):
Thank you @aninahpets!

PR #2010, opened 2025-12-22
[DOCS-1842] Sync Kuberetes deploy and config docs

mdlinville:
## Description

[DOCS-1842] Sync Kuberetes deploy and config docs to prep for the larger IA restructure in DOCS-1496

Ensure recommendations are in sync for:
- Reference Architecture
- Operator install
- Bare-metal install
- Air-gapped install

1. Use snippets where possible and reconcile to the most recently updated content

2. Consolidate to Reference Architecture for recommendations that are broadly the same for all deployment types, link from other pages to the relevant sections of RA

3. For the most part:

    - Operator is the main page
    - Bare Metal relies on Operator but with differences
    - Airgapped relies on Operator but with differences
    - Reference architecture now has clear recommendations

4. Sync / consolidate database recommendations

[DOCS-1842]: https://wandb.atlassian.net/browse/DOCS-1842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


snippets/en/_includes/self-managed-version-requirements.mdx
```diff
@@ -0,0 +1,7 @@
+| Software     | Minimum version                              |
+| ------------ | -------------------------------------------- |
+| Kubernetes   | v1.29                                        |
```
flamarion: I've been testing and using W&B on the latest versions of Kubernetes without issues, so I think we should recommend the supported versions rather than `v1.29`.
I suggest we recommend starting at least from `v1.32`
https://kubernetes.io/releases/patch-releases/#non-active-branch-history

snippets/en/_includes/self-managed-object-storage-requirements.mdx
```diff
@@ -0,0 +1,11 @@
+W&B requires object storage with pre-signed URL and CORS support.
+
+For production deployments, W&B recommends using managed object storage services:
+- [Amazon S3](https://aws.amazon.com/s3/) - Object storage service offering industry-leading scalability, data availability, security, and performance.
+- [Google Cloud Storage](https://cloud.google.com/storage) - Managed service for storing unstructured data at scale.
+- [Azure Blob Storage](https://azure.microsoft.com/products/storage/blobs) - Cloud-based object storage solution for storing massive amounts of unstructured data.
+- [CoreWeave AI Object Storage](https://docs.coreweave.com/docs/products/storage/object-storage) - High-performance, S3-compatible object storage service optimized for AI workloads.
+
+For self-hosted object storage options, see the [bare-metal guide object storage section](/platform/hosting/self-managed/bare-metal/#object-storage) for detailed setup instructions including CORS configuration, MinIO deployment, and enterprise alternatives.
```
flamarion: We may need to add a warning that Minio Open Source is no longer available.
https://github.com/minio/minio?tab=readme-ov-file#maintenance-mode

platform/hosting/self-managed/bare-metal.mdx
```diff
@@ -95,7 +87,7 @@ s3://$ACCESS_KEY:$SECRET_KEY@$HOST/$BUCKET_NAME?tls=true
 ```
```
flamarion: This section is incorrect; the proper way to configure the Object Storage in W&B is either via System Console or directly in the W&B CR Spec.
Here is an example of what the configuration will look like in the W&B CR:
- AWS S3 Bucket configuration 
```
      bucket:
        kmsKey: <kms key arn>
        name: <bucket name> # Ex: wandb
        path: "" # Keep it as an empty string
        provider: s3
        region: <region> # Ex: eu-west-2
````

By using the AWS S3 the TLS is implicity

S3 Compatible
```
      bucket:
        kmsKey: null
        name: <s3 endpoint> # Ex: s3.example.com:9000
        path: <bucket name>  # Ex: wandb
        provider: s3
        region: <region> # Ex: eu-west-2
```

To enable TLS on the S3-compatible, the customer will need to add the `?tls=true` to the bucket name:
Ex:
```
      bucket:
        kmsKey: null
        name: "s3.example.com:9000"
        path: "wandb?tls=true"
        provider: "s3"
        region: "eu-west-2"
```

File: platform/hosting/self-managed/bare-metal.mdx [hunk 4]
```diff
@@ -95,7 +87,7 @@ s3://$ACCESS_KEY:$SECRET_KEY@$HOST/$BUCKET_NAME?tls=true
...
 
 <Warning>
-This works only if the SSL certificate is trusted. W&B does not support self-signed certificates.
+For SSL/TLS certificate requirements, see the [SSL/TLS section](#ssl-tls) below.
 </Warning>
 
 Set `BUCKET_QUEUE` to `internal://` if you use third-party object stores. This tells the W&B server to manage all object notifications internally instead of depending on an external SQS queue or equivalent.
```
flamarion: This is no longer necessary; the Helm chart used by the W&B Operator will handle this configuration.
Today, we no longer use a queue system; we use Redis.
I double checked, and even in our managed deployments, this env var is no longer set or set by default with `internal://`.

platform/hosting/self-managed/bare-metal.mdx
flamarion: I think it's ok to keep this, but I want to be cautious given the EOL of Minio Open Source

platform/hosting/self-managed/ref-arch.mdx
```diff
@@ -73,37 +100,113 @@ W&B requires object storage with pre-signed URL and CORS support, deployed in on
...
+<SelfManagedVersionRequirements/>
 
 ### Networking
 
-For a networked deployment, egress to these endpoints is required during _both_ installation and runtime:
-* https://deploy.wandb.ai
-* https://charts.wandb.ai
-* https://docker.io
-* https://quay.io
-* `https://gcr.io`
```
flamarion, line 89: We don't use these two in our regular deployments. Those can be safely removed.
When deploying `Bufstream` and `etcd` to support online evals for Weave, we may need `gcr.io`, but for regular deployments, we don't.

platform/hosting/self-managed/ref-arch.mdx
```diff
@@ -73,37 +100,113 @@ W&B requires object storage with pre-signed URL and CORS support, deployed in on
...
+        proxy_set_header X-Forwarded-Host $proxy_x_forwarded_host;
+
+        location / {
+            proxy_pass  http://$YOUR_UPSTREAM_SERVER_IP:8080/;
+        }
+
+        keepalive_timeout 10;
+    }
+}
+```
```
flamarion, line 194: Given we're running W&B Platform on Kubernetes, we don't need to recommend an HTTP proxy/lb anymore; the customer needs to have an `ingressclass` available to expose the services.
This recommendation was used when we ran a single container with a single endpoint. With this configuration on the platform deployed by the W&B Operator, it doesn't work because we no longer use a single endpoint; instead, multiple endpoints and services run on different ports.
Here is an example of what an ingress for self-managed, on-prem, bare-metal installation will looks like with the Weave Traces enabled
```
Name:             wandb
Labels:           app.kubernetes.io/managed-by=Helm
Namespace:        wandb
Address:          10.103.170.161
Ingress Class:    nginx
Default backend:  <default>
Rules:
  Host                   Path  Backends
  ----                   ----  --------
  wandb.fe-aws.wandb.ml  
                         /           wandb-app:8080 (192.168.222.234:8080)
                         /api        wandb-api:8081 (192.168.233.29:8081,192.168.222.232:8081)
                         /graphql    wandb-api:8081 (192.168.233.29:8081,192.168.222.232:8081)
                         /graphql2   wandb-api:8081 (192.168.233.29:8081,192.168.222.232:8081)
                         /console    wandb-console:8082 (192.168.127.182:8082)
                         /traces     wandb-weave-trace:8722 (192.168.127.166:8080)
Annotations:             meta.helm.sh/release-name: wandb
                         meta.helm.sh/release-namespace: wandb
                         nginx.ingress.kubernetes.io/proxy-body-size: 0
Events:                  <none>
```

My suggestion is reconfigure this section, pointing to Ingress Controllers instead of recommending an HTTP Proxy/LB.

platform/hosting/self-managed/ref-arch.mdx
```diff
@@ -73,37 +100,113 @@ W&B requires object storage with pre-signed URL and CORS support, deployed in on
...
 
-Please note: W&B does not recommend the use self-signed certificates and custom CAs.
+### SSL/TLS
+<SelfManagedSslTlsRequirements/>
 
 ### Supported CPU architectures
 W&B runs on the Intel (x86) CPU architecture. ARM is not supported.
```
flamarion: Intel and AMD (might be implicit, but just in case)

snippets/en/_includes/self-managed-networking-requirements.mdx
```diff
@@ -0,0 +1,11 @@
+For a networked deployment, egress to these endpoints is required during _both_ installation and runtime:
+* https://deploy.wandb.ai
+* https://charts.wandb.ai
+* https://docker.io
+* https://quay.io
+* https://gcr.io
```
flamarion: Same as the other review, we don't use these two repos in the regular deployment.

platform/hosting/operator.mdx
```diff
@@ -53,26 +60,78 @@ Configuration specifications follow a hierarchical model where higher-level spec
...
+</Note>
+
+An example security context for Kubernetes looks similar to the following:
+
+```yaml
+spec:
+  securityContext:
+    runAsUser: 100000
+    runAsGroup: 0
+```
```
flamarion, line 136: This needs to be fixed, the proper way to set the security context and pod security context goes along the following lines:
```
  api:
      install: true
      image:
        repository: wandb/megabinary
        tag: 0.74.1
      podSecurityContext:
        fsGroup: 10001
        fsGroupChangePolicy: Always
        runAsGroup: 0
        runAsNonRoot: true
        runAsUser: 10001
        seccompProfile:
          type: RuntimeDefault
      securityContext:
        allowPrivilegeEscalation: false
        capabilities:
          drop:
            - ALL
        privileged: false
        readOnlyRootFilesystem: false
```

platform/hosting/operator.mdx
flamarion, line 836: LDAP support is not present in the new Helm chart. We have a [bug report](https://wandb.atlassian.net/browse/WB-28678) for this because it wasn't carried over from the old helm chart.
For now, I'd recommend removing this section or the guide on configuring LDAP using environment variables.
Something along these lines:
```
   global:
      extraEnv:
         LDAP_ADDRESS: ldaps://ldap.company.example.com
         LDAP_BASE_DN: cn=accounts,dc=company,dc=example,dc=com
         LDAP_USER_BASE_DN: cn=users,cn=accounts,dc=company,dc=example,dc=com
         LDAP_GROUP_BASE_DN: cn=groups,cn=accounts,dc=company,dc=example,dc=com
         LDAP_BIND_DN: uid=ldapbind,cn=sysaccounts,cn=etc,dc=company,dc=example,dc=com
         LDAP_BIND_PW: ********************
         LDAP_ATTRIBUTES: email=mail,name=cn
         LDAP_TLS_ENABLE: True
         LDAP_LOGIN: True
         LDAP_USER_OBJECT_CLASS: user
         LDAP_GROUP_OBJECT_CLASS: group
```

flamarion, line 975: The new method to configure the security context looks like this:
```
    # APP
    app:
      podSecurityContext:
        fsGroup: 10001
        fsGroupChangePolicy: Always
        runAsGroup: 0
        runAsNonRoot: true
        runAsUser: 10001
        seccompProfile:
          type: RuntimeDefault
      securityContext:
        allowPrivilegeEscalation: false
        capabilities:
          drop:
            - ALL
        privileged: false
        readOnlyRootFilesystem: false
```

snippets/en/_includes/self-managed-networking-requirements.mdx
flamarion: We actually use the `quay.io` for the Prometheus images

platform/hosting/self-managed/ref-arch.mdx
flamarion: We actually use the `quay.io` for the Prometheus images.

platform/hosting/self-managed/operator-airgapped.mdx
flamarion: We stopped recommending that customers use the `wsm` because it has not been maintained for quite some time.
Although it works, I'd rather not use it for anything until it has some attention coming from the [FR](https://wandb.atlassian.net/browse/WB-26101) we created.
We can recommend connecting with the Support or the SA team to obtain the most recent list of images or pulling it directly from the helm values.

flamarion (changes requested):
Hi @mdlinville 

This is the first round of review, and I will also share it with the SA team for them to review and provide input.

Thanks a lot for this work. It will simplify the installation process and help us in the field recommend that customers follow our docs instead of creating parallel documentation.

platform/hosting/self-managed/ref-arch.mdx
```diff
@@ -43,16 +46,40 @@ The W&B Server application is deployed as a [Kubernetes Operator](/platform/host
...
 
-Consider the following when you deploy a Self-Managed MySQL database:
+**W&B strongly recommends using managed database services** (such as AWS RDS Aurora MySQL, Google Cloud SQL for MySQL, or Azure Database for MySQL) for production deployments. Managed services provide automated backups, monitoring, high availability, patching, and significantly reduce operational complexity. See the [Cloud provider instance recommendations](#cloud-provider-instance-recommendations) section below for specific service recommendations.
+
+If you choose to deploy a self-managed MySQL database, consider the following:
 
 - **Backups**. You should periodically back up the database to a separate facility. W&B recommends daily backups with at least 1 week of retention.
 - **Performance.** The disk the server is running on should be fast. W&B recommends running the database on an SSD or accelerated NAS.
 - **Monitoring.** The database should be monitored for load. If CPU usage is sustained at > 40% of the system for more than 5 minutes it is likely a good indication the server is resource starved.
```
flamarion: This threshold is very conservative. Sustained CPU usage above 40% does not, by itself, indicate resource starvation.

In most production environments, sustained CPU utilization in the 60–70% range is considered healthy and an efficient use of resources, provided latency, queue depth, and error rates remain within acceptable bounds. Alerting for resource starvation should be based on a combination of sustained high CPU utilization (e.g., 70–80%), increased query latency, and CPU contention indicators rather than utilization alone.
If we want to recommend a warning, I'd say it should start at 70% of consistent utilization and critical over 90%.

platform/hosting/self-managed/ref-arch.mdx
```diff
@@ -43,16 +46,40 @@ The W&B Server application is deployed as a [Kubernetes Operator](/platform/host
...
 - **Availability.** Depending on your availability and durability requirements you might want to configure a hot standby on a separate machine that streams all updates in realtime from the primary server and can be used to failover to in the event that the primary server crashes or become corrupted.
```
flamarion: For this topic, maybe worth adding an observation that we don't support multimaster architectures or read from replicas, and the recommendation here is purely to ensure high availability in case failures of the master nodes and a safe place to backup the data from replicas.

platform/hosting/self-managed/ref-arch.mdx
```diff
@@ -70,40 +97,127 @@ W&B requires object storage with pre-signed URL and CORS support, deployed in on
...
+<SelfManagedNetworkingRequirements/>
 
 ### DNS
 The fully qualified domain name (FQDN) of the W&B deployment must resolve to the IP address of the ingress/load balancer using an A record.
 
-### SSL/TLS
-W&B requires a valid signed SSL/TLS certificate for secure communication between clients and the server. SSL/TLS termination must occur on the ingress/load balancer. The W&B Server application does not terminate SSL or TLS connections.
+### Load balancer and ingress
+
+The W&B Kubernetes Operator exposes services using a Kubernetes ingress controller, which routes to service endpoints based on URL paths with different ports. The ingress controller must be accessible by all machines that execute machine learning payloads or access the service through web browsers.
```
flamarion, line 114: > The W&B Kubernetes Operator exposes services using a Kubernetes ingress controller,...

I'd say the W&B Operator `can` expose the services using an `Ingress Class`. 
Customers can use their own ingress class based on their ingress controllers, so they are not tied to the W&B configuration for this.

snippets/en/_includes/self-managed-mysql-requirements.mdx
```diff
@@ -0,0 +1,10 @@
+W&B requires an external MySQL database. For convenience during testing and development of proofs of concept, W&B Self-Managed can deploy a local MySQL instance (`mysql.install: true` in Helm configuration), but **this is not appropriate for production deployments**.
```
flamarion: I'd rather not recommend that if it's not appropriate for production.
We can come up with some guides on how to test W&B in another moment, but knowing the customers, if we keep this, they will use it and request us to migrate their data afterwards.
This should be an exception.

platform/hosting/self-managed/operator-airgapped.mdx
```diff
@@ -15,104 +24,85 @@ You could utilize similar commands in any continuous delivery tooling that you u
...
-sudo mv wsm /usr/local/bin
-```
+You are responsible for tracking the W&B Operator's requirements and maintaining your container registry with updated images regularly. For the most current list of required container images and versions, refer to the Helm chart, or contact [support](mailto:support@wandb.com) or your AISE.
 
-#### GitHub
-Download or clone WSM from the W&B managed `wandb/wsm` GitHub repository at `https://github.com/wandb/wsm`. See the `wandb/wsm` [release notes](https://github.com/wandb/wsm/releases) for the latest release.
+### Core W&B component containers
+* [`docker.io/wandb/controller`](https://hub.docker.com/r/wandb/controller)
+* [`docker.io/wandb/local`](https://hub.docker.com/r/wandb/local)
+* [`docker.io/wandb/console`](https://hub.docker.com/r/wandb/console)
```
flamarion, line 68: Please, add the following:
* [`docker.io/wandb/megabinary`](https://hub.docker.com/r/wandb/megabinary)
* [`docker.io/wandb/controller`](https://hub.docker.com/r/wandb/controller)

platform/hosting/self-managed/operator-airgapped.mdx
```diff
@@ -15,104 +24,85 @@ You could utilize similar commands in any continuous delivery tooling that you u
...
-wsm download
-```
+2. Inspect the `values.yaml` files to identify all container images and their versions:
 
-The output looks similar to the following:
+    ```bash
+    # Extract image references from the Helm chart
+    helm show values wandb/operator | grep -E 'repository:|tag:'
+    helm show values wandb/operator-wandb | grep -E 'repository:|tag:'
+    ```
```
flamarion, line 93: These commands will dump a bunch of non-essential information. If we want to recommend a command, we can potentially use something along these lines
```shell
helm show values wandb/operator | awk -F': *' '/^[[:space:]]*repository:/{print $2}' | grep -E '^wandb/' | sort -u
wandb/controller
flamarion@Flamarion-Jorge-HDFHVQTF7K:~ $ helm show values wandb/operator-wandb | awk -F': *' '/^[[:space:]]*repository:/{print $2}' | grep -E '^wandb/' | sort -u
wandb/anaconda2
wandb/console
wandb/frontend-nginx
wandb/local
wandb/megabinary
wandb/weave-python
wandb/weave-trace
```
But I'd stick with a fixed list of images to be honest.

platform/hosting/self-managed/operator-airgapped.mdx
flamarion: This is inconsistent with the instructions to add the local repo on step  4.
If we suggest this
```shell
helm repo add local-repo https://charts.yourdomain.com
helm repo update
```
Then the command to install should use the `local-repo` instead of `wandb`, something like this:
```shell
helm upgrade --install operator local-repo/operator -n wandb --create-namespace -f values.yaml
```

platform/hosting/self-managed/operator-airgapped.mdx
```diff
@@ -196,7 +194,9 @@ spec:
       extraEnv:
         ENABLE_REGISTRY_UI: 'true'
     
-    # If install: true, Helm installs a MySQL database for the deployment to use. Set to `false` to use your own external MySQL deployment.
+    # MySQL configuration
+    # Setting install: true deploys a local MySQL instance for testing/development only.
+    # For production, use an external managed MySQL service (RDS Aurora, Cloud SQL, Azure Database for MySQL) and set install: false.
     mysql:
       install: false
```
flamarion: Not installing MySQL is the default, and we shouldn't suggest customers use the MySQL distributed by W&B Operator, not even for tests, because they will use it and request help to migrate the data, or use it in production.

platform/hosting/self-managed/operator-airgapped.mdx
```diff
@@ -196,7 +194,9 @@ spec:
       extraEnv:
         ENABLE_REGISTRY_UI: 'true'
```
flamarion: We are missing very important configuration here for the executor and glue.
Example
```yaml
  values:
    global:
      host: https://wandb.example.com
      license: xxxxxxxxxxxxxxxxxxx

      defaultBucket:
        accessKey: xxxxxxxxxxxxxxxxxxx
        secretKey: xxxxxxxxxxxxxxxxxxx
        name: s3.example.com
        path: wandb
        provider: s3
        region: us-east-1

      mysql:
        database: wandb
        host: mysql.example.com
        password: password
        port: 3306
        user: wandb

      redis:
        host: redis.example.com
        port: 6379
        password: password

      api:
        enabled: true

      glue:
        enabled: true

      executor:
        enabled: true
```

platform/hosting/self-managed/operator-airgapped.mdx
flamarion, line 221: We are missing a lot f components configuration.
Example
```
# CONSOLE
    console:
      image:
        repository: wandb/console
        tag: 2.18.3

    # APP
    app:
      image:
        repository: registry.example.com/wandb/local
        tag: 0.75.2
      initContainers:
        init-db:
          image:
            repository: registry.example.com/wandb/local
            tag: 0.75.2

    # EXECUTOR
    executor:
      install: true
      image:
        repository: registry.example.com/wandb/megabinary
        tag: 0.75.2

    # SETTINGS MIGRATION JOB
    settingsMigrationJob:
      install: true
      image:
        repository: registry.example.com/wandb/megabinary
        tag: 0.75.2
      jobs:
        smj:
          containers:
            smj:
              image:
                repository: registry.example.com/wandb/megabinary
                tag: 0.75.2

    # API
    api:
      install: true
      image:
        repository: registry.example.com/wandb/megabinary
        tag: 0.75.2
      initContainers:
        init-db:
          image:
            repository: registry.example.com/wandb/local
            tag: 0.75.2

    # GLUE
    glue:
      install: true
      image:
        repository: registry.example.com/wandb/megabinary
        tag: 0.75.2
      initContainers:
        init-db:
          image:
            repository: registry.example.com/wandb/local
            tag: 0.75.2

    # PARQUET
    parquet:
      image:
        repository: registry.example.com/wandb/megabinary
        tag: 0.75.2

    # WEAVE
    weave:
      image:
        repository: registry.example.com/wandb/weave-python
        tag: 0.75.2

    # APP RENAME PRE HOOK
    appRenamePreHook:
      image:
        repository: registry.example.com/alpine/k8s
        tag: 1.31.12

    # APP RENAME POST HOOK
    appRenamePostHook:
      image:
        repository: registry.example.com/alpine/k8s
        tag: 1.31.12

    # OTEL
    otel:
      image:
        repository: registry.example.com/otel/opentelemetry-collector-contrib
        tag: 0.97.0

    # PROMETHEUS
    prometheus:
      server:
        image:
          repository: registry.example.com/prometheus/prometheus
          tag: "v2.47.0"

      configmapReload:
        prometheus:
          image:
            repository: registry.example.com/prometheus-operator/prometheus-config-reloader
            tag: "v0.67.0"

```

flamarion: Replace the `64m` with `0` to make the payload size unlimited.

flamarion (commented):
This is a second parse so I hope it's helpful.

The part I'm still confused about is why we are using different terms and covering different topics for the same purpose in different places.

Sorry if I'm misunderstanding this, but we have:

- Run W&B on Kubernetes
- Install on Public Cloud
- Deploy W&B Platform On-Premise

Image

I think we should have two main sections, something like:

- Deploy W&B Platform On-Premise
Here, we cover how to run W&B on Kubernetes

- Deploy W&B Platform on Public Cloud
Here, we focus on the W&B Terraform Modules and make references to any on-prem configuration we need.

In my view, the ideal structure would be:

- Dedicated Cloud
- Self Managed
  - Ref Arch
  - Deploy on Public Cloud
    - AWS
    - GCP
    - Azure
  - Deploy on-premise
    - Online 
    - Airgapped
- W&B Settings

mdlinville:
> This is a second parse so I hope it's helpful.
> 
> The part I'm still confused about is why we are using different terms and covering different topics for the same purpose in different places.
> 
> Sorry if I'm misunderstanding this, but we have:
> 
> * Run W&B on Kubernetes
> * Install on Public Cloud
> * Deploy W&B Platform On-Premise
> 
> Image
> I think we should have two main sections, something like:
> 
> * Deploy W&B Platform On-Premise
>   Here, we cover how to run W&B on Kubernetes
> * Deploy W&B Platform on Public Cloud
>   Here, we focus on the W&B Terraform Modules and make references to any on-prem configuration we need.
> 
> In my view, the ideal structure would be:
> 
> * Dedicated Cloud
> * Self Managed
>   
>   * Ref Arch
>   * Deploy on Public Cloud
>     
>     * AWS
>     * GCP
>     * Azure
>   * Deploy on-premise
>     
>     * Online
>     * Airgapped
> * W&B Settings

I agree with you in the long term, but this will be addressed in DOCS-1496 instead. This PR is already quite large and unwieldy to review, and is mostly about setting up some reuse and getting the facts to agree.

platform/hosting/self-managed/ref-arch.mdx
mdlinville: Asked about this in https://weightsandbiases.slack.com/archives/C01KQ5KTDC3/p1767737643453339. Support created the reference architecture, including this guidance.

platform/hosting/self-managed/operator-airgapped.mdx
mdlinville: For now, I will use your shorter command, reformatted to be narrower. The static list can and does go out of date, and this will provide a bit of protection against that.

platform/hosting/self-managed/ref-arch.mdx
mdlinville: Let's save this update for later when we have approval from technical stakeholders. That will unblock this PR. :)

flamarion (approved):
🚢

PR #2013, opened 2025-12-23
[Weave] TypeScript coverage for evals pt. 2

dbrian57:
## Description
This adds TypeScript examples to the Weave evaluation docs and calls out SDK gaps where applicable. The code examples have been tested and work.

This PR also breaks out column mapping into its own small doc.

### DIFF
--- a/weave/guides/core-types/datasets.mdx
+++ b/weave/guides/core-types/datasets.mdx
@@ -161,7 +161,9 @@ Select a tab to see Python and TypeScript-specific code.
 
   </Tab>
   <Tab title="TypeScript">
-   This feature is not currently available in TypeScript.  Stay tuned!
+  ```plaintext
+   This feature is not currently available in TypeScript yet.
+   ```
   </Tab>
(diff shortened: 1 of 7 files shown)


weave/guides/core-types/datasets.mdx
```diff
@@ -161,7 +161,9 @@ Select a tab to see Python and TypeScript-specific code.
 
   </Tab>
   <Tab title="TypeScript">
-   This feature is not currently available in TypeScript.  Stay tuned!
+  ```plaintext
```
mdlinville, line 164: Should this be a proper callout instead of code font?

mdlinville (approved):

weave/guides/core-types/datasets.mdx
dbrian57: It's applied like this throughout the Weave docs (before I got here), so I'm just being consistent with the style for the moment. Gonna change all of these at once in a week or so here.

PR #2018, opened 2025-12-30
[DOCS-1956] Update docs for hashed API keys

mdlinville:
## Description
[DOCS-1956] Update docs for hashed API keys

- Update references to `authorize` to point to User Settings instead
- Create reusable snippets for finding, creating, and storing org and team service accounts and API keys
- Update content in the snippets to align with hashed API key updates
- Update pages to use the snippets
- Mention minimum SDK versions
- Consolidate articles about rotating and finding API keys add redirect
- Copyedits

## Testing
- [x] Test SDKs locally
- [x] Local build succeeds without errors (`mint dev`)
- [x] Local link check succeeds without errors (`mint broken-links`)
- [x] PR tests succeed

[DOCS-1956]: https://wandb.atlassian.net/browse/DOCS-1956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2020, opened 2026-01-05
[DOCS-1972] Add script to remove metadata from notebook source

mdlinville:
## Description
[DOCS-1972] Add script to remove metadata from notebook source

Uses `nb-clean`, checks for it locally and tries to install it locally if not found.

To run:

```bash
bash scripts/clean-notebook.sh <path-to-notebook>
```

## Testing
- [x] Tested in scope of #2018
- [x] PR tests succeed

[DOCS-1972]: https://wandb.atlassian.net/browse/DOCS-1972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/clean-notebook.sh
+++ b/scripts/clean-notebook.sh
@@ -4,44 +4,38 @@ set -e
 # Script to clean Jupyter notebook metadata using nb-clean
 # This removes execution counts, outputs, and cell metadata that causes noisy diffs
 
-# Color codes for output
-RED='\033[0;31m'
-GREEN='\033[0;32m'
-YELLOW='\033[1;33m'
-NC='\033[0m' # No Color
-
 # Function to print error messages
 error() {
-  echo -e "${RED}Error:${NC} $1" >&2
+  echo "Error: $1" >&2
   exit 1
 }
 
 # Function to print success messages
 success() {
-  echo -e "${GREEN}✓${NC} $1"
+  echo "✓ $1"
 }
 
 # Function to print info messages
 info() {
-  echo -e "${YELLOW}→${NC} $1"
+  echo "→ $1"
 }
 
-# Check if nb-clean is installed (try both command and Python module)
-NB_CLEAN_CMD=""
-if command -v nb-clean &> /dev/null; then
-  NB_CLEAN_CMD="nb-clean"
-elif python3 -c "import nb_clean" &> /dev/null; then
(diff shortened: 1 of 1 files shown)


scripts/clean-notebook.sh
```diff
@@ -0,0 +1,87 @@
...
+set -e
+
+# Script to clean Jupyter notebook metadata using nb-clean
+# This removes execution counts, outputs, and cell metadata that causes noisy diffs
+
+# Color codes for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m' # No Color
```
ngrayluna: Seems a bit extra, can we omit these colors? A simple print message will suffice.

scripts/clean-notebook.sh
```diff
@@ -0,0 +1,87 @@
...
+info() {
+  echo -e "${YELLOW}→${NC} $1"
+}
+
+# Check if nb-clean is installed (try both command and Python module)
+NB_CLEAN_CMD=""
+if command -v nb-clean &> /dev/null; then
+  NB_CLEAN_CMD="nb-clean"
+elif python3 -c "import nb_clean" &> /dev/null; then
+  NB_CLEAN_CMD="python3 -m nb_clean"
```
ngrayluna: Incorrect command, use `nb-clean`

scripts/clean-notebook.sh
```diff
@@ -0,0 +1,87 @@
...
+else
+  error "nb-clean is not installed.
+
+To install nb-clean, run:
```
ngrayluna: Let's remove this and instead make an elif check that checks if `nb-clean` is installed. If not, install it for them. That'll make it more set up for eventual CI.

ngrayluna (changes requested):

scripts/clean-notebook.sh
mdlinville: Done

scripts/clean-notebook.sh
mdlinville: Done

scripts/clean-notebook.sh
mdlinville: Done

ngrayluna (approved):

PR #2022, opened 2026-01-07
Reduce the links on the homepage to be onboarding only

johndmulhausen:
Goals:
- Create a more "scannable" experience
- For the flagship products (Models, Weave) allow three links, and for the others allow two. 
- Focus only on onboarding, do not attempt to recreate the TOC or "promote" new functionality with high-vis links.

Rather than overwhelm the user with 19 links, we're down to 10.


mdlinville (approved):
It seems a little strange to keep the Inference playground link and lose the Weave one, since Weave Playground seems like something we are pushing.

johndmulhausen:
> It seems a little strange to keep the Inference playground link and lose the Weave one, since Weave Playground seems like something we are pushing.

Something had to go here; the playground is actually pretty advanced as a tool and better to use after you've absorbed more context. Trying to focus on what gives people the best "first few minutes" with the products.

PR #2029, opened 2026-01-09
Updating trace view page for completeness and accuracy and graph view-DOCS-1903 

anastasiaguspan:
## Description
Adding the graph view like the ticket requested, plus fixing this other stuff that was noted upon my review:
Trace view
1
“Left sidebar” — I think this was supposed to be “left panel”
The sidebar is kinda the menu… and the leading sentence says ‘three panels”
2
‘Trace view navigation’
: Stack arrows: says it moves ‘up and down the stack’
But, assuming the Stack is the same thing at the bottom panel below the tree, that does something different.
It seems the up down buttons move between traces, effectively navigating the LEFT pane from the center pane (which is admittedly unexpected behavior).

3
Scrubbers and contextual nav’
“The panel below the tree”. - what tree? This term wasn’t really introduced let’s identify this better/officially
4
Add <collapsible> panel - mine was small and needed to be expanded to match the view.
5
“click the ^ button “   odd word choice.  expansion button? Just me?
6
Alternate trace tree views
Says
‘In the upper right corner’
But it’s not.
Its inside the center panel, and not in the corner

“Near the top of the panel” would be better.
7 Ohhh this has moved …. ALL the screenshots should be updated…none of them show graph view
8
Traces (default)
The default view showing, stack hierarchy, cost per op, execution time, and status indicators.
“Cost per op” not showing on mine? Has this changed or does it depend on the trace? 
>Add 'if avl'
9
The ‘code’ trace picture also shows cost … mine doesn’t (perhaps my example uses free usage?)
If cost only shows sometimes we should clarify this?

10)
Clean up content for standardized way to refer to controls, ‘ops’ - clarify what ops is.

11. Add alt text for images.
12. [[Uhh, we say left panel, center panel, and right panel, but what about on mobile? One of the videos talk about mobile users but do we think people are using this on mobile. John says yes, mobile is a thing (or will be) - still, I’ll leave this as is for now…  won't fix]]
13. Oddly, this page really only talks about the center panel. Add a little info on the other two for completeness.

Questions from engineering:
In the Trace Detail panel:
1.Under “Code View” why does the scrubber change to “Path” - what is the significance of Path when, from what I can tell, it behaves similar to “Peer” in the default view (e.g. jumping to the next instance of the same type).
>adding description for Path
2.Doc implied there is a UI control for this but couldn’t find it:
- _Filter persistence_: Selecting operations across traces retains the sub-path context for easier comparison.
Is this still a thing, Or was this dropped with the change to Regex for the filter? I don’t understand what this is trying to tell the user.
>Remove this

3.I was thinking of adding the Hotkeys to traverse trace tree of Opt up/down - what is it on Windows: Alt, Or not implemented ?
>Yes, alt. Adding.

## Testing
- [ ] PR tests succeed

## Related issues

- Fixes DOCS-1903


weave/guides/tracking/trace-tree.mdx
```diff
@@ -1,92 +1,96 @@
...
+3. Choose a trace to open the Trace Details view. The Trace Details view displays additional panels with a hierarchical breakdown of the trace execution.
 
 ## Traces page overview
 
 The Traces page is composed of three core panels:
 
-- _Left sidebar_: A sortable, paginated list of all trace runs for the project.
-- _Center panel_: Interactive [trace view](#trace-view-navigation) showing the stack and ops hierarchy for a selected trace. 
-- _Right panel_: Detailed view for a selected op (Call, Code, Feedback, Scores, Summary, Use).
+- _Left panel_: A sortable, paginated list of all trace runs for the project.
```
dbrian57: I'm not sure these should be italicized. Typically, we use bold here on bulleted lists. Google style guide seems to agree

weave/guides/tracking/trace-tree.mdx
```diff
@@ -1,92 +1,96 @@
...
-## Trace view navigation
 
-- _Breadcrumbs_: At the top of the center panel, navigate up and down the trace stack via the breadcrumb trail.
-- _Stack arrows_: Use the `↑` and `↓` buttons to move up and down the stack.
-- _Double-click_: Double-click on an op to focus the view exclusively on that substack.
-- _"Jump to Top" Button_: Return to the root of the trace stack.
+### Filter within a trace
 
-### Filter and search
+- _Regex filter by name(s)_: Use the text field above the trace tree to filter ops by name or type, such as `tool`, `openai.response.create`.
```
dbrian57: same comment as above

weave/guides/tracking/trace-tree.mdx
```diff
@@ -1,92 +1,96 @@
...
 The Traces page is composed of three core panels:
 
-- _Left sidebar_: A sortable, paginated list of all trace runs for the project.
-- _Center panel_: Interactive [trace view](#trace-view-navigation) showing the stack and ops hierarchy for a selected trace. 
-- _Right panel_: Detailed view for a selected op (Call, Code, Feedback, Scores, Summary, Use).
+- _Left panel_: A sortable, paginated list of all trace runs for the project.
+  - This traces table includes additional data such as tokens, cost, and latency.
+- _Center panel_: Interactive trace view for a selected trace. The trace tree shows a hierarchy of all the methods tracked within the trace. 
+  - The trace tree displays ops, which are `@weave.op()`-decorated functions, that were called during the trace. 
+  - A Weave op is a versioned function that automatically logs all calls.
```
dbrian57: I think the wording on this is kinda confusing, but I also don't think this line is necessary. IMO, we should nix it.

weave/guides/tracking/trace-tree.mdx
```diff
@@ -1,92 +1,96 @@
...
-- _Filter persistence_: Selecting ops across traces retains the sub-path context for easier comparison.
+### Navigate a trace
+The trace tree shows a hierarchy of all the methods tracked within the trace. To move up/down the tree, use `Cmd` (macOS) or `Alt` (Windows/Linux) + Up Arrow (↑) / Down Arrow (↓).
 
-<Frame>
-![Filter an op by name](/weave/guides/tracking/imgs/trace-tree-filter.png)
-</Frame>
-
-### Scrubbers and contextual navigation
+There are several scrubbers available below the trace tree.  These scrubbers provide rapid navigation across states within the trace tree.  You can use the sliders to navigate strategically through your trace.
```
dbrian57:
```suggestion
There are several scrubbers below the trace tree that provide rapid navigation across states within it.  You can use the sliders to strategically navigate through your trace.
```
This reads cleaner to me

weave/guides/tracking/trace-tree.mdx
```diff
@@ -1,92 +1,96 @@
...
 
-The panel below the tree includes multiple scrubbers for navigating across calls:
-
-- **Timeline**: Chronological order of events.
-- **Peers**: Ops sharing the same type.
-- **Siblings**: Ops with the same parent.
+Expand the panel to see all available scrubbers:
+- **Timeline**: Chronological order of events within the trace.
+- **Peers**: Ops sharing the same type. For example, if you were examining details of a function called `predict`, you can use this scrubber to immediately jump to the next execution of `predict` within the trace.
+- **Siblings**: Ops with the same parent.  Use this scrubber to iterate over ops nested under the parent function call.
```
dbrian57, line 45: all of this is good stuff

weave/guides/tracking/trace-tree.mdx
```diff
@@ -1,92 +1,96 @@
...
+
+### View details for a trace op
+
+After you have selected an op in the trace tree, details for that op display in the next panel. These details are grouped into the following tabs:
 
-- Use the **"Filter by op name”** search bar at the top of the trace tree view to quickly isolate relevant tool or LLM calls.
-- Switch between views based on your debugging need. Use **Code View** for call logic, **Flame Graph** for to understand performance over time, and **Graph View** to understand structure.
+- **Call**: The input and output to the op execution.
+- **Code**: The code that was used when the call was made.
+- **Feedback**: Any available feedback for the op. You can provide feedback directly within Weave or through the API.
```
dbrian57: might be worth linking to the feedback doc here: https://docs.wandb.ai/weave/guides/tracking/feedback

weave/guides/tracking/trace-tree.mdx
```diff
@@ -1,92 +1,96 @@
...
+- **Scores**: Any available scores for the op. Calls are scored by running Evaluations.
```
dbrian57: same for scores

dbrian57 (approved):
This looks great! Few suggestions inline. The suggestions are small enough that I don't think I'll need to review this again, so I'm setting to Approve.

PR #2034, opened 2026-01-13
Workspaces: Updated default workspace template

ngrayluna:
## Description

Updated workspace template table. To avoid nested markdown table, I split the table into smaller ones.

## Related issues

- Fixes DOCS-1977


models/track/workspaces.mdx
```diff
@@ -71,14 +71,49 @@ Use _workspace templates_ to quickly create workspaces using the same settings a
...
+
+| Display preferences | Default |
+|--------------|--------------|
+| Color run names          | On  |
+| Display full run name    | Off |
+| Show X range in tooltip  | Off |
+| Tooltip runs             | Default |
+| Sync zoom across charts  | Off |
+| Show higlighted run only | Off | 
+
```
mbarrramsey, line 109: I think some settings and defaults are missing here -- this is the list of values that have a default
Screenshot 2026-01-13 at 2 33 15 PM

and this is the list of total settings supported:
```
// x-axis settings
xAxis
xLogScale
xAxisMin
xAxisMax
xAxisFormat
// y-axis settings
yLogScale
yAxisMin
yAxisMax
// smoothing settings
smoothingType
smoothingWeight
showOriginalAfterSmoothing
// runs settings
maxRuns
pointVisualizationMethod
excludeOutliers
// grouping settings
useRunsTableGroupingInPanels
groupAgg
// display preferences
colorRunNames
displayFullRunName
showLegend
showXRange
tooltipNumberOfRuns
coordinateZooming
highlightedCompanionRunOnly
```

ngrayluna:
@mbarrramsey Updated PR to include most up to date version of default template settings.

Let's create another doc ticket for tracking non default settings. I imagine that it'd be useful to document (for e.g.) possible value ranges, types (int, float, etc.) for each option.

models/track/workspaces.mdx
```diff
@@ -70,14 +70,43 @@ Use _workspace templates_ to quickly create workspaces using the same settings a
...
 
-| Setting | Default |
-|-------|----------
-| X axis             | Step |
-| Smoothing type     | Time weight EMA |
-| Smoothing weight   | 0 |
-| Max runs           | 10 |
-| Grouping in charts | on |
-| Group aggregation  | Mean |
+| X axis       | Default      |
```
mdlinville, line 73: Nit: This seems inefficient. What about a 3-column layout where the first column is X axis, Y axis, Smoothing, etc, second column is the setting within that, like Value, Log scale, Weight, and third column is the default?

mdlinville (approved):

models/track/workspaces.mdx
ngrayluna: I originally looked into a nested table. But the markdown quickly turned messy-looking. So I opted for the less efficient (but more maintainable) approach.

PR #2037, opened 2026-01-14
[DOCS-1988] Update Lychee link check actions to support crawling

mdlinville:
## Description
Fixes DOCS-1988

David F on CoreWeave docs team discovered that Lychee does not crawl recursively by default. This means that in both PRs and production, we were only checking the site index for broken http:// and https:// links. This PR fixes that and makes other improvements.

- Move configuration to `lychee.toml`
- Consolidate excludes into `lychee.toml`
- Exclude non-docs directories like `.github` and `scripts`
- Move some logic from Action steps to the actual scripts
- Prod check updates:
  - Now parses the sitemap to find pages to check.
  - Now also checks internal links since it's very fast.
- PR check updates:
  - Now checks only files that the PR updates
  - Now leaves a PR comment with the results
  - No longer actually fails the PR, in case the broken link is temporary 
  - Now works in forks with a small reduction in functionality: In a fork PR that adds a new page, internal links into that new page will show as broken. The PR comment hints about this.
  - Now warns about redirects caused by trailing `/` in internal links
- Fixed some broken links:
  - Weave core-types pages that use relative links with the wrong number of `../`s
  - Links with trailing `/` in the main `index.mdx` which cause an unnecessary redirect.

## Testing
- **Prod**: Tested in my fork (see https://github.com/mdlinville/docs/actions/runs/21009619899). Checked 1134 links in 24 secxonds.
- **PR**: Tested in my fork (see https://github.com/mdlinville/docs/actions/runs/21011598457 and https://github.com/mdlinville/docs/pull/63#issuecomment-3751737490). Took 44 seconds to check links in 16 changed files.

- [x] PR tests succeed


.github/workflows/linkcheck-pr.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         id: lychee
         uses: lycheeverse/lychee-action@v2
         with:
-          args: "--threads 5 --max-retries 5 --retry-wait-time 2 --include '^https?://' --include '^http?://' --base-url='${{ github.event.deployment_status.environment_url }}' '${{ github.event.deployment_status.environment_url }}'"
+          args: "--threads 5 --max-retries 5 --retry-wait-time 2 --include '^https?://' --include '^http?://' --base-url='${{ github.event.deployment_status.environment_url }}' '.'"
```
mdlinville:
```suggestion
          args: "--threads 5 --max-retries 5 --retry-wait-time 2 --exclude '/ja' --exclude '/ko' --include '^https?://' --include '^http?://' --base-url='${{ github.event.deployment_status.environment_url }}' '.'"
```

dbrian57 (approved):

PR #2038, opened 2026-01-14
[Weave] Adds OTEL Typescript examples

dbrian57:
## Description

This PR adds the equivalent TypeScript examples to the OTEL integration doc.

### DIFF
--- a/weave/guides/tracking/otel.mdx
+++ b/weave/guides/tracking/otel.mdx
@@ -138,7 +138,7 @@ python openinference_example.py
 The TypeScript implementation of this example contains the following key differences from the Python implementation:
 
 * OpenAI must be imported before registering instrumentation (ESM modules require this).
-* Uses @opentelemetry/exporter-trace-otlp-proto (protobuf format) instead of the HTTP exporter, since W&B's endpoint only accepts protobuf.
+* Uses `@opentelemetry/exporter-trace-otlp-proto` (protobuf format) instead of the HTTP exporter, since W&B's endpoint only accepts protobuf.
 * Requires explicit `provider.shutdown()` with a delay before shutdown to ensure spans are flushed, since `BatchSpanProcessor` flushes asynchronously.
 
 Paste the following code into a TypeScript file such as `openinference_example.ts`:
--- a/weave/guides/tracking/otel.mdx
+++ b/weave/guides/tracking/otel.mdx
@@ -175,8 +175,7 @@ const exporter = new OTLPTraceExporter({
 
 const provider = new NodeTracerProvider({
   spanProcessors: [
-    new BatchSpanProcessor(exporter),
-    new BatchSpanProcessor(new ConsoleSpanExporter()),
+    new BatchSpanProcessor(exporter)
   ],
 });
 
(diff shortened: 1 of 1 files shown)


weave/guides/tracking/otel.mdx
```diff
@@ -108,23 +127,140 @@ if __name__ == "__main__":
...
+// Configure the OTLP exporter
+const exporter = new OTLPTraceExporter({
+  url: OTEL_EXPORTER_OTLP_ENDPOINT,
+  headers: OTEL_EXPORTER_OTLP_HEADERS,
+});
+
+const provider = new NodeTracerProvider({
+  spanProcessors: [
+    new BatchSpanProcessor(exporter),
+    new BatchSpanProcessor(new ConsoleSpanExporter()),
```
zbirenbaum: nit: Maybe comment this line and put a note that says 'If you want to print to console as well'.

This isn't strictly required to export to weave and might mess up the user's existing logs.

weave/guides/tracking/otel.mdx
```diff
@@ -108,23 +127,140 @@ if __name__ == "__main__":
...
+provider.register();
+
+// Register the OpenAI instrumentation with the tracer provider
+const openAIInstrumentation = new OpenAIInstrumentation();
+openAIInstrumentation.setTracerProvider(provider);
+
+// Manually instrument OpenAI since we're using ESM
+openAIInstrumentation.manuallyInstrument(OpenAI);
+
+console.log("Instrumentation registered. Is patched?", isPatched());
```
zbirenbaum: We don't do this sort of logging in the python example I don't think

weave/guides/tracking/otel.mdx
```diff
@@ -108,23 +127,140 @@ if __name__ == "__main__":
...
+async function main() {
+  console.log("OpenAI is patched?", isPatched());
+  
+  // Create a manual span to verify pipeline works
+  const tracer = trace.getTracer("test-tracer");
+  const manualSpan = tracer.startSpan("manual-test-span");
+  manualSpan.setAttribute("test", "manual span works");
+  manualSpan.end();
+  console.log("Manual span created");
+
```
zbirenbaum: None of this should be required if you are using Arize. The logs are also probably not necessary

weave/guides/tracking/otel.mdx
```diff
@@ -108,23 +127,140 @@ if __name__ == "__main__":
...
+await main();
+
+// Give spans time to flush
+console.log("Waiting 2 seconds for spans to flush...");
+await new Promise(resolve => setTimeout(resolve, 2000));
+
+await provider.shutdown(); // flush all pending spans before exit
+console.log("Shutdown complete");
+```
```
zbirenbaum, line 218: Don't await outside an async function.

zbirenbaum (commented):
The lower examples look good, but I think the instrumented examples can be improved a bit.

I think you can do something like this to perform a lot of the work in one step (I haven't verified but google thinks so).

Also, soon we will use ResourceAttributes for handling project and entity, alongside any other wandb specific attributes like wb_run_id so the below code is even more applicable.

```
// instrumentation.ts
import { NodeSDK } from "@opentelemetry/sdk-trace-node";
import { OTLPTraceExporter } from "@opentelemetry/exporter-trace-otlp-http";
import { OpenAIInstrumentation } from "@arizeai/openinference-instrumentation-openai";
import { Resource } from "@opentelemetry/resources";
import { SemanticResourceAttributes } from "@opentelemetry/semantic-conventions";

const sdk = new NodeSDK({
  resource: new Resource({
    [SemanticResourceAttributes.SERVICE_NAME]: "my-llm-app",
  }),
  traceExporter: new OTLPTraceExporter({
    // Default to http://localhost:4318/v1/traces use weave url here
  }),
  instrumentations: [new OpenAIInstrumentation()],
});

sdk.start();
```

PR #2011, opened 2025-12-23
Runs: Rewrite Runs section

ngrayluna:
## Description

Classic Noah, started off as a simple "update images" task...then grew to a complete Runs rewrite...here's a summary of what I did:

  New Documentation Pages (7 added):
  - customize-run-display.mdx - How to customize run display settings
  - delete-runs.mdx - Instructions for deleting runs
  - run-identifiers.mdx - Explanation of run IDs and identifiers
  - run-states.mdx - Documentation on run states
  - search-runs.mdx - How to search for runs
  - stop-runs.mdx - How to stop runs
  - view-logged-runs.mdx - Viewing logged run data

  Renamed File:
  - multiple-runs-per-process.mdx → initialize-run.mdx

  Major Edits:
  - models/runs.mdx - Significantly reduced (597 lines removed) as content was split out
  - models/runs/grouping.mdx - Reorganized and updated
  - models/runs/forking.mdx - Expanded with more content
  - models/runs/filter-runs.mdx - Updates and improvements
  - Updated navigation in docs.json

  Image Cleanup:
  - Removed 17 outdated images/GIFs
  - Added 6 new, updated images

 

## Related issues

- Fixes https://wandb.atlassian.net/browse/DOCS-1716

PR #2054, opened 2026-01-16
Add PR comment updating to link checker workflow

mdlinville:
## Description

Recent updates added PR commenting to the link checker, but the initial comment would not be updated by subsequent runs. This fixes that.

## Testing
- [x] In PR #2011 the [comment](https://github.com/wandb/docs/pull/2011#issuecomment-3761507554) updated successfully 
- [x] PR tests succeed


.github/workflows/linkcheck-pr.yml
```diff
@@ -87,6 +87,47 @@ jobs:
...
+              core.info(`PR number from pull_request event: ${prNumber}`);
+              core.setOutput('pr_number', prNumber);
+              return prNumber;
+            }
+            
+            // For deployment_status events, find PR from commit SHA
+            if (context.eventName === 'deployment_status') {
+              const sha = context.payload.deployment?.sha;
+              if (!sha) {
+                core.warning('No deployment SHA found');
```
ngrayluna, line 107: Is the deployment SHA the same has the GitHub PR SHA? If yay, you can use the baked in environment variable 
```bash
echo "The commit SHA is: ${{ github.sha }}"
```

If not, disregard this comment.

.github/workflows/linkcheck-pr.yml
```diff
@@ -87,6 +87,47 @@ jobs:
             **/*.md
             **/*.mdx
 
+      - name: Get PR number
+        id: get-pr
+        uses: actions/github-script@v8
+        with:
+          script: |
+            // For pull_request events, get PR number directly
```
ngrayluna, line 95: Another idea, not sure if it's relevant, but depending on the event this is triggered for, you can automatically get access to PR #'s , etc.

E.g. for `pull_request` events, we get the PR number:

```bash
echo "The PR number is ${{ github.event.number }}"
```

mdlinville: We are, from within the `context.payload` struct, see line 97. I don't mind David's approach of saving it to a local variable.

.github/workflows/linkcheck-pr.yml
mdlinville: This is the SHA that the Mintlify preview is built from, from Mintlify's POV. If Mintlify stopped building previews for subsequent hashes for some reason, this is how we would get visibility into that.

.github/workflows/validate-mdx.yml
```diff
@@ -3,8 +3,7 @@ on:
   pull_request:
   push:
     branches:
-      - '**'
-      - '!main'
+      - main
```
mdlinville, line 6: This just prevents a duplicate `validate-mdx` trigger on every PR push. This branch definition means that the check will run on direct commits to `main` (outside the scope of a PR.

PR #2060, opened 2026-01-20
[DOCS-1909] Multi-metric line plots

mdlinville:
## Description
[DOCS-1909] Multi-metric line plots

## Testing
- [x] Verified against Multi-tenant Cloud UI
- [x] PR tests succeed

[DOCS-1909]: https://wandb.atlassian.net/browse/DOCS-1909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


models/app/features/panels/line-plot.mdx
```diff
@@ -13,6 +13,45 @@ Line plots display by default for metrics logged with `wandb.Run.log()` over tim
...
+    1. Click **Apply**.
+</Tab>
+<Tab title="Multi-metric line plot">
+
+<Note>
+This feature is in preview, available by invitation only. To request enrollment, contact [support](mailto:support@wandb.com) or your AISE.
+</Note>
+
+In an [automatic workspace](/models/app/features/panels#workspace-modes), a single-metric line plot is created automatically for each logged metric. This section shows how to create a single line plot that shows multiple metrics together, defined by a JavaScript regular expression. You can optionally consolidate many single-metric plots into a single multi-metric plot. This can improve the performance of a workspace with many logged metrics, and can help you analyze the results of your runs efficiently.
+
```
julia-g-rose, line 49: Since the steps below are visual flows, what are your thoughts on including images or a GIF?

models/app/features/panels/line-plot.mdx
```diff
@@ -13,6 +13,45 @@ Line plots display by default for metrics logged with `wandb.Run.log()` over tim
...
+1. To add a line plot globally, click **Add panels** in the control bar near the panel search field.
+
+    To add a line plot directly to a section instead, click the section's action `...` menu, then click **+ Add panels**.
+1. Click **Quick panel builder**, then click the **Multi-metric panels** tab.
+1. In **Regex** enter an expression in [JavaScript regular expression](https://www.w3schools.com/js/js_regexp.asp) format. The plot will include lines for all metrics that match the expression, including metrics logged in the future. As you type, the UI updates to show which metrics match the expression.
+1. To optionally remove duplicate single-metric panels when the multi-metric plot is created, toggle **Clean up auto-generated panels**. A preview shows which panels will be cleaned up. <Note>When this option is turned on, a single-metric plot will not be created for a newly logged metric that matches the expression. Instead, it will be included only in this multi-metric plot.</Note>
+1. Click **Create &lt;number&gt; panels**.
```
julia-g-rose, line 57: Would it help to explicitly add common use cases and examples? For example, looking at metrics across layers in the same panel, rather than separate panels.

I think we should also explain that capture groups create multiple panels, as well as the difference between `(layer_0|layer_10)` vs. `(:?layer_0|layer_10)`, which was a point of confusion for one of our private preview customers who tested it out . See thread: https://weightsandbiases.slack.com/archives/C09FE63E7R9/p1768847580043079

models/app/features/panels/line-plot.mdx
mdlinville: We strongly prefer to avoid animated Gifs, which are impossible to maintain. In this case, we have changed these UIs in tiny ways so many times over the last year that I don't feel enthusiastic about taking on the maintenance of even non-animated images here. For example, nobody told me that we changed **Quick add** to **Quick panel builder**.

models/app/features/panels/line-plot.mdx
julia-g-rose: @mdlinville I just wanted to follow up on this. Could we add examples of common syntax and their effects on the resulting panels, such as `(layer_0|layer_10)` vs. `(:?layer_0|layer_10)`, similar to what we have in the UI, since this was a point of confusion for the customer? The in-UI examples are not visible after the user starts typing (and sometimes they never see it), so we were hoping that providing a link to the docs would help clarify the behavior of things like capture groups so there are no surprises. Customer quote: "It's a bit surprising and UX wise I wouldn't like to accidentally add a dozen panels" because they didn't understand the `(layer_0|layer_10)` syntax creates multiple panels, while `(:?layer_0|layer_10)` puts everything in a single panel

Screenshot 2026-01-22 at 4 16 55 PM

mdlinville: Done, PTAL!

snippets/en/_includes/line-plot-capture-groups.mdx
julia-g-rose, line 1: This is great and very clear, thank you!

models/app/features/panels/line-plot/reference.mdx
```diff
@@ -159,3 +160,16 @@ Rescale the x-axis using calculated values with custom expressions.
...
+
+### Multi-metric panel expressions
+Use a regular expression to create a single line plot that shows multiple metrics together (including matching metrics logged in the future). For detailed instructions, see [Add a line plot](/models/app/features/panels/line-plot#multi-metric-line-plot).
+
+For example:
+- Instead of creating separate panels for each layer's metrics, you can view them together in a single panel. For example, if you log metrics with consistent naming, like `layer_0_loss`, `layer_1_loss`, and `layer_2_loss`, you can use a regex like `layer_\d+_loss` to display all layer losses on one plot.
+- Match all metrics that share a common naming pattern. For example:
+    - `train_.*` matches all training metrics like `train_loss`, `train_accuracy`, `train_f1_score`
+    - `.*_accuracy` matches accuracy metrics across different datasets like `train_accuracy`, `val_accuracy`, `test_accuracy`
+- Use alternation to match only the metrics you want. For example, `(layer_0|layer_10)_loss` matches just the first and tenth layer losses, excluding intermediate layers.
```
julia-g-rose: I feel like this last example implies that when specifying `(layer_0|layer_10)_loss`, layer_0 and layer_10 loss will show up in the same panel, rather than separate panels. Should we use the non capture group example `(?:layer_0|layer_10)_loss` instead?

models/app/features/panels/line-plot.mdx
```diff
@@ -13,6 +15,77 @@ Line plots display by default for metrics logged with `wandb.Run.log()` over tim
...
+### More about multi-metric regular expressions
+
+Multi-metric line plots use [JavaScript regular expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions) to match metric names. These section demonstrate some common use cases and gives more details about how the regular expressions work, such as how capture groups affect the panels that are created.
+
+#### Common use cases
+This section shows some ways you can use multi-metric panels to help you analyze your experiment results.
+
+**Compare metrics across layers or model components**
+Instead of creating separate panels for each layer's metrics, you can view them together in a single panel. For example, if you log metrics with consistent naming, like `layer_0_loss`, `layer_1_loss`, and `layer_2_loss` in this Python example code, you can use the regex `layer_\d+_loss` to display all layer losses on one plot.
+```python
```
julia-g-rose, line 68: I don't feel strongly about this, but we could probably remove the python code snippet for brevity. That said, totally fine to keep it if you think it's clearer

mdlinville: I think it's better to keep it, so that you can see how you'd name the metrics when logging them to be able to match them this way. If you're pretty new to W&B, you might not understand it as well without seeing the `run.log()` bit.

julia-g-rose: Sounds good!

snippets/en/_includes/line-plot-capture-groups.mdx
```diff
@@ -0,0 +1,19 @@
...
+
+**Capture groups create multiple panels**
+When your regular expression includes parentheses that form a capture group, the UI creates a separate panel for each unique value captured by that group.
+
+For example, the expression `(layer_0|layer_10)_loss` includes a capture group and will create two separate panels:
+1. One panel for metrics matching `layer_0`.
+2. One panel for metrics matching `layer_10`.
+
+**Non-capturing groups keep metrics together**
+To match multiple alternatives without creating separate panels, use a non-capturing group with the `?:` syntax. The expression `(?:layer_0|layer_10)_loss` matches the same metrics as the previous example, but displays them together in a single panel.
```
mdlinville:
```suggestion
- **Capture groups create multiple panels**
    When your regular expression includes parentheses that form a capture group, the UI creates a separate panel for each unique value captured by that group.

    For example, the expression `(layer_0|layer_10)_loss` includes a capture group and will create two separate panels:
    1. One panel for metrics matching `layer_0`.
    2. One panel for metrics matching `layer_10`.

- **Non-capturing groups keep metrics together**
    To match multiple alternatives without creating separate panels, use a non-capturing group with the `?:` syntax. The expression `(?:layer_0|layer_10)_loss` matches the same metrics as the previous example, but displays them together in a single panel.
```

models/app/features/panels/line-plot/reference.mdx
mdlinville: Done (here and in the main line plot page)

julia-g-rose (approved):
thank youuuu!

models/app/features/panels/line-plot.mdx
```diff
@@ -13,6 +15,77 @@ Line plots display by default for metrics logged with `wandb.Run.log()` over tim
...
 
+## Add a line plot
+
+This section shows how to create a line plot for a single metric or multiple metrics.
+
+<Tabs>
+<Tab title="Single-metric line plot">
+In an [automatic workspace](/models/app/features/panels#workspace-modes), a single-metric line plot is created automatically for each logged metric. Follow these steps to re-add a line plot that was deleted from an automatic workspace, or to add a line plot to a manual workspace.
+
+1. To add a line plot globally, click **Add panels** in the control bar near the panel search field.
```
anastasiaguspan, line 29: if someone gets to this page from Google, i feel like an additional step 1 is needed for both tabs.  Where am i in the UI?  To add a line plot on what page? like, i think this is for the workspace page, but as a newbie it took me a few to figure that out.   being more explicit would be good.

models/app/features/panels/line-plot.mdx
```diff
@@ -13,6 +15,77 @@ Line plots display by default for metrics logged with `wandb.Run.log()` over tim
...
+
+    To add a line plot directly to a section instead, click the section's action `...` menu, then click **+ Add panels**.
+1. To add a single-metric plot with default settings, click **Quick panel builder**.
+    1. In the **Single-key panels** tab, hover over a metric, then click **Add**. Repeat this step for each panel you want to add.
+    1. Click **Create &lt;number&gt; panels**.
+1. To add a custom line plot instead, click **Line plot**.
+    1. Configure the line plot's data, grouping, and display preferences using the corresponding tabs. For details, see [Edit line plot settings](#edit-line-plot-settings).
+    1. To add calculated expressions to the x or y axis, click **Expressions**. [JavaScript regular expressions](https://www.w3schools.com/js/js_regexp.asp) are supported.
+Select the type of panel to add, such as a chart. The panel's configuration details appear, with defaults selected.
```
anastasiaguspan: minor suggestion: The panel's configuration details appear with selected defaults.

models/app/features/panels/line-plot.mdx
```diff
@@ -13,6 +15,77 @@ Line plots display by default for metrics logged with `wandb.Run.log()` over tim
...
+1. Optionally, customize the panel and its display preferences. Configuration options depend on the type of panel you select. To learn more about the options for each type of panel, refer to the relevant section below, such as [Line plots](/models/app/features/panels/line-plot/) or [Bar plots](/models/app/features/panels/bar-plot/).
+    1. Click **Apply**.
```
anastasiaguspan: i like these two steps :) very clean way of ending the procedure succinctly but linking out to the varied options and complexity.

models/app/features/panels/line-plot.mdx
```diff
@@ -13,6 +15,77 @@ Line plots display by default for metrics logged with `wandb.Run.log()` over tim
...
+
+In an [automatic workspace](/models/app/features/panels#workspace-modes), a single-metric line plot is created automatically for each logged metric. This section shows how to create a single line plot that shows multiple metrics together, defined by a JavaScript regular expression. You can optionally consolidate many single-metric plots into a single multi-metric plot. This can improve the performance of a workspace with many logged metrics, and can help you analyze the results of your runs efficiently.
+
+1. To add a line plot globally, click **Add panels** in the control bar near the panel search field.
+
+    To add a line plot directly to a section instead, click the section's action `...` menu, then click **+ Add panels**.
+1. Click **Quick panel builder**, then click the **Multi-metric panels** tab.
+1. In **Regex** enter an expression in [JavaScript regular expression](https://www.w3schools.com/js/js_regexp.asp) format. The plot name will show the regular expression the plot uses, and the plot will include lines for all metrics that match the expression, including metrics logged in the future. As you type, the UI updates to show which metrics match the expression.
```
anastasiaguspan: The plot name shows the regular expression used by the plot, and the plot includes lines for all metrics that match the expression, including metrics logged in the future.

models/app/features/panels/line-plot.mdx
```diff
@@ -13,6 +15,77 @@ Line plots display by default for metrics logged with `wandb.Run.log()` over tim
...
+
+**Match specific metric variations**
+Use alternation to match only the metrics you want. For example, the non-capture group `(?:layer_0|layer_10)_loss` matches only the first and tenth layer losses, excluding intermediate layers.
+
+#### Understanding capture groups
+<LinePlotCaptureGroups/>
+
+</Tab>
+</Tabs>
+
```
anastasiaguspan, line 92: I know we were talking about starting a style guide, this page brought up an excellent example of something that has been bugging me visually - the tabbed controls in mintlify aren't 'visually distinguished' as to where they *end*.  when you choose single vs multiple, you can't tell where the 'multiple - related' text ends, and since it's a much longer tab content, it is a great example. Like, if you are scanning the page, and pick multiple, and then want to continue reading, it isn't visually clear if I'm still reading about 'multiple' or when I can start reading the 'next section', i have to click back to single to figure it out.   
that's a really long explanation to my thought that if we put a horizontal rule (---) at the conclusion of a </Tabs> set, there'd be a nice visual indicator?  So, not saying you should do that here, but that I want to bring it up for team discussion for the style guide :)

models/app/features/panels/line-plot.mdx
```diff
@@ -13,6 +15,77 @@ Line plots display by default for metrics logged with `wandb.Run.log()` over tim
...
+1. In **Regex** enter an expression in [JavaScript regular expression](https://www.w3schools.com/js/js_regexp.asp) format. The plot name will show the regular expression the plot uses, and the plot will include lines for all metrics that match the expression, including metrics logged in the future. As you type, the UI updates to show which metrics match the expression.
+1. To optionally remove duplicate single-metric panels when the multi-metric plot is created, toggle **Clean up auto-generated panels**. A preview shows which panels will be cleaned up. <Note>When this option is turned on, a single-metric plot will not be created for a newly logged metric that matches the expression. Instead, it will be included only in this multi-metric plot.</Note>
+1. Click **Create &lt;number&gt; panels**.
+
+### More about multi-metric regular expressions
+
+Multi-metric line plots use [JavaScript regular expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions) to match metric names. These section demonstrate some common use cases and gives more details about how the regular expressions work, such as how capture groups affect the panels that are created.
```
anastasiaguspan, line 61: You include this link again at the end of this section, yet, you use a different link text, implying it is a different reference.  I'd recommend trying to use the same link text, so it is clearer they don't have to click it if they already examined it, or removing one of them, since they are in the same section.
i like that you are giving some 'handy shortcut examples', but yeah, regex can become extremely complex and powerful, so , imo, in the start of the section you could say that it uses Javascript regular expressions, and not link,  and then at the end, if they want more power and control than what you provided, link out to the full regex world.

snippets/en/_includes/line-plot-capture-groups.mdx
```diff
@@ -0,0 +1,19 @@
...
+    For example, the expression `(layer_0|layer_10)_loss` includes a capture group and will create two separate panels:
+    1. One panel for metrics matching `layer_0`.
+    2. One panel for metrics matching `layer_10`.
+
+- **Non-capturing groups keep metrics together**
+    To match multiple alternatives without creating separate panels, use a non-capturing group with the `?:` syntax. The expression `(?:layer_0|layer_10)_loss` matches the same metrics as the previous example, but displays them together in a single panel.
+
+Here's the difference:
+- `(layer_0|layer_10)_loss` - Creates two panels, one for each layer
+- `(?:layer_0|layer_10)_loss` - Creates one panel showing both layers together
```
anastasiaguspan: Since these qualify as standalone sentences, i think Google says they should have periods a the ends of the bullets.

anastasiaguspan (approved):
Looks great!

models/app/features/panels/line-plot.mdx
mdlinville: Oof, look at that mis-indent. 🥇

models/app/features/panels/line-plot.mdx
mdlinville: Mintlify actually just dropped a fix for my feature request about this. <3

PR #2068, opened 2026-01-21
Artifacts: Rewrite text, mention artifact versions

ngrayluna:
## Description

* Updated code snippets
* Updated language, word smithed "Construct an artifact" page.
* Mentioned that each time you log an artifact (with same name) that it creates an artifact version.

## Related issues

- Fixes https://wandb.atlassian.net/browse/DOCS-1236


models/artifacts/construct-an-artifact.mdx
```diff
@@ -81,110 +91,121 @@ with wandb.init(project="<project>", job_type="<job-type>") as run:
...
+hello.txt
 ```
 
-Optionally, pass the desired path within the artifact for the `name` parameter.
+Optionally, pass a different name to the `name` parameter to rename the file within the artifact object itself:
 
 ```python
-artifact.add_file(local_path="path/file.format", name="new/path/file.format")
+artifact.add_file(
+    local_path="path/file.txt",
```
dbrian57: should this be `hello.txt`?

dbrian57 (approved):
one question inline. Looks good otherwise!

PR #2074, opened 2026-01-22
Experiments: Update experiment limits page

ngrayluna:
## Description

Clarify how steps, metrics, and logged points relate to performance when logging to W&B. This PR also:

- Defines how W&B counts logged data (steps, metrics, and logged points)
- Explains how logging scales at large step counts
- Disambiguates per-metric, per-run, and project-level limits
- Improves examples and comments for metric frequency and commit behavior
- Reorganizes section hierarchy to separate concepts, limits, and performance guidance
- Word smithing (active voice, etc. )

## Related issues

- Fixes [DOCS-12345](https://wandb.atlassian.net/browse/DOCS-1727)


tristan-spaulding2:
@ngrayluna I think the clarity is improved here, thanks,  You may want to check with  @dawita66 on the best guidance to give on the numbers right now.  For example as we roll out v2, it's no longer so cut and dried between SaaS and Dedicated/Customer Managed.  Some Dedicated do have v2 and with the mysql size (not a factor mentioned on this page, but relevant), can do well above 10,000 runs performantly.  Meanwhile non-paid SaaS do not have v2 on and can't do 100,000 runs.  etc.

models/track/limits.mdx
```diff
@@ -1,20 +1,87 @@
...
+The following sections describe recommended limits and performance considerations when logging experiment data to W&B.
 
-Keep your pages in W&B faster and more responsive by logging within the following suggested bounds.
+## How W&B counts logged data
 
-## Logging considerations
+W&B organizes logged data along three dimensions:
 
-Use `wandb.Run.log()` to track experiment metrics.
+* **Steps**: The number of time steps in a run that you finalize by committing logged data. Each step represents a single time index (such as a training step or epoch) and is finalized when you call `wandb.Run.log()` with `commit=True`, or implicitly when `commit` and `step` are not specified.
```
wai-ip, line 12: "The number of time steps in a run"  or just  "The number of steps in a run"

models/track/limits.mdx
```diff
@@ -1,20 +1,87 @@
...
+</Info>
+
+## Summary of recommended limits
+
+The following table summarizes recommended limits for logging at scale:
+
+| Dimension              | Guidance at scale                         |
+|-----------------------|--------------------------------------------|
+| Steps per run         | Millions of steps per run are common       |
+| Distinct metrics      | Fewer than 100,000 per run                 |
```
wai-ip, line 48: if a project has 10 runs, each run has 100K non-overlapping distinct metrics, then the project would have a total of 10 x 100K = 1M distinct metrics.  The project page would not be able to load that because the per project limit is 100,000 distinct metrics.

however, if all the runs log to the same 100K metrics, then it would be okay to say it's "100,000 per run".

would it be more accurate to say "Fewer than 100,000 per project"?

wai-ip: I am assuming this is for MT SaaS.  

For dedicated instances, the distinct metric limits vary by instance size from our internal sizing benchmark. We probably don't want to publish them may be worthwhile to point out they are different from this MT SaaS number.

ngrayluna:
Closing in favor of: https://github.com/wandb/docs/pull/2514

PR #2088, opened 2026-01-28
Fix cli ref intro whitespace

mdlinville:
## Description
Fix bug where CLI ref code blocks appear unstyled

This is happening because Mintlify doesn't support implicit code blocks
by indentation. This fix transforms them into code fences in the
generated references. The CLI code can't use code fences in the docstrings.

Fixes DOCS-1463

## Testing
- [x] Regenerated the refs with the fix
- [x] PR tests succeed

### DIFF
--- a/scripts/cli-docs-generator.py
+++ b/scripts/cli-docs-generator.py
@@ -34,6 +34,7 @@ import sys
 import argparse
 import shutil
 import json
+import inspect
 from pathlib import Path
 from typing import Dict, Any, List, Optional, Union
 import click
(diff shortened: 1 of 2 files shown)


scripts/cli-docs-generator.py
```diff
@@ -40,29 +40,97 @@
...
+    if indents:
+        min_indent = min(indents)
+        # Remove the common indentation from all lines
+        dedented_lines = []
+        for line in lines:
+            if line.strip():  # Non-empty lines
+                dedented_lines.append(line[min_indent:])
+            else:  # Empty lines
+                dedented_lines.append('')
+        text = '\n'.join(dedented_lines)
```
tonyyli-wandb: **suggestion:** would it be easier just to use [textwrap.dedent()](https://docs.python.org/3/library/textwrap.html#textwrap.dedent)?

Or since this is specifically for parsing docstrings: typically most docgen tools just call inspect.getdoc() or [inspect.cleandoc()](https://docs.python.org/3/library/inspect.html#inspect.cleandoc) (which necessarily dedents).  The latter takes a string as input, so it's more directly applicable here.

mdlinville: Great idea, done! Just pushed a significant refactor. Then I ran the script again to regenerate the refs, and there was no diff from the previous output.

PR #2094, opened 2026-01-29
Reformat some Args in the docs

travis-the-dragon:
Formatting tweaks to some **Args:**

## Description

As it says on the tin.


mdlinville (changes requested):
Sorry, this content is generated so you should not edit it in the docs repo. I think you'd need to adjust the docstrings in the `wandb/wandb` source.

ngrayluna:
Closing. I have a PR that fixes the issue. )

PR #2095, opened 2026-01-29
weave101updates-docs-1871

anastasiaguspan:
Simplifying introductory content (plus adding introductory material that was scraped from the now-being-deprecated Weave 101 AI Academy course), moving api fundamentals into their respective sections instead of introducing them at the start, and moving service API explanatory page down to the Service Ref section

For example:
## Testing
- [ ] PR tests succeed

## Related issues

- Fixes DOCS-1871


weave/concepts/what-is-weave.mdx
```diff
@@ -14,87 +14,98 @@ Weave addresses these challenges by providing:
 - **Version tracking** for prompts, models, and data so you can understand what changed
 - **Feedback collection** to capture human judgments and production signals
 
-## Debug with traces
+## The main threads of Weave
```
dbrian57: I think we should avoid headers immediately followed by sub-headers. Can we put an intro sentence in here that introduces what the threads of Weave are? Basically just say it contains the following feature set or whatever

weave/concepts/what-is-weave.mdx
```diff
@@ -14,87 +14,98 @@ Weave addresses these challenges by providing:
...
+### Traces
+Tracks end-to-end how a specific LLM application comes to its response.
 
-Weave automatically traces your LLM calls and shows them in an interactive UI. You can see exactly what went into each call, what came out, how long it took, and how calls relate to each other.
+- See inputs and outputs of each application usage.
+- See source documents used to produce the LLM feedback.
+- See cost, token count, latency of the LLM call.
```
dbrian57:
```suggestion
- See cost, token count, and latency of LLM calls.
```

docs.json
```diff
@@ -1040,6 +1027,7 @@
                   {
                     "group": "Service API",
                     "pages": [
+                      "weave/cookbooks/weave_via_service_api",
```
dbrian57, line 1033: Any pages we deleted here (e.g. tracking.mdx) we should alias at the bottom of this file (docs.json).

weave/guides/tracking/tracing.mdx
```diff
@@ -17,10 +17,35 @@ description: "Track and monitor your AI application's execution with Weave traci
...
 
-Calls are the fundamental building block in Weave. They represent a single execution of a function, including:
-- Inputs (arguments)
-- Outputs (return value) 
-- Metadata (duration, exceptions, LLM usage, etc.)
+## Ops
+
+An **Op** is a versioned, tracked function. When you decorate a function with `@weave.op()` (Python) or wrap it with `weave.op()` (TypeScript), Weave automatically captures its code, inputs, outputs, and execution metadata.
+
+```python
```
dbrian57: We should have a TS version of this example and any other applicable examples here.

weave/guides/tracking/tracing.mdx
```diff
@@ -17,10 +17,35 @@ description: "Track and monitor your AI application's execution with Weave traci
...
+import weave
+
+@weave.op()
+def generate_response(prompt: str) -> str:
+    # Your LLM logic here
+    return response
+```
+
+Ops are the building blocks of tracing, evaluation scorers, and any tracked computation.
```
dbrian57: I think we should just include this in the intro paragraph. It feels kinda orphaned below the code block

weave/concepts/what-is-weave.mdx
```diff
@@ -14,87 +14,98 @@ Weave addresses these challenges by providing:
 - **Version tracking** for prompts, models, and data so you can understand what changed
 - **Feedback collection** to capture human judgments and production signals
 
-## Debug with traces
+## The main threads of Weave
+### Traces
+Tracks end-to-end how a specific LLM application comes to its response.
```
dbrian57:
```suggestion
Track end-to-end how data flows through your LLM application.
```

"...how a specific LLM application comes to its response" 
I'm not sure we can technically say this because I don't think we really have observability into how models are arriving at their responses. I would just double check, otherwise, I would go with this suggestion or something like it

weave/concepts/what-is-weave.mdx
```diff
@@ -14,87 +14,98 @@ Weave addresses these challenges by providing:
...
+import weave
+client = weave.init('your-team/your-project-name')
+```
+
+```TypeScript Typescript
+import * as weave from 'weave';
+const client = await weave.init('your-team/your-project-name');
+```
+</CodeGroup>
+You're now ready to start using Weave!
```
dbrian57: I think we generally try to avoid this kind of statement and exclamation points 😅 . Instead, I would say something like, "Once you've configured the Weave client, you can start tracking your LLM calls," which I think segues nicely into the next step.

I also think that line somewhat contradicts the idea of a step 4.

weave/concepts/what-is-weave.mdx
```diff
@@ -14,87 +14,98 @@ Weave addresses these challenges by providing:
...
+- See cost, token count, latency of the LLM call.
+- Drill down into specific prompts and how answers are produced.
+- Collect feedback on responses from users.
+- In your code, you can use Weave [ops and calls](/weave/guides/tracking/tracing) to track what your functions are doing.
 
 [Get started with tracing](/weave/quickstart)
 
-## Evaluate systematically
+### Evaluations
+Systematic benchmarking of your application for you to know how good it is performing and be confident to deploy it to production.
```
dbrian57: Thoughts?
```suggestion
Systematically benchmark your LLM application's performance to ensure confidence when deploying to production."
```

imo, this is kind of an awkward sentence fragment. I think we can tighten it up into a full sentence with a clear action.

weave/concepts/what-is-weave.mdx
```diff
@@ -14,87 +14,98 @@ Weave addresses these challenges by providing:
...
-<Tab title="TypeScript">
-```bash
+
+```Typescript Typescript
 npm install weave
 ```
-</Tab>
-</Tabs>
+</CodeGroup>
+3. In your script, import Weave and initialize a project::
```
dbrian57:
```suggestion
3. In your script, import Weave and initialize a project:
```

weave/concepts/what-is-weave.mdx
```diff
@@ -14,87 +14,98 @@ Weave addresses these challenges by providing:
...
+```TypeScript Typescript
+import * as weave from 'weave';
+const client = await weave.init('your-team/your-project-name');
+```
+</CodeGroup>
+You're now ready to start using Weave!
 
-Both SDKs support tracing, evaluation, datasets, and the core Weave features. Some advanced features like class-based Models and Scorers are currently not available for the Weave TypeScript SDK.
+4. Weave integrates with popular LLM providers and frameworks.  When you use a [supported integration](/weave/guides/integrations/), Weave automatically traces LLM calls without additional code changes.
+However, to log traces to custom methods, add a one line decorator `weave.op` to any function. Works in development or production.
```
dbrian57:
```suggestion
However, you can still use Weave log traces for unsupported LLM providers by adding a one-line decorator (`weave.op`) to your call function. You can also use the decorator on any function in your app to track its input, output, and code.
```
I think this is more specific. I also am wary of using the word "method" unless I know for sure that the function is part of a class.

dbrian57 (changes requested):
Looking good! Few minor fixes. Requesting changes because I just want to make sure the redirects don't get missed.

weave/concepts/what-is-weave.mdx
anastasiaguspan: TIL!  thanks, wasn't aware of the method/function nuance in python.

weave/concepts/what-is-weave.mdx
anastasiaguspan: lol, so true.  admittedly I reckon this one got past me from something i copied. However, I could almost defend this particular one -- in light of Scott's emphasis on us conveying better how simple it is to get started and to hide the complexity. It doesn't, actually, contradict step 4, since if they are using an integration they really are all set.
All that being said, yep, i'll change it and clean this up

weave/guides/tracking/tracing.mdx
anastasiaguspan: ok, added, but this tracing page is already changing a lot from being split up and I'll need to merge this in - any additional tweaking we can address with that PR most ideally.

dbrian57 (approved):
Great work!

PR #2112, opened 2026-02-02
Experiments: Add note, update script on removing W&B env var

ngrayluna:
## Description

Customers reported issue running code snippet on local, single machine. This was caused from W&B-specific env vars that were already set. 

PR shows how to remove W&B env var (and which one). Also shows how to define run ID as an env var.

Some word smithing was done as well.

## Related issues

- Fixes https://wandb.atlassian.net/browse/DOCS-1289


models/track/log/distributed-training.mdx
```diff
@@ -166,32 +166,70 @@ Within the primary node, initialize a W&B run with [`wandb.init()`](/models/ref/
...
+import os
+
+# Primary node
+run = wandb.init( ... )
+os.environ['WANDB_RUN_ID'] = run.id
+
+# Worker node
+run = wandb.init( ... , id=os.environ['WANDB_RUN_ID'])
+```
+</Tip>
```
timoffex, line 189: This example feels misleading to me because it looks like those statements run in the same script. I'm not sure this form is practical, since it requires somehow starting up worker nodes from the primary node's script to pass on the generated run ID.

The environment variable `WANDB_RUN_ID` is read automatically by `wandb`, and generally users should not read `WANDB_` variables or use them for custom purposes. I also wouldn't advise anyone to modify `os.environ` at runtime, since changes might not be detected.

The essence of the original tip is to (1) generate your own ID and (2) set the `WANDB_RUN_ID` environment variable however you do it in the framework you're using to orchestrate the nodes. There's no general code sample we can provide as it's highly dependent on the user's setup.

models/track/log/distributed-training.mdx
```diff
@@ -166,32 +166,70 @@ Within the primary node, initialize a W&B run with [`wandb.init()`](/models/ref/
...
+Worker processes may inherit W&B-related environment variables (such as `WANDB_SERVICE`) from parent processes in distributed environments. This can cause 
+initialization conflicts.
+
+Clear inherited W&B environment variables (such as `WANDB_SERVICE`) before calling `wandb.init()` on worker nodes to avoid initialization conflicts:
+
+```python                                                                                                                                                   
+import os                                                                                                                                                   
+os.environ.pop('WANDB_SERVICE', None)                                                                                                                    
+```
+</Warning>
```
timoffex, line 221: Oof, this one might be necessary in some cases, but that's definitely a bug on our side. Not sure about including this in the docs as it's a super janky thing to ask users to do.

ngrayluna:
Closing in favor of: https://github.com/wandb/docs/pull/3044

PR #2119, opened 2026-02-04
[Weave] annotation queue docs-1869

anastasiaguspan:
## Description

Adding two new pages under "monitor and collect feedback" -> Annotate:
1. Set up annotation queues
2. Review items in an annotation queue

## Testing
- [ ] PR tests succeed

## Related issues

- Fixes DOCS-1869


weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+- **Manual trace scoring**: Have SMEs rate model outputs on correctness, quality, or style.
+- **Failure analysis**: Annotate failure modes (hallucinations, refusals, loops) to understand where your model breaks.
+- **Domain expert review**: Enable medical, legal, or safety experts to review content with a task-focused interface.
+- **Dataset creation**: Turn annotated traces into evaluation or training datasets.
+
+
+## End-to-end workflow
+
+As the engineer, you are responsible for the following lifecycle:
```
chance-wnb: nit: The term "engineer" could be ambiguous—does it encompass roles like Project Manager, Data Analyst, or QA? Often, "engineer" is interpreted narrowly to mean software development engineers specifically. Should we consider a more inclusive term to cover these roles broadly?

Alternatively, is this feature intentionally scoped to "engineer" in its strictest sense? If so, I suggest peer-reviewing this definition and make sure folks are on the same page.

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+Annotation fields define the feedback that the annotator provides for each trace item. Fields are reusable across queues and projects.
+
+Field types include:
+- Boolean judgments such as correctness or acceptability.
+- Numeric or integer values such as quality or confidence.
+- Categorical labels such as failure mode or intent.
+- Free-form text for qualitative feedback.
+
+Define all fields before creating a queue so they can be selected during queue setup.
```
chance-wnb: Nit: how about we highlight this with `<Tip>` ?

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+To create an annotation field:
+1. Navigate to [wandb.ai](https://wandb.ai).
+2. In the Weave sidebar menu, click **Annotate**.
```
chance-wnb: Friendly reminder: The **Annotate** might be hidden in the `More` menu item in case a project contains both models and weave traces usage.

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+To create an annotation queue:
+1. In the Weave sidebar menu, click **Annotate**.
+2. In the tab bar, click the **Queues** tab.
+3. In the Queues table toolbar, click **Create Queue**.
+4. In the **Create Annotation Queue** modal dialog, configure:
+    - **Queue name**: This is the queue name the annotator will select to complete their work.
+    - **Guidelines** (Optional): Any additional instructions for the annotator.
+5. Click **Next**.
+6. Click **Manage fields** and choose what Annotation Fields should be included in this review work. All existing Annotation Fields for the Project are available for selection.
```
chance-wnb: Nit: I guess we probably need some screenshot for this section. The other section has, this one doesn't which might render the doc unbalanced.

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+Traces can be added to an annotation queue directly from the **Traces** page:
+1. In the Weave sidebar menu, click **Traces**.
+2. In the Traces table toolbar, filter traces as needed (such as by hallucination scores, failure modes, or specific ops).
+3. In the table, select the traces you want annotated.
+4. In the table's action bar, click **Add to queue** to add the selected rows to an annotation queue.
+<Frame>
+  ![Traces table with multiple rows selected and a bulk action bar showing actions that apply to the selected rows, such as Add to queue and Add to dataset.](./imgs/AnnotationAddTraces.png)
+</Frame>
+5. In the **Add to annotation queue** modal dialog, configure:
```
chance-wnb, line 95: Nit: maybe we need a screenshot of this modal dialog.

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+## Monitor review progress
+
+Once the annotation queue has been created, and traces added, share the queue name with your annotator for them to begin their review. See [Review items in an annotation queue](/weave/guides/tracking/annotation-review) for details on the review process.
+
+To share a direct link to the annotation queue with an annotator:
+1. In the Weave sidebar menu, click **Annotate**.
+2. In the tab bar, click the **Queues** tab.
+3. In the Annotation Queues table, click the name of your queue to open the queue items.
+4. In the Queue header bar, click the link button to copy a direct link to this queue.  You can also copy the URL from the browser address bar.
```
chance-wnb, line 114: It might be hard to locate the link button simply by reading. can we have a screenshot to better indicate its location?

weave/guides/tracking/annotation-review.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+<Frame>
+  ![Annotation queue review interface for a single trace item, showing a trace information pane alongside an annotation form pane where reviewers submit structured feedback and move through the queue.](./imgs/AnnotationSubmitItem.png)
+</Frame>
+
+<Tip>
+In the trace pane header, you can click the Trace ID suffix (for example, "b3ea") to open a panel showing the full trace details for the call.
+</Tip>
+
+If any annotator has already reviewed an item in the queue, both pane headers of the review interface are decorated with "Has feedback" and "Response submitted" indicators.
```
chance-wnb: I feel it is necessary to explain to the user what is going to happen if a queue item has already been annotated?

1. Can another user still give feedback(annotate) to the same queue item?
2. What is the implications if another user still give feedback? how would it impact the current queue reviewing progress?

chance-wnb (commented):
Thanks, overall great work. I shared some initial thoughts from an feature author perspective. Maybe you also want a doc team member to sign off to ensure it is compliant to the doc team's standards.

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: this is a great point - i will have a chat with my team about 'personas' and how we refer to our audiences.

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: good point, will add this.

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: I don't think it qualifies as a "Tip" since it's required - but if you'd prefer this statement be highlighted I can make it a Note.

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: Yep, will do.

weave/guides/tracking/annotation-review.mdx
anastasiaguspan: I agree the Docs should clarify this.  But, do we know the answers or has this behavior not been sorted out yet?

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: Generally, we (Docs) try to avoid (or are trying more to avoid) screenshots for the sake of screenshots.  They are hard to maintain and know when they are out of date.  Thus, I had some extra screenshots but ultimately chose to hold back the ones that I didn't think added true value (e.g. showed something non-intuitive).
That being said, if you or others really feel that additional ones are needed, we can certainly add them.

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: See above about screenshots - trying to not add lots unless they help explain something that might not be obvious.  I thought the modal dialog was intuitive and easy to follow. if there's some aspect you think a screenshot may make more clear, let's discuss.

weave/guides/tracking/annotation-queues.mdx
chance-wnb: Sure, as long as you are intentionally avoid doing that for a reason. It is good.

weave/guides/tracking/annotation-queues.mdx
chance-wnb: Sounds good.

weave/guides/tracking/annotation-review.mdx
chance-wnb: Yes the answers are (at the moment):

1. Yes.
2. Only the first annotation will advance the queue’s overall progress. Later users may still leave feedback on completed items, but their comments will be separate—they will not alter or remove the original feedback, nor will they change the item’s status. We do not track subsequent edits, aside from allowing users to submit feedback on a completed call. This maintains consistency, since users can already freely edit their own feedback from the call details pane even without the annotation feature.

weave/guides/tracking/annotation-queues.mdx
chance-wnb: Yes, lets high-lighten it with Note.

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: Advice was to avoid naming them unless necessary. I reworked the two sentences on this page referring to 'engineer' so that they no longer needed to be labeled.

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: I only added it on the first one though, just like only the first one indicates to 'navigate to wandb.ai', the rest of the procedures on the same page assume that.

weave/guides/tracking/annotation-review.mdx
anastasiaguspan: So question about2:  If i navigate to like Item 2 of 5, and someone else already had submitted feedback for that item (thus it has the "Has feedback" and "Response submitted" indicators)... what does the submit button say: does is say 'Submit' or does it say 'Edit feedback'?  Hopefully it says 'Submit' based on what you describe above (and i know that if **I** am the one that had already entered that feedback that it would say 'Edit Feedback' which makes sense).

chance-wnb: You have a good point here, however the current implementation doesn't make a distinction yet and always say "Edit feedback". It can be a  logical defect we tolerate, otherwise lets backlog a bug for now.

anastasiaguspan: okay, i guess I should add a Note then to the docs that if you were not the annotator to originally submit the feedback, that this button will not edit the feedback but will instead make an additional feedback entry for this item.

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
+---
+title: "Set up annotation queues"
+description: "Create annotation queues, route traces to domain experts, and export structured feedback."
+---
+
+## Overview
+
+Annotation queues let you route selected traces to domain experts for structured review without requiring them to navigate the full Weave UI. You define what feedback is collected, select which traces require review, and later can export completed annotations for analysis or dataset creation.
```
dbrian57:
```suggestion
Annotation queues let you route selected traces to domain experts for structured review without requiring them to navigate the full Weave UI. You define what feedback is collected, select which traces require review, and later can export completed annotations for analysis or dataset creation. Use cases can include:
```
Doesn't have to be this exact clause, but I think we should introduce what the bullet points are.

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+The following workflow summarizes how to use annotation queues to obtain reviews:
+
+1. Define annotation fields.
+2. Create an annotation queue.
+3. Load traces into the queue for review.
+4. Monitor progress while domain experts complete reviews.
+5. Filter and export completed annotations.
+
+
+## Define annotation fields
```
dbrian57, line 29: Feel like we need to denote somehow that we're starting the workflow here, because the prior two sections are informational and then we just kinda jump into the first sub-task of the setup without any warning.

I would approach it by adding a new header above "Define annotation fields" with an introductory sentence, then set the remaining sub-tasks as H3s below it, but I also see you're using H3s to introduce other nested-tasks, which would nest them further as H4s, and I don't know how the TOC on the right would handle that. 

Maybe just add a sentence at the beginning of this section that says, "To start creating an annotation queue, you need to define some annotation fields first. Annotation fields define the..."

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+
+Annotation fields define the feedback that the annotator provides for each trace item. Fields are reusable across queues and projects.
+
+Field types include:
+- Boolean judgments such as correctness or acceptability.
+- Numeric or integer values such as quality or confidence.
+- Categorical labels such as failure mode or intent.
+- Free-form text for qualitative feedback.
+<Note>
```
dbrian57: I don't think this needs to be a note. It's necessary to the workflow and not really a caveat, pro tip, or additional insightful information. 

Also, sorry in advance, I work to eliminate Note boxes as if they've offended me because there's a growing amount of evidence that users just gloss over them. It's totally a byproduct of everyone's brains being trained to navigate ad-ridden minefields of news story content. 😅

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+Define all fields before creating a queue so they can be selected during queue setup.
+</Note>
+
+To create an annotation field:
+1. Navigate to [wandb.ai](https://wandb.ai) and select your project.
+2. In the Weave project sidebar, click **Annotate**.  If you don't see **Annotate**, it might be nested in the menu under **More**.
+3. In the tab bar, click the **Fields** tab.
+4. In the Fields table toolbar, click **New Field**.
+5. In the **Create annotation field** modal dialog, configure:
+    - **Type**: Boolean, Integer, Number, String, or categorical options
```
dbrian57: Should we describe these fields? I know they're kinda self-explanatory. I always do, but I've often wondered whether we actually should.

The Google style guide has a slightly different approach in the last example here: https://developers.google.com/style/ui-elements#term-windows

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+- A set of annotation fields.
+- Guidelines that provide task instructions for annotators.
+- A collection of trace items awaiting review.
+
+To create an annotation queue:
+1. In the Weave project sidebar, click **Annotate**.
+2. In the tab bar, click the **Queues** tab.
+3. In the Queues table toolbar, click **Create Queue**.
+4. In the **Create Annotation Queue** modal dialog, configure:
+    - **Queue name**: This is the queue name the annotator will select to complete their work.
```
dbrian57:
```suggestion
    - **Queue name**: This is the queue name the annotator selects to complete their work.
```

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+    - **Guidelines** (Optional): Any additional instructions for the annotator.
+5. Click **Next**.
+6. Click **Manage fields** and choose what Annotation Fields should be included in this review work. All existing Annotation Fields for the Project are available for selection.
```
dbrian57:
```suggestion
6. Click **Manage fields** and choose what Annotation Fields to include in this review work. All existing Annotation Fields for the Project are available for selection.
```

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+<Frame>
+  ![Annotation Queues page displaying a table of queues in the project, including queue names, descriptions, and review progress to track annotation workflows.](./imgs/AnnotationQueues.png)
+</Frame>
+
+
+Creating an annotation queue defines the fields and guidelines for evaluation, but you still need to add traces to the queue to identify what data should be evaluated.
+
+### Add traces to a queue
+
+Traces can be added to an annotation queue directly from the **Traces** page:
```
dbrian57:
```suggestion
To add traces to an annotation queue directly from the **Traces** page:
```
Makes this a little more active

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+1. In the Weave project sidebar, click **Traces**.
+2. In the Traces table toolbar, filter traces as needed (such as by hallucination scores, failure modes, or specific ops).
+3. In the table, select the traces you want annotated.
+4. In the table's action bar, click **Add to queue** to add the selected rows to an annotation queue.
+<Frame>
+  ![Traces table with multiple rows selected and a bulk action bar showing actions that apply to the selected rows, such as Add to queue and Add to dataset.](./imgs/AnnotationAddTraces.png)
+</Frame>
+5. In the **Add to annotation queue** modal dialog, configure:
+    - **Queue Name**: In the dropdown, select the name of the existing queue to add these traces to.
```
dbrian57:
```suggestion
    - **Queue Name**: In the list, select the name of the existing queue to add these traces to.
```
Google Style Guide recommends not using dropdown and using either "list" or "menu' instead: https://developers.google.com/style/word-list#drop-down

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+        - **Outputs**: Select which trace output fields to show during annoation.
+6. Click **Add ***[Count]*** traces to annotation queue** to assign these traces as a part of the annotation queue review.
+
+
+When adding traces, you control which trace inputs and which outputs or model responses are reviewed. This way you can present annotators with only the context needed to make accurate judgments.
+
+
+## Monitor review progress
+
+Once the annotation queue has been created, and traces added, share the queue name with your annotator for them to begin their review. See [Review items in an annotation queue](/weave/guides/tracking/annotation-review) for details on the review process.
```
dbrian57:
```suggestion
Once you have created the annotation queue and added traces to it, share the queue name with your annotator for them to begin their review. See [Review items in an annotation queue](/weave/guides/tracking/annotation-review) for details on the review process.
```

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+In the Annotation Queues table, the **State** column indicates reviewing progress:
+- **Not started**: Queue has items but no annotations have been submitted.
+- **In progress**: At least one item has been reviewed.
+- **Completed**: All items have been reviewed.
+
+In the Annotation Queues table, the **Calls with responses** column indicates the percent of items (out of the total number of **Calls**) that have at least one submitted review.
+
+## Filter and export annotations
+
+Completed annotations are stored as structured metadata on traces.
```
dbrian57:
```suggestion
Weave stores completed annotations as structured metadata on traces.
```

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+You can:
+- Filter traces by queue assignment and annotation completion.
+- Save filtered views for reuse.
+- Export annotated traces to datasets for evaluation or training workflows.
+
+This connects expert human feedback directly to model evaluation and iteration.
+
+### Filter annotated traces
+
+1. In the Weave project sidebar, click **Traces**.
```
dbrian57, line 140: Need an introductory clause here to set off these directions. I know it kinda seems redundant, but it's the style used across the entire website

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+2. In the Traces table toolbar, click **Filter**.
+3. Add three values to a filter row:
+    - For **Column**, type "Queue", then press **Enter**.
+    - For the second dropdown, choose **Text: "is"**.
+    - For **Select a queue**, choose your annotation queue name.
+4. To also filter on 'completed' queue items only, click **+ Add Filter**:
+    - For **Column**, type "feedback".  A popup will populate with Annotations and include your Annotation Field names.  Choose a required Field from your queue.
```
dbrian57:
```suggestion
    - For **Column**, type "feedback".  A dialog populates with Annotations and includes your Annotation Field names.  Choose a required Field from your queue.
```
https://developers.google.com/style/word-list#pop-up

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+2. In the Traces table, select the traces that you want to export.
+3. In the table toolbar, click **Add to dataset**. Follow the on-screen prompts to complete the addition.
+
+To learn more about using datasets, see [Collect and track datasets](/weave/guides/core-types/datasets).
+
+#### Access annotations programmatically
+
+If you want to integrate annotations programmatically, you must know your project name and queue ID:
+ - **Project**: The W&B project name (can be project or team/project). If you don’t specify a W&B team (such as "team/project"), your default team is used.
+ - **Queue ID**: The queue ID is the unique identifier for an annotation queue.
```
dbrian57:
```suggestion
 - **Queue ID**: The annotation queue's unique identifier.
```

weave/guides/tracking/annotation-queues.mdx
```diff
@@ -0,0 +1,207 @@
...
+https://wandb.ai/.../annotation-queues/019c0f63-7acb-7497-8f87-08873368fcd4
+```
+
+In this example, the queue ID is:
+
+019c0f63-7acb-7497-8f87-08873368fcd4
+
+You can iterate through the traces (calls) in your queue using the following code.
+
+```python
```
dbrian57:
```suggestion
```python lines
```

Add line numbers to the code example.

weave/guides/tracking/annotation-review.mdx
```diff
@@ -0,0 +1,45 @@
+---
+title: "Review items in an annotation queue"
+description: "Evaluate trace items and submit structured feedback using a simplified review interface."
+---
+
+
+Annotation queues provide a focused review interface for domain experts. You review one item at a time, examine the provided context, and submit structured feedback using predefined fields.
+
+You do not need to understand the underlying model or tracing system to complete reviews.
```
dbrian57: I feel weird about this line, because my first question after reading it was, "why?" but you kinda answered that in the opening line. Maybe we can add some more context here about how the queue interface is designed for external stakeholders or something.

It's not a super big deal; it just kinda caught on the edges of my brain.

weave/guides/tracking/annotation-review.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+
+## Annotation workflow
+
+As an annotator, your task consists of:
+
+1. Open the shared annotation queue.
```
dbrian57:
```suggestion
1. Opening the shared annotation queue.
```
gerund agreement with the prior clause. We should do this for all three

weave/guides/tracking/annotation-review.mdx
```diff
@@ -0,0 +1,45 @@
...
+2. Review each item. You can move forward and backward through items in the queue at any time.
+3. Submit feedback using the provided annotation fields.
+
+Your progress is saved automatically after you submit an item. If you need to pause and resume your work later, the review automatically resumes at the first incomplete item. Once all items have been reviewed, the Annotation Queue table reflects completion of the work.
```
dbrian57:
```suggestion
Weave saves your progress automatically after you submit an item. If you need to pause and resume your work later, the review automatically resumes at the first incomplete item. Once all items have been reviewed, the Annotation Queue table reflects completion of the work.
```

weave/guides/tracking/annotation-review.mdx
```diff
@@ -0,0 +1,45 @@
...
+  ![Annotation queue review interface for a single trace item, showing a trace information pane alongside an annotation form pane where reviewers submit structured feedback and move through the queue.](./imgs/AnnotationSubmitItem.png)
+</Frame>
+
+<Tip>
+In the trace pane header, you can click the Trace ID suffix (for example, "b3ea") to open a panel showing the full trace details for the call.
+</Tip>
+
+If any annotator has already reviewed an item in the queue, both pane headers of the review interface are decorated with "Has feedback" and "Response submitted" indicators.
+
+You can still provide feedback for an item even if it has already been reviewed by someone else.  In this case, although the feedback submission button will still say **Edit feedback**, it will create a new feedback entry with your responses.
```
dbrian57: I feel like I would need to be more familiar with the UI's behavior to rewrite this, but I think we should try to rewrite the second line to not use the future tense.

weave/guides/tracking/annotation-review.mdx
```diff
@@ -0,0 +1,45 @@
...
+<Note>
+If you submit feedback to an item already reviewed by someone else, your feedback will be stored **in addition to** the original annotator's feedback.
```
dbrian57: Again, I don't this this needs to be a note and can be rewritten to use present tense and more active voice, like this:

```suggestion
If you submit feedback on an item that has already been reviewed by someone else, Weave stores the feedback **in addition to** the original annotator's feedback. It does not edit or overwrite the original feedback.
```

dbrian57 (approved):
This is really well structured and thorough! Minor stuff inline.

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: concur, lol

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: not sure what on that link I was supposed to focus on.  regarding explaining boolean, integer, etc,... I'd lean towards that being 'expected knowledge' for our audience.  Though yeah, if we decide to align on including that level of detail I don't feel strongly either way.  (tho i think i'll leave this this way for now and circle back if we decide we should)

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: i made the change, cuz google says so, but let the record show I strongly dislike this :)
saying 'list' makes me think it is something displayed on the page I should look for, 'dropdown' , to me, implies I need to take an action in order to see it (e.g. click)....

weave/guides/tracking/annotation-review.mdx
anastasiaguspan: we want this imperative, since it's steps. so instead I changed the lead in text to make that more clear and standard

weave/guides/tracking/annotation-review.mdx
anastasiaguspan: I fixed the tense, but for now i'm leaving the note because 1. in this case it is intentionally redundant to the printed text (said a different way) - e.g. if they really are scanning right past note boxes they are still safe.  
if it wasn't a note box, it would have to be removed because it is redundant - but in this case we need to highlight this because it is kinda like a  'known bug' that they cannot currently address.  so i'm trying to catch both cases, those that only read the important notes and those that skip them.  (we did this a lot in AWS, knowing the same fact you mentioned, some people don't see note boxes.  But general sentiment at AWS was there is a subclass that skims so hard they *only* see note boxes, so for important stuff we kinda did both.  happy to chat this out further and align with the team.)

weave/guides/tracking/annotation-queues.mdx
anastasiaguspan: or maybe you meant the name and description.. yeah, seemed too obvious, but anyway, i went and added them now for completeness

chance-wnb:
Hi @anastasiaguspan so sorry to deliver this message. but in a recent PR, we are going to rename

`Has feedback` -> `Has annotation`
`Edit feedback` -> `Edit annotation`
`Confirm feedback` -> `Confirm annotation`

Change is made from https://github.com/wandb/core/pull/39315

PR #2132, opened 2026-02-09
Sync localized navigation with English

johndmulhausen:
Translated docs navigation fix. Also checks in the script that generates fresh localized docs navigation based on a remote docs.json (presumably not the one in main, but one from a point in time, when the translation branch was originally cut)

## Description
- Add a script to mirror English navigation into ja and ko using a local or remote docs.json.
- Prune localized nav entries that do not exist on disk and report unreferenced localized files.
- Update docs.json to reflect the synchronized ja and ko navigation.

## Testing
- [ ] Not run (documentation and script changes only).

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -1239,62 +1239,73 @@
         "language": "ja",
         "tabs": [
           {
-            "tab": "Home",
-            "pages": []
-          },
-          {
             "tab": "Platform",
+            "icon": "/icons/cropped-artifacts.svg",
             "pages": [
+              "ja/index",
+              "ja/get-started",
               {
-                "group": "\u30c7\u30d7\u30ed\u30a4\u30e1\u30f3\u30c8\u30aa\u30d7\u30b7\u30e7\u30f3",
+                "group": "Deployment options",
                 "pages": [
                   "ja/platform/hosting",
+                  "ja/platform/hosting/hosting-options/multi_tenant_cloud",
                   {
-                    "group": "\u30c7\u30d7\u30ed\u30a4\u30e1\u30f3\u30c8\u30aa\u30d7\u30b7\u30e7\u30f3",
+                    "group": "Dedicated Cloud",
                     "pages": [
-                      "ja/platform/hosting/hosting-options/multi_tenant_cloud",
-                      {
-                        "group": "\u81ea\u5df1\u7ba1\u7406",
-                        "pages": [
-                          "ja/platform/hosting/hosting-options/self-managed",
-                          "ja/platform/hosting/self-managed/ref-arch",
-                          {
-                            "group": "W&B \u30b5\u30fc\u30d0\u30fc\u3092 Kubernetes \u3067\u5b9f\u884c\u3059\u308b",
-                            "pages": [
-                              "ja/platform/hosting/operator",
-                              "ja/platform/hosting/self-managed/operator-airgapped"
-                            ]
-                          },
-                          {
-                            "group": "\u30d1\u30d6\u30ea\u30c3\u30af \u30af\u30e9\u30a6\u30c9\u306b\u30a4\u30f3\u30b9\u30c8\u30fc\u30eb\u3059\u308b",
-                            "pages": [
-                              "ja/platform/hosting/self-managed/aws-tf",
-                              "ja/platform/hosting/self-managed/gcp-tf",
-                              "ja/platform/hosting/self-managed/azure-tf"
-                            ]
-                          },
-                          "ja/platform/hosting/self-managed/bare-metal",
-                          "ja/platform/hosting/server-upgrade-process"
-                        ]
-                      },
-                      {
-                        "group": "\u5c02\u7528\u30af\u30e9\u30a6\u30c9",
-                        "pages": [
-                          "ja/platform/hosting/hosting-options/dedicated_cloud",
-                          "ja/platform/hosting/hosting-options/dedicated_regions",
-                          "ja/platform/hosting/export-data-from-dedicated-cloud"
-                        ]
-                      }
+                      "ja/platform/hosting/hosting-options/dedicated_cloud",
+                      "ja/platform/hosting/hosting-options/dedicated_regions",
+                      "ja/platform/hosting/export-data-from-dedicated-cloud"
                     ]
                   },
                   {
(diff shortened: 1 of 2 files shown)


scripts/sync_localized_navigation.py
```diff
@@ -0,0 +1,279 @@
...
+    with urlopen(url) as response:
+        payload = response.read().decode("utf-8")
+    return json.loads(payload)
+
+
+def main() -> None:
+    args = parse_args()
+    docs_json_path = Path(args.docs_json)
+    languages = [lang.strip() for lang in args.languages.split(",") if lang.strip()]
+    en_source: dict[str, Any] | None = None
```
ngrayluna, line 256: This is not very Python-y...we don't normally declare variables (it makes sense if this was in a function definition.

scripts/sync_localized_navigation.py
```diff
@@ -0,0 +1,279 @@
...
+
+import argparse
+import copy
+import json
+from pathlib import Path
+from typing import Any, Iterable
+from urllib.request import urlopen
+
+
+def prefix_path(path: str, language: str) -> str:
```
ngrayluna, line 39: For future prompts, please add function docstrings as well.

scripts/sync_localized_navigation.py
```diff
@@ -0,0 +1,279 @@
...
+        if (repo_root / f"{page}{ext}").is_file():
+            return True
+        if (repo_root / page / f"index{ext}").is_file():
+            return True
+
+    return False
+
+
+def page_from_file(path: Path) -> str | None:
+    if path.suffix not in {".mdx", ".md", ".ipynb"}:
```
ngrayluna: Do we have .ipynb in our docs? If nay, we can remove the check for ipynb

scripts/sync_localized_navigation.py
```diff
@@ -0,0 +1,279 @@
...
+        return str(path.parent.as_posix())
+    return str(path.with_suffix("").as_posix())
+
+
+def collect_nav_pages(pages: list[Any]) -> set[str]:
+    collected: set[str] = set()
+    for item in pages:
+        if isinstance(item, str):
+            if not is_external_path(item):
+                collected.add(item)
```
ngrayluna, line 81: Suggestion: You can consolidate this into one statement. IF ... AND NOT ... like you do in the following check

ngrayluna (commented):
Quick review of some tech debt considerations.

scripts/sync_localized_navigation.py
```diff
@@ -0,0 +1,279 @@
...
+        if isinstance(item, dict) and isinstance(item.get("pages"), list):
+            collected.update(collect_nav_pages(item["pages"]))
+    return collected
+
+
+def list_language_files(repo_root: Path, language: str) -> set[str]:
+    language_root = repo_root / language
+    if not language_root.is_dir():
+        return set()
+    pages: set[str] = set()
```
ngrayluna, line 92: See comment about declaring variables outside of a function or Class def. Not a pattern you see in Python. 

You can acheinve the same using:

```python
pages = set()
```

scripts/sync_localized_navigation.py
johndmulhausen:
```suggestion
    if path.suffix not in {".mdx", ".md"}:
```

docs.json
```diff
@@ -1238,71 +1238,82 @@
...
-          },
           {
             "tab": "Platform",
+            "icon": "/icons/cropped-artifacts.svg",
             "pages": [
+              "ja/index",
+              "ja/get-started",
               {
-                "group": "\u30c7\u30d7\u30ed\u30a4\u30e1\u30f3\u30c8\u30aa\u30d7\u30b7\u30e7\u30f3",
+                "group": "Deployment options",
```
mdlinville, line 1248: Is this a regression of the translated section name? Ditto for other pre-existing sections below.

docs.json
```diff
@@ -1238,71 +1238,82 @@
...
+                      },
+                      "ja/platform/hosting/self-managed/bare-metal",
+                      "ja/platform/hosting/server-upgrade-process",
+                      "ja/platform/hosting/self-managed/disable-automatic-app-version-updates"
+                    ]
+                  }
+                ]
+              },
+              {
+                "group": "Configure W&B",
```
mdlinville, line 1288: The updated translation tooling or process used by the translators will need to identify new groups that need translation.

ngrayluna (approved):
Approving for time sake. Please do another pass for tech debt considerations (stuff I called out and other stuff I didn't catch).

PS
Would love a small post mortem to explain why the JA and KO docs shipped with non updated docs.json.

mdlinville (commented):
Reviewed `docs.json`:
- Updates to `pages` keys look good
- Updates to `group` values look like regressions. It might be that each of these group names has changed since the last translation, but I sort of doubt it. It looks like this replaces unicode characters (which I assume are translated) with English text.

docs.json
johndmulhausen: I don't know whether those characters actually are the translation for "Deployment options"

mdlinville:
> Approving for time sake. Please do another pass for tech debt considerations (stuff I called out and other stuff I didn't catch).
> 
> PS Would love a small post mortem to explain why the JA and KO docs shipped with non updated docs.json.

**TL;DR**:  From my POV:
1. Translation tooling has not yet been updated for Mintlify. The initial PR commit was produced using this tooling.
2. The PR was raised in a fork, with no PR preview.
3. Matt speaks only English.
4. To resolve the issues and get the content to build, I worked locally to convert the translated output to Mintlify.
    - I was primarily focused on getting the actual Markdown converted,  making sure that the snippets were propagated, and fixing the Korean content to use Korean snippets rather than the English ones.
    - Once `mint dev` worked, I focused on fixing some broken links we had never seen before, because the translated content is not subject to the site-wide link check and the PR link checker only checks files changed by the PR.
    - **I completely forgot to consider `docs.json`. Since it wasn't updated, it wasn't part of the PR I was reviewing, and since I don't read Japanese or Korean, I didn't notice any problems in the output.**
5. We were facing difficult time zone skew, so when the PR looked to me to be ready, I went ahead and merged it.
6. The next day when @hw-wandb reported issues, and John and I did not initially understand that the actual problem was missing pages in the translated TOCs. I was also unexpectedly out sick so I couldn't help troubleshoot or diagnose the problem directly.

PR #2144, opened 2026-02-11
Weave: Fixes example links across Weave docs

dbrian57:
## Description
We used to link out to several pieces of example code that lived in the Weave repo. Those examples have since moved over here and I'm updating the links for them.


anastasiaguspan (approved):
Yes, this fixes it. thanks for finding where it went

PR #2151, opened 2026-02-12
Inference - use codegen and deprecate 3 models

jamie-rasmussen:
## Description

Extending the work done in https://github.com/wandb/docs/pull/2146 there are now four pages where table or list content blocks can be programmatically updated as part of the model lifecycle process, reducing the time and labor required from the docs team and ensuring consistency between the docs and other places like the model catalog UI.

The codegen blocks are of the form
`{/* takeru <id> - This [table|list] is automatically generated, do not edit manually. */}`

Some slight tweaks from the first version of the script include adding a `(deprecated)` suffix in the large available models table and sorting those models to the bottom. I have also switched to a model name sorting method that is not purely alphabetical but sorts in a descending manner for integer pieces - this results in the newer and bigger models sorting before older and smaller versions.

As a test for the codegen, this PR also includes a content change, adding three more models to the deprecated state. I also added a "Retirement date" column to that table, which was feedback from the team.

This also restores the `Qwen2.5 14B Instruct` model to the available models - it was previously deprecated and so got accidentally filtered out.

## Testing
- [x] PR tests succeed


inference/models.mdx
anastasiaguspan:
```suggestion
When using the API, specify the model using its `Model ID` from the table above. For example:
```

anastasiaguspan (approved):

PR #2154, opened 2026-02-17
Lychee improvements and a script to fix anchors in translated content

mdlinville:
## Description
- Improve Lychee config
    - Remove `include` since it turns off how `exclude` regex matcing (!).The 'scheme' key is sufficient.
    - Improve regex to match W&B prod URLs.
    - Exclude `mintlify.app` preview URLs since those links are checked by `mint validate` and cause `mint dev` to fail
- Fix broken Weave link discovered in the process
- Fix newly-broken anchors in localized content after recent Translation refresh.

    Mintlify composes heading slugs algorithmically from the heading title, in both Markdown and HTML headings. There is no support for setting the heading slug manually, even by setting `id` on a HTML heading. This means that translated content will have different slugs than the English content, which will break anchor-level links in translated content.
    
    This script syncs deep links in translated content with the translated headings. A translator would run it toward the end of the translation refresh, perhaps just before creating the PR.
        - Full docs are in the README
        - Extracting and rewriting are handled separately
        - Supports dry-run and verbose modes
        - Supports operating on a specific translationor a specific file or all translations
- Run the script to validate the functionality. I first ran it in dry-run mode, then ran it for real and committed the result.
    ```bash
    python3 scripts/localized-fragment-links/rewrite_fragment_links.py --verbose
    ```
  - Before: `mint broken-links` reports: `found 1765 broken links in 512 files` (See https://github.com/wandb/docs/actions/runs/22117148644/job/63928073034#step:8:39)
  - After: `mint broken-links` reports: `success no broken links found` (See https://github.com/wandb/docs/actions/runs/22160011726/job/64074066310#step:8:192)

## Testing
- [x] Locally, the equivalent Lychee command to the Github Action works as intended against the site at `localhost:3000`.
- [x] Local build succeeds without errors (`mint dev`) (validates the Weave fix)
- [x] PR tests succeed


weave/guides/evaluation/monitors.mdx
```diff
@@ -43,7 +43,7 @@ To create a monitor in Weave:
 
 You can also [compare](/weave/guides/tools/comparison) and visualize the monitor's trace data in the Weave UI, or download it in various formats (such as CSV and JSON) using the download button (<Icon icon="download" iconType="regular" />) in the Traces tab. 
 
-Weave automatically stores all scorer results in the [Call](/weave/concepts/fundamental-components#calls) object's `feedback` field.
+Weave automatically stores all scorer results in the [Call](/weave/guides/tracking/tracing#calls) object's `feedback` field.
```
dbrian57, line 46: 👍

mdlinville:
This is passing everything, and I'm not sure what additional testing we can really do before the next translation refresh. So here we go. 🚀

PR #2210, opened 2026-03-02
OpenAI agent ts docs 1934

anastasiaguspan:
## Description

Documentation for TypeScript SDK: OpenAI agents Integration

I have tested this 'getting started' code example but it *does* require 'await weave.instrumentOpenAIAgents();' to work.

if that's accurate, Proposing we remove these lines under Section "manual instrumentation":
In most cases, automatic instrumentation handles everything. If you need manual control, for example when using dynamic imports or bundlers that bypass module hooks, you can call `instrumentOpenAIAgents()` explicitly:

```typescript
import * as weave from "weave";

await weave.init("openai-agents");
await weave.instrumentOpenAIAgents();
```

## Testing
- [ ]x Local link check succeeds without errors (`mint broken-links`)
- [ ] PR tests succeed

## Related issues

- Fixes DOCS-1934


weave/guides/integrations/js.mdx
```diff
@@ -18,29 +18,73 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+    
+Because ESM loads modules differently than CommonJS, Weave must be loaded before other modules to enable automatic instrumentation. Using the `import` flag ensures Weave is loaded before other modules so it can [patch](/weave/guides/integrations) them.
+
+### If you're unsure which type of project you have
+
+If you are running a TypeScript file directly with a tool like:
+```
+npx tsx test.ts
+```
+your project may still be using ESM depending on your `package.json` configuration.
```
chance-wnb: Thanks for mentioning this. However, I assume that often the case is that people don't have a `package.json` ahead of time but still want to run `npx tsx *`. In this case, is `"type": "module"` still the default case I wonder? 

anyway, we could suggest to our customers that explicitly having a package.json is the best practice.

weave/guides/integrations/js.mdx
```diff
@@ -18,29 +18,73 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+
+   ```bash
+   npx tsc
+   ```
+   This assumes your `tsconfig.json` outputs compiled files to a `dist` directory, for example:
+
+   ```json
+   {
+     "compilerOptions": {
+       "rootDir": ".",
```
chance-wnb, line 130: Actually, to guarantee that the project is compiled into ESM. I usually have to add this:

```
    "target": "es2022",
    "module": "nodenext",
    "moduleResolution": "nodenext",
```

[References 1](https://www.typescriptlang.org/tsconfig/#target)
[References 2](https://www.typescriptlang.org/docs/handbook/modules/reference.html#the-module-compiler-option)

From my reading, the official doc describes it very fuzzily. (Please correct me if I am wrong) So I guess the default value is undefined and the actual behavior might change according to different versions of tsc.

weave/guides/integrations/openai_agents.mdx
```diff
@@ -58,3 +59,98 @@ if __name__ == "__main__":
...
+
+- **CommonJS projects:** No additional configuration is required.
+- **ESM projects:** Node must be started with the `--import=weave/instrument` flag so the instrumentation loads before other modules.
+
+The following code sample assumes an ESM project. If you need help setting up your project or determining which type you are using, see [TypeScript SDK: Third-Party Integration Guide](/weave/guides/integrations/js).
+
+In the following code sample, an OpenAI Agent is created and integrated with Weave for traceability. First, update `your-team-name/your-project-name` in the Weave project initialization to real values. A `get_weather` tool is defined that returns a sample weather report. An agent named `Hello world` is configured with basic instructions and access to the weather tool. The main function runs the agent with a sample input (`What's the weather in Tokyo?`) and outputs the final response.
+
+```typescript
+import * as weave from "weave";
```
chance-wnb, line 92: in the CommonJS way, since we are not relying on pre-loader, if we move `import * as weave from "weave";` after `import { Agent, run, tool } from "@openai/agents";` (which means the import of `@openai/agents` runs first), does it still work I wonder?

If it doesn't, and for some reason users are unable to put `import * as weave from "weave";` first. We can suggest them to manually patch in such a situtation. 

If it still works, then we are good!

chance-wnb (commented):
Thank you so much Anastasia. Overall this looks great! 👍  I thought about some minor details. Can you help me verify a few things please?

weave/guides/integrations/openai_agents.mdx
anastasiaguspan: in cjs, we don't use import, so assuming for cjs you meant
const weave = require("weave");
const { Agent, run, tool } = require("@openai/agents");

changed to
const { Agent, run, tool } = require("@openai/agents");
const weave = require("weave");

no, it didn't work.
So i will call this ordering out specifically, good catch.  Tho calling this out might be a little confusing since the import code is different from the example, i'll see what I can do
(just to be sure i did test this for the esm version and it was fine)

weave/guides/integrations/js.mdx
anastasiaguspan: will do

weave/guides/integrations/js.mdx
anastasiaguspan: okay, I'll add more clarification of the setup.

weave/guides/integrations/openai_agents.mdx
chance-wnb: I see. what I said earlier is confusing, let me clarify: regardless of whether it is CJS or ESM, people always write it as:
```
import * as weave from "weave";
import { Agent, run, tool } from "@openai/agents"; 
``` 
as **TypeScript**. This looks identical like ESM, but it is TypeScript :) It just happens that TypeScript uses the modern ESM syntax. 

Eventually, under CJS config, it will be configured to:

```
const weave = require("weave");
const { Agent, run, tool } = require("@openai/agents");
```

weave/guides/integrations/openai_agents.mdx
```diff
@@ -58,3 +59,106 @@ if __name__ == "__main__":
...
+  ]);
+  console.log(result.finalOutput);
+}
+
+main();
+```
+
+The example code on this page uses ESM syntax. In CommonJS projects, the equivalent import statement is:
+```typescript lines
+const weave = require("weave");
```
chance-wnb: so here maybe you also want to mention the typescript syntax as well:

```
import * as weave from "weave";
import { Agent, run, tool } from "@openai/agents"; 
```

and with CommonJS configuration, the order of that two import statements(which will be translated to require() in compiled code) matters.

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,156 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+
+   These settings configure TypeScript to compile for CommonJS:
+
+   - **`module: "CommonJS"`** — Compiles modules to CommonJS format (`require`/`module.exports`).
+     For details on this compiler option, see [TypeScript - Module](https://www.typescriptlang.org/tsconfig/#module).
+
+   - **`target: "es2022"`** *(Recommended)* — Emits modern JavaScript compatible with recent Node.js versions.
+     For details on this compiler option, see [TypeScript - Target](https://www.typescriptlang.org/tsconfig/#target).
+
+   - **`verbatimModuleSyntax: false`** — Allows ESM-style `import`/`export` syntax in source files while emitting CommonJS output.
```
chance-wnb: nit: in my opinion, this doesn't seem to be a commonly-seen option. with its default value being `false`, we probably don't need to mention it. So the simpler the user's config is, the better. 

Since tsconfig is such a huge topic, our doc is not the go-to place for every details, as long as we bring something basic that works, we already fulfill our limited liability :)

Just my two cents, and it is a nit-picky comment. I will leave it to your decision.

chance-wnb (approved):
Thank you so much for the due diligence and attention to every detail! It is much appreciated. Great job!

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
...
 ## Usage instructions
 
-You can either use CommonJS and ESM.
+TypeScript projects can use either the CommonJS or ESM module system.
 
-### CommonJS 
+### CommonJS
 
-For CommonJS, no special configuration is required. Automatic patching works out of the box. Simply install Weave:
+For CommonJS projects, automatic instrumentation works out of the box.
```
dbrian57:
```suggestion
For CommonJS projects, automatic instrumentation works without additional configuration.
```
I think we should try to avoid idioms because they often don't translate well.

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+
+1. Configure your project for CommonJS.
```
dbrian57: I feel like we need some introduction to what we're doing, either here or in the intro sentence. This kinda just jumps into the procedure without telling the user what they're about to do.

My suggestion would be:

```suggestion
To configure your Weave project for CommonJS:

1. Create or update your `package.json` file and set its `type` field to `commonjs`:
```
And then nix the next line

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+
+   Create or update your `package.json`:
+
+   ```json
+   {
+     "type": "commonjs"
+   }
+   ```
+
+   Create a `tsconfig.json` with CommonJS-compatible settings:
```
dbrian57:
```suggestion
   Then create a `tsconfig.json` with CommonJS-compatible settings:
```

Just adding "then" because it sounds kind of awkward without a step number next to it.

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+   }
+   ```
+
+   These settings configure TypeScript to compile for CommonJS:
+
+   - **`module: "CommonJS"`** — Compiles modules to CommonJS format (`require`/`module.exports`).
+     For details on this compiler option, see [TypeScript - Module](https://www.typescriptlang.org/tsconfig/#module).
+
+   - **`target: "es2022"`** *(Recommended)* — Emits modern JavaScript compatible with recent Node.js versions.
+     For details on this compiler option, see [TypeScript - Target](https://www.typescriptlang.org/tsconfig/#target).
```
dbrian57, line 75: I would err on the side of caution and define the other two fields, as well.

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+   ```bash
+   npm install weave
+   ```
+
+3. Compile your TypeScript file.
+
+   For an example file `test.ts`:
+
+   ```bash
+   npx tsc
```
dbrian57, line 95: Is this command correct? It doesn't target a file.

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
 
 ## Usage instructions
 
-You can either use CommonJS and ESM.
+TypeScript projects can use either the CommonJS or ESM module system.
 
-### CommonJS 
+### CommonJS
```
dbrian57: Should this header be more task-based?

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+   node dist/test.js
+   ```
+
+Because CommonJS uses Node.js's `require` module loader, Weave can automatically instrument supported libraries without requiring the `--import` flag used in ESM projects.
+
+### ESM
+
+To use Weave with an ESM TypeScript project, configure your project for Node.js ESM, compile your code with `tsc`, and start Node.js with the `--import` flag so Weave can register its instrumentation before other modules load.
+
+1. Configure your project for ESM.
```
dbrian57: These two lines (87 and 89) sound like separate instruction steps because they both start with imperative verbs, with the first line missing the information I need.

I understand that you're using these intro fragments as summaries for the entire step, but I would recommend against this because it's not clear to the user that this is a summary upon first read. There are a few different ways to tackle this, but IMO, the easiest way is to just make the sub-step (Creating the `tsconfig.json`) a separate numbered step.

If you feel differently about this, let me know! I'm totally open to having my mind changed 😄

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+
+   - **`module: "nodenext"`** — Compiles modules using Node.js ESM semantics.  
+     For details on this compiler option, see [TypeScript - Module](https://www.typescriptlang.org/tsconfig/#module).
+
+   - **`moduleResolution: "nodenext"`** — Ensures module resolution follows Node.js ESM rules.  
+     For details on this compiler option, see [TypeScript - Module Resolution](https://www.typescriptlang.org/tsconfig/#moduleResolution).
+
+   - **`target: "es2022"`** *(Recommended)* — Emits modern JavaScript compatible with recent Node.js versions.  
+     For details on this compiler option, see [TypeScript - Target](https://www.typescriptlang.org/tsconfig/#target).
+
```
dbrian57, line 146: Same feedback here about documenting the other fields.

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+   ```bash
+   npm install weave
+   ```
+
+3. Compile your TypeScript file.
+
+   For an example file `test.ts`:
+
+   ```bash
+   npx tsc
```
dbrian57, line 164: I'm assuming that because I've seen this twice now, it must be correct 😅

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+
+   ```bash
+   node --import=weave/instrument dist/test.js
+   ```
+
+  The `--import` flag ensures the `weave/instrument` module loads before other modules so Weave can automatically instrument supported libraries and integrations.
+
+   Weave must be installed locally in the project you run.
+
+### If you're unsure which type of project you have
```
dbrian57: To me, this is kind of an awkward place to introduce this. Maybe we should put this at the top of the page in a short section that also elaborates on the differences between using Common JS and ESM?

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
...
+If you run a TypeScript file directly with a tool such as:
 
 ```bash
-npm install weave
+npx tsx test.ts
 ```
 
-### ESM 
+the module system may be determined implicitly by your environment.  
+For consistent behavior, it is recommended to define a `package.json` and `tsconfig.json` explicitly.
```
dbrian57:
```suggestion
To ensure consistent behavior, we recommended explicitly defining `package.json` and `tsconfig.json` files.
```
We should own our suggestions.

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
...
 
-For ESM, use Node's `--import` flag to enable auto-instrumentation. The `weave/instrument` module is available as long as the `weave` package is installed.
+To determine whether a project uses CommonJS or ESM, check the `type` field in `package.json`:
+
+```json
+"type": "module"
+```
+
+- If `type` is `"module"`, the project uses **ESM**.
+- If the `type` field is missing or set to `"commonjs"`, the project uses **CommonJS**.
```
dbrian57:
```suggestion
- If the `type` field is missing or set to `"commonjs"`, the project defaults to using **CommonJS**.
```
total nit. up to you

weave/guides/integrations/openai_agents.mdx
```diff
@@ -58,3 +59,99 @@ if __name__ == "__main__":
 ## View traces
 
 When the above code sample is run, a link to the Weave dashboard is generated. To see what happened during your agent execution, follow the link to see your agent traces. 
+</Tab>
+
+<Tab title="TypeScript">
+The [OpenAI Agents Node SDK](https://github.com/openai/openai-node) is a lightweight and powerful framework for building multi-agent workflows. You can use W&B Weave with the OpenAI Agents SDK to trace and monitor your agentic applications in TypeScript.
```
dbrian57: I think we should put this description outside of the tabs, as there is little difference between the TS and Python versions. Also, the Python version incorrectly uses "track" instead of "trace"

weave/guides/integrations/openai_agents.mdx
```diff
@@ -58,3 +59,99 @@ if __name__ == "__main__":
...
+Weave automatically instruments `@openai/agents` when the module is loaded.
+
+Automatic instrumentation works by registering a module loader hook:
+
+- **CommonJS projects:** No additional configuration is required.
+- **ESM projects:** Node must be started with the `--import=weave/instrument` flag so the instrumentation loads before other modules.
+
+If you need help setting up your project or determining which type you are using, see [TypeScript SDK: Third-Party Integration Guide](/weave/guides/integrations/js).
+
+In the following code sample, an OpenAI Agent is created and integrated with Weave for traceability. First, update `your-team-name/your-project-name` in the Weave project initialization to real values. A `get_weather` tool is defined that returns a sample weather report. An agent named `Hello world` is configured with basic instructions and access to the weather tool. The main function runs the agent with a sample input (`What's the weather in Tokyo?`) and outputs the final response.
```
dbrian57:
```suggestion
The following code sample creates an OpenAI Agent and integrates it with Weave for traceability. First, update your-team-name/your-project-name in the Weave project initialization to real values. The sample defines a `get_weather` tool that returns a sample weather report, then configures an agent named `Hello world` with basic instructions and access to the weather tool. The main function runs the agent with a sample input (`What's the weather in Tokyo?`) and outputs the final response.
```

Passive voice

weave/guides/integrations/openai_agents.mdx
```diff
@@ -58,3 +59,99 @@ if __name__ == "__main__":
...
+```
+
+In CommonJS, the `import * as weave from "weave"` statement must appear **before** `import { Agent, run, tool } from "@openai/agents"` for automatic instrumentation. If your code cannot follow this ordering, use the manual instrumentation technique that follows.
+
+
+### Manual instrumentation
+
+Manual instrumentation is only needed in cases where the module loader hook cannot run, such as:
+
+- bundlers that bundle dependencies into a single file
```
dbrian57: I think all of these should start with capitalization? https://developers.google.com/style/lists#introductory-sentences-for-lists

weave/guides/integrations/js.mdx
```diff
@@ -18,28 +18,153 @@ Weave will generally handle this automatically. However, there may be [edge case
 
 ## Usage instructions
 
-You can either use CommonJS and ESM.
+TypeScript projects can use either the CommonJS or ESM module system.
```
dbrian57, line 22: Should we include information here about when you should be using one over the other?

dbrian57 (approved):
Hey, I'm approving because I know this is blocking other things, but I do think this needs a little rework. Please feel free to hit me up for another review if you have time or want to discuss some of the feedback.

I'm mostly concerned about the step summarization stuff and the suggested ESM vs Common section.

weave/guides/integrations/openai_agents.mdx
anastasiaguspan: crap - yeah, that is a classic case of copying from what was there for the Python tab and not noticing it wasn't right from the get-go.   sigh, good catch.

weave/guides/integrations/js.mdx
anastasiaguspan: No, i believe they are equal citizens. prb just depends on other project dependencies etc

weave/guides/integrations/js.mdx
chance-wnb: It should be fine. It targets the CWD, if there is a tsconfig.json, it picks that up.

weave/guides/integrations/openai_agents.mdx
anastasiaguspan: wow, i missed this a second time in a row almost? poor :(

weave/guides/integrations/js.mdx
anastasiaguspan: If the "files" or "include" properties are not specified, it defaults to including all TypeScript files (*.ts, *.tsx) within the directory and its subdirectories.

weave/guides/integrations/openai_agents.mdx
anastasiaguspan: But the library and link are different.

i did fix the word 'track' tho in the python one.

PR #2220, opened 2026-03-03
Runs: Resume run word smithing

ngrayluna:
## Description

Auto enable resume run section used the wrong argument. Also reworded some content to (hopefully) make it more clear.


models/runs/resuming.mdx
```diff
@@ -3,31 +3,24 @@ description: Resume a paused or exited W&B Run
 title: Resume a run
 ---
 
-Specify how a run should behave in the event that run stops or crashes. To resume or enable a run to automatically resume, you will need to specify the unique run ID associated with that run for the `id` parameter:
+Specify how W&B should respond if a run stops or crashes. When you initialize a run you can specify how W&B should respond if the run stops or crashes by passing an argument to the `resume` parameter. W&B checks if the run ID already exists and responds based on the argument passed to the `resume` parameter.
```
mdlinville:
```suggestion
This page shows how to configure how W&B will handle a stopped or crashed run. By default, W&B always initializes a new run. If you an argument to the `resume` parameter when you initialize the run, W&B attempts to resume the run with the exact same run ID instead of initializing a new run with the same run ID.
```
WDYT? As currently written, this seems kind of hard to understand, at least for me.

ngrayluna: How bout this? 

Specify how W&B should respond if a run stops or crashes by setting the `resume` parameter in `wandb.init()`. When you initialize a run, W&B checks whether the run ID already exists and applies the behavior defined by the `resume` value.

mdlinville (approved):
Much clearer!

PR #2224, opened 2026-03-04
Remove the huggingface integration guide link from the huggingface integration guide

TuringTux:
## Description

The huggingface integration guide contains a link to itself. This PR removes that link.


ngrayluna (approved):
Thanks!

PR #2228, opened 2026-03-04
[Weave] Export metrics documentation

dbrian57:
## Description
Resolves DOCS-1926. This adds documentation on how to retrieve metrics about your Weave Calls.

### DIFF
--- a/weave/guides/tracking/querying-calls.mdx
+++ b/weave/guides/tracking/querying-calls.mdx
@@ -109,7 +109,7 @@ For complete details on call properties and fields, see the [Call schema referen
 
 ## Export call metrics
 
-You can also use the Weave Service API's [POST `/calls/stats` endpoint](/weave/reference/service-api/calls/call-stats) to retrieve metrics about your Calls without retrieving the Call data itself. This allows you to quickly and efficiently retrieve information about your Calls, such as latency and cost, and aggregate them by sum, average, minimum, maximum, and count. For example, you can retrieve:
+You can also use the Weave Service API's [POST `/calls/stats` endpoint](https://docs.wandb.ai/weave/reference/service-api/calls/call-stats) to retrieve metrics about your Calls without retrieving the Call data itself. This allows you to quickly and efficiently retrieve information about your Calls, such as latency and cost, and aggregate them by sum, average, minimum, maximum, and count. For example, you can retrieve:
 
 * Total token usage
 * Average latency
@@ -284,4 +284,4 @@ Set the granularity field (in seconds) in the request to change the bucket size.
 }
 ```
 
-You can query metrics for a maximum time range of 31 days. See the [Service API reference](/weave/reference/service-api/calls/call-stats) for more details and available options.
+You can query metrics for a maximum time range of 31 days. See the [Service API reference](https://docs.wandb.ai/weave/reference/service-api/calls/call-stats) for more details and available options.
--- a/weave/guides/tracking/querying-calls.mdx
+++ b/weave/guides/tracking/querying-calls.mdx
@@ -15,8 +15,6 @@ To export Calls:
 ![Traces page showing selected Calls ready for export in the Export modal.](/weave/guides/tracking/imgs/trace_export_modal.png)
 </Frame>
 
-
-
 ## Fetch calls programmatically
 
 <Tabs>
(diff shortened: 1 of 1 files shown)


weave/guides/tracking/querying-calls.mdx
```diff
@@ -109,3 +107,178 @@ To export Calls:
 For complete details on call properties and fields, see the [Call schema reference](/weave/guides/tracking/call-schema-reference).
 
 
+## Export call metrics
+
+You can also use the Weave Service API's [POST `/calls/stats` endpoint](/weave/reference/service-api/calls/call-stats) to retrieve metrics about your Calls without retrieving the Call data itself. This allows you to quickly and efficiently retrieve information about your Calls, such as:
+
+* Total token usage
+* Average latency
+* Total trace storage
```
bcsherma: storage is not supported in the api

weave/guides/tracking/querying-calls.mdx
```diff
@@ -109,3 +107,178 @@ To export Calls:
...
+```
+
+The example response shows two days' worth of metrics. Each day is broken up into its own object in the `usage_buckets` and `call_buckets` arrays. Each array breaks down the metrics differently:
+
+* `usage_buckets`: Groups Call metrics for each day by the model used.
+* `call_buckets`: Groups Call metrics for each day regardless of the model used.
+
+Set the granularity field (in seconds) in the request to change the bucket size.
+
+You can only query metrics for the last 31 days. See the [Service API reference](/weave/reference/service-api/calls/call-stats) for more details and available options.
```
bcsherma: We cap the time domain at 31 days but it doesn't need to be the last 31 days, could be 6 months ago etc

weave/guides/tracking/querying-calls.mdx
bcsherma: I almost interpret this as suggesting that you can only get average latency, but the API supports a variety of aggregations. I think if we clarify that there are a set of metrics you can request (tokens, latency, cost) and then also a set of aggregations that can be applied to any of those (sum, mean, quantiles, etc)

bcsherma (commented):
A few corrections and suggestions, but otherwise good. Thanks!

weave/guides/tracking/querying-calls.mdx
```diff
@@ -109,3 +107,179 @@ To export Calls:
...
+  },
+  body: JSON.stringify(payload),
+});
+
+const data = await response.json();
+console.log(JSON.stringify(data, null, 2));
+```
+</CodeGroup>
+
+The request also specifies how to aggregate the metrics. You can aggregate metrics by `sum`, `count`, `avg`, `min`, `max`, and `count`.
```
anastasiaguspan, line 205: Question, is this specified up in "aggregations": ["sum"]
..it took me a bit to find where this was specified in the code (I was curious) - perhaps adding the line number would help
or, adding this as a comment directly in the code above where it first happens: //You can aggregate metrics by `sum`, `count`, `avg`, `min`, `max`, and `count`.

weave/guides/tracking/querying-calls.mdx
```diff
@@ -109,3 +107,179 @@ To export Calls:
...
+
+* Op name
+* Trace ID
+* Thread ID
+* User ID
+
+The following example demonstrates how to retrieve Calls generated from an Op named `web_app` over a two-day period:
+
+<CodeGroup>
+```python Python
```
anastasiaguspan: want to turn on line numbers for both the code blocks in this file?

weave/guides/tracking/querying-calls.mdx
```diff
@@ -109,3 +107,179 @@ To export Calls:
...
+    }
+  ]
+}
+```
+
+The example response shows two days' worth of metrics. Each day is broken up into its own object in the `usage_buckets` and `call_buckets` arrays. Each array breaks down the metrics differently:
+
+* `usage_buckets`: Groups Call metrics for each day by the model used.
+* `call_buckets`: Groups Call metrics for each day regardless of the model used.
+
```
anastasiaguspan: [278-281] is illustrating 'sum' right? it'll be different if it's not sum?-> maybe should mention that somewhere briefly .  Also, imo i'd actually move these to before the json, it isn't excessive and gives them good introduction as to what the json contains.

anastasiaguspan (approved):
looks good, couple curiosities reported + turn on line numbers.

weave/guides/tracking/querying-calls.mdx
dbrian57: good call! Thank you

PR #2229, opened 2026-03-04
Inference - codegen model lifecycle updates

jamie-rasmussen:
## Description

Moving 5 models from deprecated status to retired. These changes are codegen by the takeru tool.

## Testing
- [x] PR tests succeed


mdlinville (approved):
For the future, it might be good to have a standard thing to put in the Recommended Replacement column when there is no recommendation.

PR #2244, opened 2026-03-06
fix: remove disable automatic app version updates page (WB-28205)

flamarion:
## Summary

Removes the "Disable automatic app version updates" page and all references to it while the underlying bug [WB-28205](https://wandb.atlassian.net/browse/WB-28205) is unresolved.

- Delete `platform/hosting/self-managed/disable-automatic-app-version-updates.mdx` (en, ja, ko)
- Remove nav entries from `docs.json` for all three locales
- Remove the now-broken link from `kubernetes-airgapped.mdx`

## Test plan

- [x] Confirm the page no longer appears in the sidebar
- [x] Confirm no broken links to the removed page

[WB-28205]: https://wandb.atlassian.net/browse/WB-28205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


flamarion:
Do we have an ETA to review and merge this PR?

platform/hosting/self-managed/on-premises-deployments/kubernetes-airgapped.mdx
```diff
@@ -812,8 +812,6 @@ Configure the operator to not automatically update W&B:
 2. Control version updates by manually updating the `spec.chart.version` in your Custom Resource
 3. Optionally, disable automatic updates from the W&B System Console
```
mdlinville: I'm confused about this. If we are removing the link with specific instructions, should we also remove the detail on line 813?

flamarion: 100%!!! I missed that!

mdlinville:
Thanks for this, @flamarion, and sorry it sat so long. Closing it, but the underlying observation was right at the time — here's where things landed.

**The premise changed.** This PR removes the page while [WB-28205](https://coreweave.atlassian.net/browse/WB-28205) is open. That ticket is still `In Review`, but both halves of the bug were fixed after the PR was opened:

- [wandb/console#146](https://github.com/wandb/console/pull/146) (merged 2026-06-24, shipped in [v2.18.16](https://github.com/wandb/console/releases/tag/v2.18.16)) — version pinning now resolves the license from `global.licenseSecret`, so operator installs stop sending an empty license.
- [wandb/core#48158](https://github.com/wandb/core/pull/48158) (merged 2026-07-15) — the deployer assigns new manual deployments to the stable channel instead of returning no releases.

The feature is also in active use. Support pointed a customer at the Console pinning tool on 2026-08-06 to roll back from 0.83 to 0.82.2 after an HTTPS assumption broke their deployment, and another customer pinned to 0.82.2 successfully. Deleting the page would take away the reference the field is pointing customers at, so documenting the caveats is the better trade.

**The branch also can't be merged mechanically.** `wandb/docs` history was rewritten on 2026-05-28 — `main` is now rooted at `0f1b19754` and shares no ancestry with this branch. `git merge-base` returns nothing, and a test merge with `--allow-unrelated-histories` produces 1,895 add/add conflicts across the whole repo. Nothing here is salvageable by resolving conflicts; it would have to be re-applied onto current `main` regardless.

**What replaces it:** coreweave/docs#3760 keeps the page and adds the System Console `v2.18.16` requirement for license-secret installs, a note about release-channel assignment, and a Troubleshooting section keyed to the two error strings users actually hit. It targets `cw-forge-big-merge`, since `wandb/docs` went into soft freeze today and new work lands in `coreweave/docs` for the 2026-09-30 Forge launch.

Your review catch on `kubernetes-airgapped.mdx` — that bullet 3 and the link should go together — was correct and still applies if we ever do remove the page.

[WB-28205]: https://coreweave.atlassian.net/browse/WB-28205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2264, opened 2026-03-10
[DOCS-1898] SCIM API support for creating a team with BYOB storage

mdlinville:
## Description
[DOCS-1898] SCIM API support for creating a team with BYOB storage

## Testing
- [x] PR tests succeed

[DOCS-1898]: https://wandb.atlassian.net/browse/DOCS-1898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


niall-wb (changes requested):
What you have looks good, but I think we should probably provide examples for each of the storage providers that we support. I sent you an example for each in this thread. https://weightsandbiases.slack.com/archives/C069VUZS152/p1770251353361269?thread_ts=1770250332.682799&cid=C069VUZS152 Seems important because there are slight differences for each

mdlinville:
> What you have looks good, but I think we should probably provide examples for each of the storage providers that we support. I sent you an example for each in this thread. https://weightsandbiases.slack.com/archives/C069VUZS152/p1770251353361269?thread_ts=1770250332.682799&cid=C069VUZS152 Seems important because there are slight differences for each

Done, thanks for the reminder! I handled the tabbed output a bit differently for these examples, with a request example per cloud provider but only a single response tab.

theniallriver (approved):

PR #2278, opened 2026-03-12
[DOCS-812] Document how Sweeps responds to UNIX signals

mdlinville:
## Description
[DOCS-812] Document how Sweeps responds to UNIX signals

This aims to clear up persistent and long-running
customer confusion about managing Sweeps agents
and child processes.

## Testing
- [x] Sent lots of signals and observed
- [x] PR tests succeed

[DOCS-812]: https://wandb.atlassian.net/browse/DOCS-812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


models/sweeps/signal-handling-sweep-runs.mdx
```diff
@@ -0,0 +1,103 @@
...
+```python
+import signal
+import sys
+import wandb
+
+
+def signal_handler(signum, frame):
+    if wandb.run is not None:
+        wandb.log({"signal_received": signum})
+    sys.exit(1)
```
kelu-wandb: You left out the `mark_preempting()`.

I'm also not sure we should `wandb.log` it since the signal received could be only temporary and shouldn't be associated directly with the other run properties?  Are people doing this?

(And this next part isn't that important, but [there is a standard for returning exit code as 128 + signum](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_08_02))

I would recommend this:

```python
def signal_handler(signum, frame):
    if wandb.run is not None:
        # Optional model checkpointing, etc.
        print(f"Preempted with signal: {signal.Signals(signum).name}.")
        wandb.mark_preempting()
    sys.exit(128 + signum)
```

models/runs/resuming.mdx
```diff
@@ -113,10 +113,12 @@ Use the [`mark_preempting`](/models/ref/python/experiments/run#mark_preempting)
...
-The following table outlines how W&B handles runs based on the exit status of a sweep run.
+The following table outlines how W&B handles runs based on the exit status of the sweep run.
 
 |Status| Behavior |
 |------| ---------|
 |Status code 0| Run is considered to have terminated successfully and it will not be requeued.  |
 |Nonzero status| W&B automatically appends the run to a run queue associated with the sweep.|
-|No status| Run is added to the sweep run queue. Sweep agents consume runs off the run queue until the queue is empty. Once the queue is empty, the sweep queue resumes generating new runs based on the sweep search algorithm.|
\ No newline at end of file
+|No status| Run is added to the sweep run queue. Sweep agents consume runs off the run queue until the queue is empty. Once the queue is empty, the sweep queue resumes generating new runs based on the sweep search algorithm.|
```
kelu-wandb: I don't think we should make the table based on exit code status, and instead provide scenarios.

For a simplified version, we can do this:

(1) Advise them to register a signal handler that calls `mark_preempting()` then exits with a non-zero exit code, and NOT the always call `mark_preempting()` example above.

(2) Then the below is what happens if you do that:

| Scenario | Run State |
| - | - |
| Run completes normally and returns exit code 0 | FINISHED |
| Run fails and returns non-zero exit code | FAILED |
| Run gets unhandled signal, e.g. `SIGKILL` | CRASHED after 5 min |
| Run gets handled signal, e.g. `SIGTERM` or `SIGUSR1` |  PREEMPTED, queued to be picked up by next agent request |

Also mention that this is all only supported for use with the `wandb agent` CLI command.

(We probably still need to do some work to make it work well with the Python API path.)

models/runs/resuming.mdx
```diff
@@ -113,10 +113,12 @@ Use the [`mark_preempting`](/models/ref/python/experiments/run#mark_preempting)
 with wandb.init() as run:
     run.mark_preempting()
```
kelu-wandb: Please see my note below. I think we should not present this always run `mark_preempting()` example upfront and give the false impression that it's preferred.

We should either just link to the separate page, or put the best practice code here instead.

models/sweeps/signal-handling-sweep-runs.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+
+def train():
+    signal.signal(signal.SIGINT, signal_handler)
```
kelu-wandb: I recommend using `SIGUSR1` in this example instead of `SIGINT`, as I think we're unlikely to rely on SIGINT for preemption, and it's better to leave it open for manual cancellation.

models/sweeps/signal-handling-sweep-runs.mdx
```diff
@@ -0,0 +1,103 @@
...
+`SIGKILL` cannot be caught or ignored. The process terminates immediately with no chance to run handlers or atexit callbacks. W&B cannot write a final summary for the run. The agent still recovers and continues the sweep, but run data for that run is incomplete. Use `SIGKILL` only as a last resort; prefer `SIGTERM` or `SIGINT` when you need graceful shutdown.
+
+## Forwarding signals from agent to child
+
+The sweep agent runs your training script as a child process. When you interrupt the **agent** (for example, with Ctrl+C or when a scheduler sends `SIGTERM` to the job), the **child** (training process) does not receive the signal by default; the training script cannot run its handler or call `mark_preempting()`. This is described in [GitHub #3667](https://github.com/wandb/wandb/issues/3667).
+
+To let the child shut down gracefully and optionally mark the run as preempting, forward signals from the agent to the child:
+
+- **CLI**: Run the agent with the `--forward-signals` flag: `wandb agent --forward-signals entity/project/sweep_ID`
+- **Python**: Pass `forward_signals=True` to [`wandb.agent()`](/models/ref/python/functions/agent)
```
kelu-wandb: `forward_signals` actually doesn't work in case of python API. It only works with CLI.

models/runs/resuming.mdx
kelu-wandb: Meanwhile, here's the complete picture. I'm not sure how exactly this should feed into the details doc below.

| Scenario | No `mark_preempting()` | Signal handler calls `mark_preempting()` and returns non-zero exit code | `mark_preempting()` always called after `init()` |
| - | - | - | - |
| Run completes normally with exit code 0 | FINISHED | FINISHED | FINISHED |
| Run fails with exit code non-zero | FAILED | FAILED | PREEMPTED |
| Run gets `SIGKILL` | CRASHED after 5 min |  CRASHED after 5 min (uncatchable) | PREEMPTED after 5 min |
| Run gets `SIGINT` | KILLED | PREEMPTED (with `SIGINT` handler) | PREEMPTED |
| Run gets other signal, e.g. `SIGTERM` or `SIGUSR1` |  CRASHED after 5 minutes | PREEMPTED (with signal handler) | PREEMPTED after 5 min |


If you put `mark_preempting()` in your signal handler, it means means you won't catch cases where the signal handler never gets called, such as if you get a `SIGKILL`.

However, if you always call it after `init()`, that means ANY failure will count as a preemption and queue up the run for re-execution, which might lead to repeated failures.

So generally, if you're on a system with a well-defined preemption signal, the correct implementation would be to implement a signal handler.

Other notes:

- `wandb sweep --cancel` or the "Cancel" button on the Sweeps UI Controls page triggers a `SIGKILL` on runs.
- By default, the agent can be killed independently of the run process, unless you enable the `--forward-signals` flag when running the agent.
- If you hit Ctrl-C twice on the agent process, it sends `SIGTERM` by default, unless `--forward-signals` is true, in which case the `SIGINT` is forwarded.

As for the API version:

To be honest I don't think all of this would work that well for the API version right now, since that runs the child job in a thread instead of a process. I think the correct way to handle that case would be for you to be able to ask the wandb API itself to register a signal handler, which would then terminate the child process then call your callback (which would presumably call `wandb.mark_preempting()` and `wandb.finish()`).  I'll probably go file a feature request for that. :P

kelu-wandb (changes requested):
Sorry this is so confusing!

Please let me know if you have more questions.

models/runs/resuming.mdx
```diff
@@ -107,16 +107,19 @@ If you can not share a filesystem, specify the `WANDB_RUN_ID` environment variab
 ## Resume preemptible Sweeps runs
 Automatically requeue interrupted [sweep](/models/sweeps/) runs. This is particularly useful if you run a sweep agent in a compute environment that is subject to preemption such as a SLURM job in a preemptible queue, an EC2 spot instance, or a Google Cloud preemptible VM.
```
kelu-wandb: I know this was already here, but this is an awkward sentence that's more of a subtitle than a sentence. Should probably rephrase it?

models/sweeps/signal-handling-sweep-runs.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+This applies whether the process exits from a signal handler, from an exception, or from an explicit `sys.exit()` call. Understanding and relying on this contract is vitally important in preemptible or cluster environments.
+
+When the process exits due to a [**catchable** signal](#catchable-signals-and-preemption), your handler can run, call [`wandb.run.mark_preempting()`](/models/ref/python/experiments/run#mark_preempting) if you want the run requeued, perform cleanup (for example, save a checkpoint), then exit with a non-zero code. A common convention is `sys.exit(128 + signum)` for termination by signal. W&B records that exit code and the same [requeue rules](/models/runs/resuming#resume-preemptible-sweeps-runs) apply. When the process is killed by the operating system kernel with [**`SIGKILL`**](#sigkill-uncatchable), the process cannot run exit hooks, so no final summary is written and the run may appear as crashed or killed; the agent still starts the next run.
+
```
kelu-wandb, line 20: I would maybe add a section here about stale runs, mentioning that if a run doesn't finish or post any metrics for a long time (approx 5 minutes), the server will mark it as crashed.

models/sweeps/signal-handling-sweep-runs.mdx
```diff
@@ -0,0 +1,128 @@
...
+## Catchable signals and preemption
+
+You can register custom signal handlers in your training script. When a catchable signal is delivered, your handler runs; metrics already sent to W&B are preserved, and the agent detects the process exit and starts the next run.
+
+**Best practices:**
+
+- Register handlers early (for example, before entering the main training loop).
+- In the handler, call [`wandb.run.mark_preempting()`](/models/ref/python/experiments/run#mark_preempting) when you intend the run to be requeued after preemption, perform cleanup (for example, save a checkpoint), then exit with a non-zero code. Avoid logging ad hoc metrics for the signal alone; a signal can be transient, and a dedicated metric can be misleading next to normal training metrics.
```
kelu-wandb: I don't think you need to explicitly call out the bit about avoiding logging ad hoc metrics about the signal?  Up to you. That was just my comment about the sample code. :)

kelu-wandb (approved):
Approving with minor comments

PR #2288, opened 2026-03-16
[Weave] Reference dataset when using eval logger

dbrian57:
## Description
Resolves DOCS-1234. Adds documentation demonstrating how to reference an already published dataset to keep users from resubmitting their data every time.

### DIFF
--- a/weave/guides/evaluation/evaluation_logger.mdx
+++ b/weave/guides/evaluation/evaluation_logger.mdx
@@ -321,7 +321,7 @@ While TypeScript doesn't have automatic cleanup with context managers, `logSumma
 
 ### Link to an existing dataset
 
-When you pass raw dicts as `inputs` to `log_prediction`, Weave re-ingests the data with every evaluation run. For large datasets, this is inefficient because the same data is stored multiple times.
+When you pass raw datasets as `inputs` to `log_prediction`, Weave re-ingests the data with every evaluation run. This stores duplicate data, which may waste space if the dataset is large or if a large number of evaluations reuse it.
 
 To avoid this duplication, publish your dataset to Weave before running any evaluations, then pass the published dataset's rows as `inputs`. Weave resolves references to published rows using internal references instead of re-ingesting the data. This technique gives you the same linked experience as the standard [Evaluation framework](../core-types/evaluations), where each prediction links back to a specific dataset row in the Weave UI.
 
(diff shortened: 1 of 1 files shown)


weave/guides/evaluation/evaluation_logger.mdx
```diff
@@ -319,6 +319,61 @@ While TypeScript doesn't have automatic cleanup with context managers, `logSumma
 
 
 
+### Link to an existing dataset
+
+When you pass raw dicts as `inputs` to `log_prediction`, Weave re-ingests the data with every evaluation run. For large datasets, this is inefficient because the same data is stored multiple times.
```
mdlinville: Is "raw dict" really what they are called? What about something like "This stores duplicate data, which may waste space if the dataset is large or if a large number of evaluations reuse it" or something?

weave/guides/evaluation/evaluation_logger.mdx
```diff
@@ -319,6 +319,61 @@ While TypeScript doesn't have automatic cleanup with context managers, `logSumma
...
+
+To avoid this, publish your dataset to Weave first, then pass the published dataset's rows as `inputs`. Published rows carry internal references that Weave resolves instead of re-ingesting. This gives you the same linked experience as the standard [Evaluation framework](../core-types/evaluations), where each prediction links back to a specific dataset row in the Weave UI.
```
mdlinville:
```suggestion
To avoid this duplication, publish your dataset to Weave before running any evaluations, then pass the published dataset's rows as `inputs`. Weave resolves references to published rows using internal references instead of re-ingesting the data. This technique gives you the same linked experience as the standard [Evaluation framework](../core-types/evaluations), where each prediction links back to a specific dataset row in the Weave UI.
```
Something like this? Ignore if I'm getting it wrong.

weave/guides/evaluation/evaluation_logger.mdx
```diff
@@ -319,6 +319,61 @@ While TypeScript doesn't have automatic cleanup with context managers, `logSumma
...
+
+The following example shows how to publish a dataset and link to it in the `EvaluationLogger`. Once you have retrieved the dataset, you can access and iterate over it like any other dataset.
```
mdlinville:
```suggestion
The following example publishes a dataset and links to it in the `EvaluationLogger`, before retrieving and iterating over it like any other dataset.
```

mdlinville (approved):

PR #2305, opened 2026-03-17
Prioritize style guide over matching existing content in AGENTS.md

anastasiaguspan:
addl notes on changes in ticket https://wandb.atlassian.net/browse/DOCS-1986
## Summary

- Consolidates seven "match existing content" / "surrounding content" mentions into one clear directive: follow the style guide for all new and edited content, fix style issues within the section you're editing, and save broad refactors for dedicated PRs.
- Reframes the style hierarchy so that this style guide and the Google Developer Style Guide are the primary authorities, rather than "match existing content first."
- Clarifies list capitalization: "Start each list item with a capital letter unless the first word is a case-sensitive term like a code identifier."  (I added this because it got it wrong yesterday on mine)
- Adds `weave.init` pattern guidance.

## Why

The repeated "match existing content even if it's not perfect" messaging risked agents replicating style errors from older content instead of following the style guide. This change makes the style guide the source of truth while still discouraging drive-by style refactors in content PRs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -20,7 +20,7 @@ As a W&B docs team member, you:
 
 ## Core principles
 
-The content we produce in this repo must be accurate and clear, must adhere to professional tech writing and coding standards, and must be consistent with its surrounding existing content. Avoid "hallucinating" features, functionality, or code that doesn't exist.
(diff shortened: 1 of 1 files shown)


mdlinville:
Closing here but please resurrect in `coreweave/docs` if you'd like.

PR #2321, opened 2026-03-19
[Weave] Adds self-managed URL to reference docs

dbrian57:
## Description
Resolves DOCS-1072. Updates the Weave Service API reference overview page with the base URL for self-managed instances.

### DIFF
--- a/weave/reference/service-api.mdx
+++ b/weave/reference/service-api.mdx
@@ -5,13 +5,13 @@ description: "Use the Weave Service API, which provides REST endpoints for progr
 
 ## Base URL
 
-For [W&B Multi-tenant Cloud](/platform/hosting/hosting-options/multi_tenant_cloud) and [Dedicated Cloud](/platform/hosting/hosting-options/dedicated-cloud), use the following base URL:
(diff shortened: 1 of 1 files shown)


weave/reference/service-api.mdx
```diff
@@ -5,10 +5,18 @@ description: "Use the Weave Service API, which provides REST endpoints for progr
 
 ## Base URL
 
+For [W&B Multi-tenant Cloud](/platform/hosting/hosting-options/multi_tenant_cloud) and [Dedicated Cloud](/platform/hosting/hosting-options/dedicated-cloud), use the following base URL:
```
anastasiaguspan: Since Dedicated Cloud and Self-Managed (line 14 below) are products, I think we're supposed to have W&B precede them on first use per page: - **First mention pattern**: Use "W&B [Product]" on first mention in a page, then drop "W&B" for subsequent mentions.

it seems like the platform pages treat them this way.

anastasiaguspan (approved):
very small indeed, you weren't exaggerating lol

weave/reference/service-api.mdx
dbrian57: ah, yeah, good call.

PR #2325, opened 2026-03-20
feat(scripts): add Mintlify assistant conversations export

johndmulhausen:
## Summary

Adds `scripts/download_mintlify_assistant_conversations.py` to download AI assistant conversation history from the [Mintlify Analytics API](https://www.mintlify.com/docs/api/analytics/assistant-conversations) (`GET /v1/analytics/{projectId}/assistant`), following pagination until `hasMore` is false. Output is JSON with `count` and `conversations`.

## Credentials

- Read only from repo-root `.env` and `.env.local` (`.env.local` overrides duplicate keys).
- Required: `MINTLIFY_PROJECT_ID`, `MINTLIFY_API_KEY` (organization admin key, usually `mint_`).

## Implementation

- Standard library only (`urllib`).
- Cloudflare: browser-like `User-Agent`; on HTTP 403, retries via `curl` when available.
- Strips a duplicated `Bearer ` prefix, scrubs invisible Unicode from secrets, URL-encodes `projectId` in the path.
- `--verbose` for debugging 401s; extra stderr hints when unauthorized.

## Other

- `.gitignore`: ignore `.env` so API keys are not committed.

## How to test

```bash
python scripts/download_mintlify_assistant_conversations.py -o /tmp/conversations.json
python scripts/download_mintlify_assistant_conversations.py --verbose
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ lychee/
 # Python
 __pycache__/
 *.pyc
+.env
--- a/scripts/download_mintlify_assistant_conversations.py
+++ b/scripts/download_mintlify_assistant_conversations.py
@@ -1,0 +1,343 @@
+#!/usr/bin/env python3
+"""
+Download AI assistant conversation history from Mintlify Analytics.
+
+Calls GET https://api.mintlify.com/v1/analytics/{projectId}/assistant and
+paginates until all conversations are collected. See:
+https://www.mintlify.com/docs/api/analytics/assistant-conversations
+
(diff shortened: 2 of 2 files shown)


ngrayluna:
I'm not sure exactly what this PR does. I created a simple shell script that got responses. 🤷 

Script to get responses: https://github.com/ngrayluna/explore_mintlify_agent/blob/main/get_responses.sh
Then I convert the JSON into CSV: https://github.com/ngrayluna/explore_mintlify_agent/blob/main/json_to_csv.sh

PR #2367, opened 2026-03-26
[Weave] signals content-DOCS-1994

anastasiaguspan:
## Description

Adding new Signals content (naming still TBD).
Signals is replacing the goto starter user journey, making existing 'monitors' the more advanced, custom path - as such it is moved to a subsequent page 'custom monitors'.

## Testing
- [ ]x Local build succeeds without errors (`mint dev`)
- [ ] PR tests succeed

## Related issues

- Fixes DOCS-1994


NiWaRe (commented):
I like the phrasing as "Monitor using built-in signals" - we're still debating on the exact name but this might be a good middle ground.

Three points that we could add: 

- The most natural way of "using" signals is through the dashboard under "Project" tab (still behind feature flag but can add in screenshots) and the added tags in the "Signals" column in the trace table (if you hover over a tag it shows you the reason and confidence score, for example [here](https://wandb.ai/wandb-smle/email-agent-online-rl/weave/traces?view=traces_2026-03-12_17-11-30-841&peekPath=%2Fwandb-smle%2Femail-agent-online-rl%2Fcalls%2F019d4674-1843-715d-b928-246193518a86%3FhideTraceTree%3D0))
- We might want to include the specific output pattern that will make Weave render the scorer output as a tag in case people want to create their own custom signals? 
- Could we add a disclaimer with potentially a form or email address to tell people that this is currently in private preview and be tested by reaching out to us?

weave/guides/evaluation/custom-monitors.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+You can monitor text, images, and audio in your application's input and output.
+
+Monitors require no code changes to your application. Set them up using the W&B Weave UI.
+
+If you need to actively intervene in your application's behavior based on scores, use [guardrails](/weave/guides/evaluation/guardrails) instead.
+
+### When to use signals and custom monitors
+
+Use [signals](/weave/guides/evaluation/monitors) to get started with production monitoring quickly, then add custom monitors for evaluation criteria specific to your application.
```
dbrian57: I think we need to just briefly explain what signals and custom monitors here are first before we move into when you should use each.

weave/guides/evaluation/monitors.mdx
```diff
@@ -1,164 +1,114 @@
...
-
-7. After running the script using several different statements, open the W&B UI and navigate to the **Traces** tab. Select any **LLMAsAJudgeScorer.score** trace to see the results.
-
-![Monitor trace](/weave/guides/evaluation/img/monitors-4.png)
+In modern agent development, standard system metrics like latency, token count, and cost are insufficient for understanding complex agent behavior. While inspecting individual traces provides deep insight, it is impossible to scale manual reviews across the millions of traces generated in a live environment.
+
+Signals provide a high-level monitoring solution designed to bridge this gap by offering automated, behavioral scoring for agents in production.
+
+Signals utilize a robust backend infrastructure to provide real-time performance insights:
+- Automated scoring: Every incoming production trace is automatically processed and scored based on predefined metrics.
```
dbrian57: Should these opening terms be bolded?

weave/guides/evaluation/monitors.mdx
```diff
@@ -1,164 +1,114 @@
...
-</Tabs>
-
-7. After running the script using several different statements, open the W&B UI and navigate to the **Traces** tab. Select any **LLMAsAJudgeScorer.score** trace to see the results.
-
-![Monitor trace](/weave/guides/evaluation/img/monitors-4.png)
+In modern agent development, standard system metrics like latency, token count, and cost are insufficient for understanding complex agent behavior. While inspecting individual traces provides deep insight, it is impossible to scale manual reviews across the millions of traces generated in a live environment.
+
+Signals provide a high-level monitoring solution designed to bridge this gap by offering automated, behavioral scoring for agents in production.
+
+Signals utilize a robust backend infrastructure to provide real-time performance insights:
```
dbrian57: I'm skeptical about mentioning a backend infrastructure if customers don't have to manage/configure a backend infrastructure. 

Also, I'm so sorry, but my old manager is yelling at me in my head about never using the word "utilize" when you could write "use" 😅

weave/guides/evaluation/monitors.mdx
```diff
@@ -1,164 +1,114 @@
...
+- Automated scoring: Every incoming production trace is automatically processed and scored based on predefined metrics.
+- Infrastructure: Processing is powered by Coreweave compute and Coreweave GPUs to ensure scalability across millions of traces.
+- Custom metrics: Developers can create specific metrics, such as response length or faithfulness to source material, to help understand exactly how an agent is behaving.
```
dbrian57: I think we might need to rewrite the opening line a little bit, because it makes it sound like this is going to be a list of performance metrics that it can return, but these are really features of the feature.

weave/guides/evaluation/monitors.mdx
```diff
@@ -1,164 +1,114 @@
...
+
+By using signals within production, you can:
+- Gain behavioral insight: Move beyond simple system metrics to understand if your agent is hallucinating, failing to follow conversation patterns, or losing grounding in its evidence.
+- Automate alerts: Set up automated triggers that notify your team through tools like Slack when an agent's performance drops below a certain threshold.
+- Accelerate the research loop: Use the scores and failure analyses generated by signals to identify specific weaknesses, which can then be used to kick off the research loop for offline model improvement, data annotation, or reinforcement learning.
```
dbrian57:
```suggestion
- Accelerate the research loop: Use the scores and failure analyses generated by signals to identify specific weaknesses, which you can use to research model improvement, data annotation, or reinforcement learning.
```
"kick off" is kind of an idiom, so I try to avoid it.

weave/guides/evaluation/monitors.mdx
```diff
@@ -1,164 +1,114 @@
...
-await generateStatement("Water freezes at 0 degrees Celsius.");
-await generateStatement("The Great Wall of China was built over several centuries.");
-```
-  </Tab>
-</Tabs>
-
-7. After running the script using several different statements, open the W&B UI and navigate to the **Traces** tab. Select any **LLMAsAJudgeScorer.score** trace to see the results.
-
-![Monitor trace](/weave/guides/evaluation/img/monitors-4.png)
+In modern agent development, standard system metrics like latency, token count, and cost are insufficient for understanding complex agent behavior. While inspecting individual traces provides deep insight, it is impossible to scale manual reviews across the millions of traces generated in a live environment.
```
dbrian57: I think we should plainly state what signals are at the start. Otherwise, this kinda reads like a company blog post. We could amend this quickly by moving the second paragraph to the top and fixing it up a little at the end. We should also make the connection that signals are a type of monitor.

Also, I _feel_ like this paragraph is very opinionated. I think we should consider nixing it entirely and in lieu of something that plainly states what Signals do and how they benefit the customer.

weave/guides/evaluation/monitors.mdx
```diff
@@ -1,164 +1,114 @@
...
+
+By using signals within production, you can:
+- Gain behavioral insight: Move beyond simple system metrics to understand if your agent is hallucinating, failing to follow conversation patterns, or losing grounding in its evidence.
+- Automate alerts: Set up automated triggers that notify your team through tools like Slack when an agent's performance drops below a certain threshold.
+- Accelerate the research loop: Use the scores and failure analyses generated by signals to identify specific weaknesses, which can then be used to kick off the research loop for offline model improvement, data annotation, or reinforcement learning.
+
+
+## Available signals
+
+W&B Weave offers monitors with built-in signals: preset scorers that evaluate production traces for common quality issues and errors out of the box, with no custom setup. Each built-in signal uses a benchmarked LLM prompt to classify traces and saves the results as comma-delimited tags representing the detected issues.
```
dbrian57, line 18: I feel like this is more introductory stuff that should be at the top of the page.

weave/guides/evaluation/monitors.mdx
```diff
@@ -1,164 +1,114 @@
...
+
+To start classifying traces immediately, enable signals from the Monitors page. Signals don't require prompt engineering or scorer configuration.
```
dbrian57: The workflows below guide people through this, so I don't think we need this line.

weave/guides/evaluation/monitors.mdx
```diff
@@ -1,164 +1,114 @@
...
+
+Signals use a [W&B Inference](/inference/) model to score traces, so no external API keys are required.
```
dbrian57, line 20: Probably should either be in the intro area or the How signals work section. I'd opt for the introductory section and also add whether or not this costs customers additional money to use.

weave/guides/evaluation/monitors.mdx
```diff
@@ -1,164 +1,114 @@
...
+
+### Alert on signals
+To get notified when a signal is triggered, set up an [automation](/weave/guides/evaluation/automations).
+
+
+## How signals work
+
+Each signal uses an LLM-as-a-judge approach to classify traces:
+
+1. **Trace selection**: Quality signals evaluate successful root-level traces. Error signals evaluate failed traces. Child spans and intermediate calls are not scored.
```
dbrian57: I feel like I'm missing context here. Are each of these a step that Weave goes through to assess and apply a signal to a trace?

Also, the "Trace selection" bullet introduces new information: quality signals only evaluate root-level traces. I feel like this should be mentioned earlier in "Available signals" so users don't wonder why child spans aren't scored.

dbrian57 (changes requested):
Hey, so I think this doc needs some rework. The opening section is very markety and contains three consecutive and sizable blocks of info before anything actionable is mentioned.

I would opt to:
1) Rework the intro section to more plainly and concisely state what signals are and their benefits. I would also make sure to explicitly mention that they're a type of monitor.
2) Give the same treatment to the intro of the "Available signals" section
3) Move some wording into the intro section (I noted these in my annotations)
4) Clarify the "How signals work" section. This might just need an intro adjustment, but I'm not sure. I just found it difficult to follow.

I'm happy to discuss any of this, obviously.

weave/guides/evaluation/monitors.mdx
anastasiaguspan: funny you mention that, because no, i don't think they actually "should" (i don't think they technically qualify as 'lead-in')- but yes, everything AI ***loves*** to bold these, so now it looks weird if we don't?  style guide grey area. there was a slack chat about this a while ago. i'm opting for no here, because they aren't that juicy/useful of terms for the user.

weave/guides/evaluation/monitors.mdx
anastasiaguspan: hmm, i kinda disagree? this is starting with the problem statement. the problem that signals are solving. I think that explains the benefit to the customer more than trying to explain the benefit first.
I tightened up the sentences in the intro however, so it is still problem statement, then how signals answers that problem. take a look and see if you dislike it less ;)

weave/guides/evaluation/monitors.mdx
anastasiaguspan: I updated the automated scoring bullet to includ the important most-basic bit, but other than that, it is the first section.i tightened up the intro so this is now closer to teh top as well

weave/guides/evaluation/monitors.mdx
anastasiaguspan: moved section up

anastasiaguspan: and ur right, shouldn't be numbered.

dbrian57 (approved):
Looks good!

anastasiaguspan:
Superseded — the signals docs and custom-monitors page have been brought forward into the `weave_agents` release branch (via PR #2647). The content from this PR was cherry-picked as a single squashed commit on `aguspan_agents2`, with the "W&B Inference" → "Serverless Inference" rebrand applied to match the rest of `weave_agents`.

Closing this PR since it would otherwise stay perpetually conflicting against `main`.

PR #2386, opened 2026-03-30
[Weave] Update prompt versioning docs to recommend aliases

rgao-coreweave:
## Summary
- Update prompt versioning page to recommend aliases (e.g. `production`) over version indices for production use
- Clarify ref URI docs to mention aliases and version indices
- Fix step numbering in the "View and compare" section
- Update production workflow to use `set_aliases` instead of manual version pinning

## Test plan
- [x] Verify page renders correctly
- [x] Confirm all internal links resolve

### DIFF
--- a/weave/guides/core-types/prompts-version.mdx
+++ b/weave/guides/core-types/prompts-version.mdx
@@ -68,7 +68,7 @@ prompt = weave.ref("support_prompt:v3").get()
 
 When deploying prompts in production systems, use an [alias](#add-prompt-tags-and-aliases) like `production` rather than a version index or latest version. In your production application, load the prompt with `weave.ref(<name>:<alias>).get()`.
 
-Using an alias ensures that production behavior remains stable and predictable. When you're ready to promote a new version, move the alias and all consumers automatically pick up the change.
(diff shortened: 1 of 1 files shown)


weave/guides/core-types/prompts-version.mdx
```diff
@@ -66,17 +64,17 @@ weave.init("your-team-name/your-project-name")
...
 ```
 
-## Use prompt versions in production
+## Use prompts in production
 
-When deploying prompts in production systems, pin a specific prompt version rather than referencing the latest version. In your production application, load that version with `weave.ref(<name>:<version>).get()`.
+When deploying prompts in production systems, use an [alias](#add-prompt-tags-and-aliases) like `production` rather than a version index or latest version. In your production application, load the prompt with `weave.ref(<name>:<alias>).get()`.
 
-Pinning a version ensures that production behavior remains stable even if new prompt versions are published later.
+Using an alias ensures that production behavior remains stable and predictable. When you're ready to promote a new version, move the alias and all consumers automatically pick up the change.
```
anastasiaguspan:
```suggestion
Using an alias ensures that production behavior remains stable and predictable. When you're ready to promote a new version, move the alias to that version and then all consumers automatically pick up the change.
```

anastasiaguspan (approved):

PR #2403, opened 2026-03-31
Add docs-skills as submodule at .claude

dfinster:
As titled.


johndmulhausen (approved):
I could hardly be more excited about this!

PR #2408, opened 2026-04-01
Inference - add generated model price table

jamie-rasmussen:
## Description

Currently the product docs just defer to the marketing site for pricing information. This was originally done to reduce manual effort and the likelihood of unsynchronized information.

Now, with the `takeru` tooling, keeping information in the product docs synchronized with the model catalog is significantly simplified. It seems better to keep users in the docs then send them elsewhere.

Note that this change reflects an upcoming pricing adjustment for `moonshotai/Kimi-K2.5` and merging this PR should wait on synchronization with changes elsewhere.

Before:

image

After:

image

## Testing
- [x] PR tests succeed

### DIFF
--- a/inference/usage-limits.mdx
+++ b/inference/usage-limits.mdx
@@ -9,7 +9,29 @@ Learn about pricing, limits, and other important usage information before using 
 
 ## Pricing
 
-For detailed model pricing information, visit [W&B Inference pricing](https://wandb.ai/site/pricing/inference).
+Model pricing is subject to change. Prices shown are per 1 million tokens.
+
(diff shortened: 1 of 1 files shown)


mdlinville (changes requested):
In general we avoid specific pricing details in the docs in favor of the pages owned by product and the billing team. This helps us to avoid stale details that could cause misunderstandings and also acknowledges that contract customers may have bespoke terms. I wonder if we can consider an approach here that accommodates those concerns.

jamie-rasmussen:
Closing as per @mdlinville comment above

PR #2410, opened 2026-04-01
Ignore .cursor/ in git

anastasiaguspan:
Adds `.cursor/` to `.gitignore` so local Cursor rules and editor config stay untracked and out of PR diffs.

Each developer can keep personal rules (for example Jira MCP routing) under `.cursor/rules` without committing them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,7 @@ go.sum
 
 # Misc
 # Cursor editor local rules and config (per-developer)
-.cursor/
+.cursor
 
 .DS_Store
 .env.local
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -16,6 +16,9 @@ go.mod
 go.sum
 
 # Misc
+# Cursor editor local rules and config (per-developer)
+.cursor/
```
mdlinville:
```suggestion
.cursor
```
nit

mdlinville (approved):

PR #2426, opened 2026-04-07
W&B Sandboxes

ngrayluna:
## Description

W&B Sandboxes (Private Preview). Currently documents:

- Overview (what it is, high-level how it works)
- Sandbox lifecycle
- Create sandboxes
- Run commands
- File access (read/write/mount)
- Secrets
- Tutorial: Train PyTorch model in a sandbox
- Tutorial: Invoke an agent within a sandbox

Things to add?
- Compute resourcing info
- Authenticating gotchas? (If they exist)


trane293:
## Review

Overall the docs are well-structured and cover the right topics. The API patterns are mostly accurate against the SDK. Flagging the following issues before merge.

---

### Must Fix

**1. `wait_until_complete()` missing `.result()` call**
Files: `sandboxes/lifecycle.mdx` (line 63), `sandboxes/run-commands.mdx` (line 23)

Both show:
```python
sandbox.wait_until_complete(timeout=3600.0)
print(f"Exit code: {sandbox.returncode}")
```

`wait_until_complete()` returns an `OperationRef[Sandbox]`, not `Sandbox`. Without `.result()`, the call doesn't block and `sandbox.returncode` will be `None`. The SDK's own docstring shows the correct pattern:

```python
sandbox.wait_until_complete(timeout=3600.0).result()
print(f"Exit code: {sandbox.returncode}")
```

---

**2. Credential propagation claim is inaccurate**
`sandboxes.mdx` (line 12)

> "your W&B credentials work automatically, and sandbox sessions can report metrics to an active W&B run"

This is not true today. W&B credentials authenticate you *to* the sandbox service (creating/managing sandboxes via ATC), but they are not propagated *inside* the sandbox container. Code running inside the sandbox has no access to your W&B API key, so `wandb.init()`, `wandb.log()`, and Weave calls won't work without explicitly passing `WANDB_API_KEY` via `environment_variables` or the Secrets API.

Suggested rewrite: "Sandboxes authenticate using your existing W&B credentials. To use W&B or Weave inside a sandbox, pass your API key using the [Secrets Manager](/sandboxes/secrets) or environment variables."

---

**3. `echo > file` doesn't work without a shell**
`sandboxes/file-access.mdx` (line 53)

```python
sandbox.exec(["echo", "Hello, world.", ">", "hello.txt"]).result()
```

When args are passed as a list, `exec()` doesn't invoke a shell. The `>` is passed as a literal argument to `echo`, so no file is created and the subsequent `read_file("hello.txt")` would fail. Should be:

```python
sandbox.exec(["sh", "-c", "echo 'Hello, world.' > hello.txt"]).result()
```

Also, `Path` is used on line 57 (`Path("hello.txt").write_bytes(content)`) but is never imported in this example.

---

**4. LangChain agent tutorial should be replaced**
`sandboxes/invoke_agent_sandbox_tutorial.mdx`

LangChain is a direct competitor. We should not be featuring their library in our docs. Beyond the competitive concern, the code has technical issues:
- `ToolRuntime[Context]` (line 94): `ToolRuntime` is a dataclass, not a generic type
- No version pins: `create_agent` was removed in LangChain v1.1.0, so `pip install langchain langgraph langchain_anthropic` without version constraints will break
- `@dataclass` used for `ResponseFormat` where `ToolStrategy` typically expects a Pydantic `BaseModel`
- Passes Anthropic API key via `environment_variables` instead of using the Secrets API, which contradicts the guidance on the secrets page

Recommend replacing with a framework-agnostic example, or using Weave.

---

**5. `from cwsandbox import SandboxExecutionError` is a direct import from the underlying library**
`sandboxes/run-commands.mdx` (line 88)

This is the only place in the docs that imports from `cwsandbox` directly. Every other import goes through `wandb.sandbox`. The issue is that `wandb.sandbox` doesn't re-export any exception classes from `cwsandbox` today (there are 12 missing). Working with Pinglei to get this fixed on the SDK side. Once that lands, this import should become `from wandb.sandbox import SandboxExecutionError`.

---

### Should Fix

**6. Typos**
- `sandboxes/run-commands.mdx` (line 52): "explictly" -> "explicitly", "runnig" -> "running"
- `sandboxes/mltrain_in_sandbox_tutorial.mdx` (line 34) and `sandboxes/invoke_agent_sandbox_tutorial.mdx` (line 46): "documenation" -> "documentation"

**7. `.mdx` extension in links**
`sandboxes/lifecycle.mdx` (line 20)

Links use `/sandboxes/create-sandbox.mdx` and `/sandboxes/run-commands.mdx` but every other page drops the `.mdx` extension. Should be `/sandboxes/create-sandbox` and `/sandboxes/run-commands`.

**8. Line number references off by 2**
`sandboxes/mltrain_in_sandbox_tutorial.mdx` (lines 223-224)

"Lines 29-31" should be "Lines 27-29" (the three print statements). "Lines 35-36" should be "Lines 33-34" (read_file and write_bytes).

**9. `docs.json` formatting**
The new entry uses `"pages" :` (space before colon) while the rest of the file uses `"pages":`.

sandboxes/create-sandbox.mdx
```diff
@@ -0,0 +1,79 @@
...
+title: Create sandboxes
+description: Learn how to create and manage sandboxes.
+---
+
+Create sandboxes with W&B. Each sandbox runs in its own container with its own [filesystem](/sandboxes/file-access), network, and process space.
+
+Use [`Sandbox`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox) to create a single sandbox. Use [`Session`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/session) to create and manage multiple sandboxes that share the same configuration.
+
+<Info>
+By default, sandboxes use `python:3.11` as the base image. To use a different image, pass `container_image` to `Sandbox.run()` or `SandboxDefaults()`. W&B supports public container images only.
```
iiilisan: I think this would be more helpful as a code snippet than in prose

`Sandbox.run(image="your-image-name:latest")`

Did not double check the syntax here.

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,92 @@
...
+| RUNNING     | Container is up, ready for operations | 
+| COMPLETED   | Main process exited with code 0     | 
+| FAILED      | Startup or runtime error            | 
+| TERMINATED   | Externally killed or lifetime exceeded |
+
+A sandbox can accept most operations only after it reaches RUNNING. After it enters a terminal state, it is no longer available for additional work.
+
+## Wait for a sandbox to start
+
+Use `Sandbox.wait()` when you want to confirm that the sandbox starts (sandbox reached RUNNING state) successfully before performing other operations, or when you want to separate startup errors from later command errors.
```
iiilisan: This is a pattern that's generally not that useful.

IIRC Under the hood, we implicitly wait on any call (exec, read, write.... ??).

```
with Sandbox.run() as sandbox:
    sandbox.exec() # waits until sandbox is ready and then calls the exec
```

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+## Stop a sandbox
+
+Use [`Sandbox.stop()`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox#stop) when you no longer need the sandbox, when you want to end a long-running process, or when you need to clean up resources before the sandbox reaches its maximum lifetime.
+
+```python
+from wandb.sandbox import Sandbox
+
+with Sandbox.run("sleep", "infinity") as sandbox:
+    sandbox.stop().result()
```
iiilisan: Also a pattern that's helpful but generally not that useful.

The python context manager will clean up on exit

```
with ContextManagerThing as sandbox:
    sandbox.do_stuff()

# automatically cleaned up when it hits this line (aka exits the context)
```

We should highlight this context manager usage since it's more ergonomic than calling stop on sandboxes

sandboxes/create-sandbox.mdx
```diff
@@ -0,0 +1,79 @@
...
+```python
+from wandb.sandbox import Sandbox, NetworkOptions
+
+with Sandbox.run(
+    network=NetworkOptions(egress_mode="internet"),
+) as sandbox:
+    print(sandbox)
+```
+
+W&B recommends using a context manager to ensure that the sandbox is stopped automatically when it exits the block. To learn how to run commands inside a sandbox, see [Run commands](/sandboxes/run-commands). To learn about sandbox lifecycle and states, see [Lifecycle](/sandboxes/lifecycle).
```
iiilisan: I'd tweak this to say the context manager pattern

sandboxes/create-sandbox.mdx
```diff
@@ -0,0 +1,79 @@
...
+This pattern is useful for interactive and multi-step workflows. To learn how to run commands inside a sandbox, see [Run commands](/sandboxes/run-commands).
+
+### Start a sandbox with a main command
+
+You can also pass a command to `Sandbox.run()`. In that case, the command becomes the sandbox's main process.
+
+```python
+from wandb.sandbox import Sandbox
+
+sandbox = Sandbox.run("python", "train.py")
```
iiilisan, line 66: I'd need to double check but this I think is the fire and forget pattern.
They could then get the result by calling `sandbox.wait_until_complete().result()`

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+### Run a single command
+
+To run a single command, pass the command directly to `Sandbox.run()`. The sandbox stops automatically when the command finishes and is no longer available for additional work.
+
+```python
+from wandb.sandbox import Sandbox
+
+sandbox = Sandbox.run("python", "train.py")
+sandbox.wait_until_complete(timeout=3600.0)
```
iiilisan: `result = sandbox.wait_until_complete(timeout=3600.0).result()` ?

should probably highlight how to get a result if they want one.

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+To run multiple commands in the same sandbox, start the sandbox without a command. Then use `Sandbox.exec()` to run commands in the running sandbox.
+
+The following code snippet starts a sandbox without a main command, then runs two commands sequentially in the same sandbox: first it installs PyTorch, then it runs a training script:
+
+```python
+from wandb.sandbox import Sandbox
+
+with Sandbox.run() as sandbox:
+    result = sandbox.exec(["pip", "install", "torch"], check=True).result()
```
iiilisan: I think we could leave the check=True out of this example for now. the param name is a little confusing about what it does and is perhaps noisy to introduce something like this without explaining it as well.

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,96 @@
...
+    print(result.stdout)
+
+    result = sandbox.exec(["python", "train.py"], check=True).result()
+    print(result.stdout)
+```
+
+The `Sandbox.exec()` method runs a command inside a running sandbox and returns a `Process` object that you can use to wait for completion, read output, and inspect the exit code.
```
iiilisan: can we link to the API for Process?

sandboxes/file-access.mdx
```diff
@@ -0,0 +1,93 @@
+---
+title: File access
```
iiilisan: File Operations perhaps makes the most sense.

sandboxes/file-access.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+# Path of the local file you want to write to the sandbox
+text_file = Path("hello.txt")
+
+with Sandbox.run() as sandbox:
+    # Write a file to the sandbox
+    sandbox.write_file("hello.txt", text_file.read_bytes()).result()
+```
+
+See the `Sandbox` class reference documentation for a full list of parameters and options for [`Sandbox.write_file()`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox#write_file).
```
iiilisan, line 43: Generally speaking, would be good to link to the class as well. I find myself wanting to click `Thing` whenever I see it

sandboxes/file-access.mdx
```diff
@@ -0,0 +1,93 @@
...
+from pathlib import Path
+from wandb.sandbox import Sandbox, NetworkOptions
+
+mounted_files = [
+    {"mount_path": "requirements.txt", "file_content": Path("requirements.txt").read_bytes()},
+    {"mount_path": "train.py", "file_content": Path("train.py").read_bytes()},
+        ] 
+
+print("Starting sandbox...")
+with Sandbox.run(mounted_files=mounted_files) as sandbox:
```
iiilisan, line 85: Oh wow. This is a documentation gap in the CW docs. We'll need to update this. @NavarrePratt FYI

sandboxes/invoke_agent_sandbox_tutorial.mdx
```diff
@@ -0,0 +1,179 @@
+---
+title: "Tutorial: Invoke an agent in a sandbox"
+description: "Learn how to invoke a LangChain agent within a sandbox environment"
```
iiilisan: Should we be cross-selling the WB Agent here instead of Langchain?

sandboxes.mdx
```diff
@@ -0,0 +1,56 @@
...
+```
+
+The sandbox automatically stops when the context manager (the `with` block) exits. For more information on sandbox lifecycle and states, see [Lifecycle of sandboxes](/sandboxes/lifecycle).
+
+
+## Tutorials and examples
+
+For more in-depth examples, see:
+
+* [W&B Sandbox tutorial](/sandboxes/invoke_agent_sandbox_tutorial)
```
iiilisan: Should this be cross selling the WB-Agent? Seems weird to provide a tutorial for another company's product.

sandboxes/mltrain_in_sandbox_tutorial.mdx
```diff
@@ -0,0 +1,224 @@
...
+    # Save the generated model file locally
+    print("Downloading zoo_wandb.pth...")
+    model_data = sandbox.read_file("zoo_wandb.pth").result()
+    Path("zoo_wandb.pth").write_bytes(model_data)
+    print("Saved zoo_wandb.pth")
+```
+
+The previous code snippet does the following:
+
+1. (Lines 6 - 9) Mount files to the sandbox at startup. The `mounted_files` parameter in `Sandbox.run()` allows you to specify a list of files to mount into the sandbox at startup. Each file is represented as a dictionary with two keys: `mount_path`, which specifies the path where the file will be mounted inside the sandbox, and `file_content`, which contains the content of the file as bytes. In this example, you mount `train.py` and `requirements.txt` into the sandbox.
```
iiilisan: I wonder if these are better as inline comments. 🤷

sandboxes/file-access.mdx
NavarrePratt: I deprioritized this because I don't want to push the mounting files thing that hard because it's extremely likely it changes once we get to implementing the CAIOS integration. I'd rather people use read and write file for now. I'm pretty sure this mount file thing is leftover from the swebaas hackathon idea.

sandboxes/secrets.mdx
```diff
@@ -0,0 +1,103 @@
...
+        "-c",
+        """
+import os
+print(f"Hugging Face token: {os.environ['HUGGINGFACE_TOKEN'][:8]}...")
+""",
+    ]).result()
+    print(result.stdout)
+```
+
+## Extract a field from a structured secret
```
nicholaspun-wandb: I don't think we actually support this with W&B Secrets Manager

sandboxes/invoke_agent_sandbox_tutorial.mdx
```diff
@@ -0,0 +1,179 @@
...
+from wandb.sandbox import Sandbox, NetworkOptions
+
+script = Path("demo.py").read_bytes()
+
+print("Starting sandbox...")
+with Sandbox.run(
+    container_image="python:3.13",
+    network=NetworkOptions(egress_mode="internet"),
+    max_lifetime_seconds=3600,
+    environment_variables={"ANTHROPIC_API_KEY": os.environ["ANTHROPIC_API_KEY"]},
```
nicholaspun-wandb: Let's use a secret here to promote better practices 😅

trane293:
## Review (Round 2)

Good progress since the last review, 5 of 9 items fully addressed. Flagging what's remaining.

### Must Fix

**1. `result.stdout` / `result.stderr` will crash at runtime**
`run-commands.mdx` (lines 25-26)

```python
result = sandbox.wait_until_complete(timeout=3600.0).result()
print(f"Standard output: {result.stdout}")
print(f"Standard error: {result.stderr}")
```

`wait_until_complete().result()` returns a `Sandbox` object, not a `ProcessResult`. `Sandbox` has `.returncode` but does not have `.stdout` or `.stderr`, so both lines raise `AttributeError`. The correct pattern (which `lifecycle.mdx` already shows correctly) is:

```python
sandbox.wait_until_complete(timeout=3600.0).result()
print(f"Exit code: {sandbox.returncode}")
```

If you want to show stdout/stderr, use `sandbox.exec()` instead, which returns a `Process` whose `.result()` gives a `ProcessResult` with `stdout`, `stderr`, and `returncode`.

---

**2. `from cwsandbox import SandboxExecutionError` can now be `from wandb.sandbox`**
`run-commands.mdx` (line 88)

Confirmed that Pinglei's branch (`pinglei/sandbox-auth-v3`, PR wandb/wandb#11661) exports all 13 user-facing exception classes from `wandb.sandbox`, including `SandboxExecutionError`. This line should become:

```python
from wandb.sandbox import SandboxExecutionError
```

---

**3. Typo still present**
`run-commands.mdx` (line 52)

`"long runnig main process"` should be `"long running main process"`

---

### Should Fix

**4. Auth language is still slightly misleading**
`sandboxes.mdx` (line 18)

> "Sandboxes authenticate using your existing W&B credentials."

This conflates two things: authenticating to the sandbox service (creating/managing sandboxes, which does use your W&B credentials) and using W&B inside the sandbox (which does not work automatically, credentials are not propagated into the container). Suggest rewording to: "W&B authenticates your identity when you create and manage sandboxes. To use W&B or Weave inside a sandbox, pass your API key using the Secrets Manager or environment variables."

---

**5. Consider adding a short note about features on the roadmap**
`sandboxes.mdx`

Worth mentioning somewhere (even a single sentence) that GPU support, persistent disk snapshots, and other capabilities are on the roadmap. Sets expectations for private preview users.

---

### Low Severity

**6. Broken anchor in secrets.mdx (line 10)**
`[structured secrets](#extract-a-field-from-a-structured-secret)` links to a section that is currently commented out (lines 78-103 in `{/* */}`). Either remove the reference or uncomment the section.

**7. Wrong anchor in sandboxes.mdx (line 18)**
`/platform/secrets#manage-secrets` should be `/platform/secrets#manage-access-to-secrets` (the actual heading is "Manage access to secrets").

**8. 404 external link in sandboxes.mdx (line 12)**
`https://docs.coreweave.com/products/coreweave-sandbox/client/overview` returns 404 (CI already flagged this).

**9. `docs.json` formatting (line 1153)**
`"pages" : [` has a space before the colon, rest of file uses `"pages":`. Also trailing whitespace on line 1163.

**10. Stray notebook cell ID in secrets.mdx (line 28)**
` ```python id="11o5yi" ` has an orphaned ID attribute, no other code block uses this.

**11. Missing `description` in frontmatter**
`sandboxes.mdx` and `mltrain_in_sandbox_tutorial.mdx` are missing the `description` field that other tab landing pages include.

**12. File naming inconsistency**
Tutorial files use underscores (`invoke_agent_sandbox_tutorial.mdx`) while other pages use hyphens. Rest of the docs repo uses hyphens consistently.

**13. Unused imports in a few snippets**
`lifecycle.mdx` imports `wandb` and `NetworkOptions` but never uses them. `invoke_agent_sandbox_tutorial.mdx` imports `os` but never uses it.

sandboxes/invoke-agent-sandbox-tutorial.mdx
```diff
@@ -0,0 +1,230 @@
+---
+title: "Tutorial: Invoke an agent in a sandbox"
+description: "Learn how to invoke OpenAI agent within a W&B Sandbox environment"
```
iiilisan, line 3: Should we be cross selling our inference product here? It seems weird to be shilling OpenAI.

ngrayluna: @iiilisan The goal was to show a dummy example on how to call an agent within a sandbox. I was struggling to find a agnostic framework that users might be aware of so I chose OpenAI. 🤷 

I can remove this tutorial if we want. I can make it a follow up task post 4/15 to figure out how to make a W&B Inference + Sandbox example.

trane293: good points, i think OpenAI is fine, not really a direct competitor and we want to make sure that all users are able to get started quickly (W&B Inference is a bit more involved to get started with)

trane293:
## Review (Round 3)

Cross-referenced all code snippets against the SDK source (`cwsandboxes-client` v0.16.0 and `wandb/sandbox` on `pinglei/sandbox-auth-v3`). Ran offline verification for import paths, return types, and enum values.

### Must Fix

**1. `print(result)` prints object repr, not script output**
`invoke-agent-sandbox-tutorial.mdx`

```python
result = sandbox.exec(["python", "demo.py"], check=True).result()
print("Script output:")
print(result)  # prints ProcessResult(stdout='...', stderr='...', returncode=0, stdout_bytes=b'...', ...)
```

`exec().result()` returns a `ProcessResult` dataclass with no custom `__str__`. This dumps the full repr with all fields including raw bytes. Change to `print(result.stdout)`.

**2. `result.stdout` / `result.stderr` after `wait_until_complete` will crash**
`run-commands.mdx`

If any code still shows:
```python
result = sandbox.wait_until_complete(timeout=3600.0).result()
print(f"Standard output: {result.stdout}")
```

This raises `AttributeError`. `wait_until_complete().result()` returns a `Sandbox` object, which has `.returncode` and `.status` but not `.stdout` or `.stderr`. Remove the `.stdout`/`.stderr` lines and use `sandbox.returncode` only. If you want to show stdout/stderr, use an `exec()` example instead:

```python
sandbox.wait_until_complete(timeout=3600.0).result()
print(f"Exit code: {sandbox.returncode}")
```

**3. Auth language still slightly misleading**
`sandboxes.mdx`

> "Sandboxes authenticate using your existing W&B credentials."

This conflates authenticating to the sandbox service (which does use your W&B credentials) with using W&B inside the sandbox (which does not work automatically). Credentials are not propagated into the container. Reword to: "W&B authenticates your identity when you create and manage sandboxes. To use W&B or Weave inside a sandbox, pass your API key using the [Secrets Manager](/sandboxes/secrets) or `environment_variables`."

### Should Fix

**4. `result` variable naming is misleading across multiple pages**
Files: `run-commands.mdx`, `create-sandbox.mdx`, `lifecycle.mdx`

```python
result = sandbox.wait_until_complete(...).result()
```

This assigns a `Sandbox` object to a variable called `result`. Readers will expect a `ProcessResult`. Drop the assignment and call `.result()` inline, then use `sandbox.returncode`:

```python
sandbox.wait_until_complete(timeout=3600.0).result()
print(f"Exit code: {sandbox.returncode}")
```

**5. Context manager + `wait_until_complete` is a confusing pattern**
`lifecycle.mdx`

```python
with Sandbox.run("python", "train.py") as sandbox:
    sandbox.wait_until_complete(timeout=3600.0).result()
```

The context manager calls `stop()` on exit, but the sandbox is already in a terminal state after `wait_until_complete`. Two things managing the lifecycle at once. The SDK's own docstring examples don't combine these. Split into two separate examples:

**Interactive sandbox (context manager):**
```python
with Sandbox.run() as sandbox:
    result = sandbox.exec(["python", "train.py"]).result()
    print(result.stdout)
# sandbox automatically stopped on exit
```

**Fire-and-forget with main command (no context manager):**
```python
sandbox = Sandbox.run("python", "train.py")
sandbox.wait_until_complete(timeout=3600.0).result()
print(f"Exit code: {sandbox.returncode}")
```

**6. Missing PAUSED state from lifecycle page**
`lifecycle.mdx`

The SDK defines `SandboxStatus.PAUSED` and includes it in `_RUNNING_STATUSES`. The state diagram and table omit it. Either add a row to the table or add a note that PAUSED is not expected in the W&B hosted environment.

**7. `sleep infinity` pattern doesn't mention default behavior**
`run-commands.mdx`

The "Keep the sandbox running" section shows `Sandbox.run("sleep", "infinity")` without mentioning that `Sandbox.run()` with no args already uses `tail -f /dev/null` as a keep-alive. Add: "Note: `Sandbox.run()` with no arguments keeps the sandbox alive by default."

**8. `session.sandbox()` returns unstarted sandbox, not explained**
`create-sandbox.mdx`

The Session example calls `session.sandbox()` and prints the result. No mention that this returns an unstarted sandbox that auto-starts on first operation. Add: "`session.sandbox()` returns an unstarted sandbox. It auto-starts when you call `exec()`, `read_file()`, or any other operation."

### Low Severity

**9. Missing `description` in frontmatter** - `sandboxes.mdx` and `mltrain-in-sandbox-tutorial.mdx` lack the `description` field that other landing pages include.

**10. Stray `id` attribute** - `secrets.mdx` line 28 has ` ```python id="11o5yi" ` which no other code block uses. Remove the `id`.

**11. Broken anchor in secrets.mdx** - `[structured secrets](#extract-a-field-from-a-structured-secret)` links to a section that's commented out. Either remove the link or uncomment the section.

**12. File naming inconsistency** - Tutorial files use underscores (`invoke_agent_sandbox_tutorial.mdx`) while other pages use hyphens. Rest of the docs repo uses hyphens consistently. Rename to match (and update `docs.json`).

support/models/articles/how-can-i-recover-deleted-runs.mdx
mdlinville, line 1: LGTM

docs.json
mdlinville, line 1: LGTM

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
+---
+title: Sandbox lifecycle
```
mdlinville, line 2: This may be better as a task: "Manage sandboxes" or something like that. I expected it to be mostly a conceptual page from the title.

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
...
+description: Learn about sandbox states, waiting, completion, and stopping.
```
mdlinville: This reads a little weird to me, maybe because some of the items seem like simple nouns and others seem like how-to actions. And it seems to run off a cliff after "stopping".

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
...
+---
+
+A sandbox moves through a series of states during its lifetime. These states determine which operations are available and what actions the sandbox can make.
```
mdlinville: Does the sandbox "make" (take?) actions? Or does the sandbox confine the scope of actions?

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+In most cases, a sandbox starts in `PENDING`, moves to `CREATING` while the container is provisioned, and then enters `RUNNING` when the container is up.
+
+The RUNNING state is when you can perform most operations, such as running commands with `Sandbox.exec()`. When the main process of the sandbox exits, the sandbox enters a terminal state such as `COMPLETED` (if the exit code is 0) or `FAILED` (if there was an error). A sandbox can also be `TERMINATED` if it is externally killed or if it exceeds its maximum lifetime.
```
mdlinville:
```suggestion
The `RUNNING` state is when you can perform most operations, such as running commands with `Sandbox.exec()`. When the main process of the sandbox exits, the sandbox enters a terminal state such as `COMPLETED` (if the exit code is 0) or `FAILED` (if there was an error). A sandbox can also be `TERMINATED` if it is externally killed or if it exceeds its maximum lifetime.
```
Formatting consistency nit

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+```text
```
mdlinville, line 16: Maybe nice as a Mermaid diagram

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
...
+PENDING -> CREATING -> RUNNING -> COMPLETED
+                               -> FAILED
+                               -> TERMINATED
+```
+
+Use this page to understand when a sandbox is ready, how to wait for state changes, and how to stop a sandbox.
```
mdlinville: I don't find this sentence to be super useful. It might be better to explain that this page explains the various sandbox states, explains how the automatic lifecycle works,  and explains how to manage a sandbox directly.

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+The next sections provide details on sandbox states, waiting for a sandbox to start or complete, and stopping a sandbox. For more information on creating sandboxes and running commands, see [Create sandboxes](/sandboxes/create-sandbox) and [Run commands](/sandboxes/run-commands).
```
mdlinville: The first sentence here feels redundant. The calls to action might be better on their own.

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+## Sandbox states
+
+The following table summarizes the states of a sandbox:
+
+| State       | Description                         | 
+|-------------|-------------------------------------|
+| PENDING     | Start accepted, waiting for scheduling |
```
mdlinville: Formatting nits: Earlier, these states were marked up in monospace.

Does "start accepted" mean "The sandbox has started and is queued waiting to be provisioned"? 

In RUNNING, "is up" feels kind of informal.

In COMPLETED, maybe "Main process exited cleanly with status `0`.", and failed could be "Sandbox failed to start or exited with a non-zero status." But are these really conflated? One would be after `CREATING` and the other would be after `RUNNING`, right?

Next, TERMINATED says externally killed or lifetime exceeded. What kind of "externally killed"? Normal termination by the process owner, or `SIGKILL` which forces the kernel to do the deed? Both?

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
...
+| State       | Description                         | 
+|-------------|-------------------------------------|
+| PENDING     | Start accepted, waiting for scheduling | 
+| CREATING    | Container is being provisioned      | 
+| RUNNING     | Container is up, ready for operations | 
+| COMPLETED   | Main process exited with code 0     | 
+| FAILED      | Startup or runtime error            | 
+| TERMINATED   | Externally killed or lifetime exceeded |
+
+A sandbox can accept most operations only after it reaches RUNNING. After it enters a terminal state, it is no longer available for additional work.
```
mdlinville: Really, it's while it is in `RUNNING`. After it reaches RUNNING, it could then transition to COMPLETED, FAILED, or TERMINATED, where it also couldn't accept most operations, I think.

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+## Wait for a sandbox to complete
+
+Use [`Sandbox.wait_until_complete()`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox#wait_until_complete) when the sandbox's main process represents the full workload and you want to wait for that job to finish.
```
mdlinville, line 66: Is this something like "block until the sandbox completes"?

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+When the main process exits, the sandbox enters a terminal state. A successful run typically ends in COMPLETED. If the process fails, the sandbox may enter FAILED.
+
+## Explicit lifecycle control
+
+Use methods such as `Sandbox.wait()` or `Sandbox.stop()` to explicitly control over the sandbox lifecycle. 
+
+In most cases, you do not need to call these methods directly. Operations such as `Sandbox.exec()`, `Sandbox.read_file()`, and `Sandbox.write_file()` automatically wait for readiness and the context manager (`with Sandbox.run() as sandbox:`) automatically stops the sandbox when the block exits.
+
+{/* However, you can use these methods when you need more control over the lifecycle, such as when debugging startup problems or when you want to distinguish startup failures from errors in later commands. */}
```
mdlinville: I'm confused why this is commented out.

sandboxes/lifecycle.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+
+### Wait for a sandbox to start
```
mdlinville: Why is this not nearer "Wait for a sandbox to stop"? It seems like it should be ahead of that?

sandboxes.mdx
```diff
@@ -0,0 +1,58 @@
+---
+title: W&B Sandboxes
+mode: wide
```
mdlinville, line 3: Consider adding a description

sandboxes.mdx
```diff
@@ -0,0 +1,58 @@
...
+---
+
+<Warning>
+W&B Sandboxes is in active development and in private preview. Contact W&B Support at support@wandb.com to request access to this feature.
```
mdlinville: When announcing private previews in the release notes, I tend to use language like this. I'd consider moving the admonition down below line 10, so the reader can get a little excited about it before we rain on their parade.

```
W&B Sandboxes is in private preview, available by invitation only. To request enrollment, contact [support](mailto:support@wandb.com) or your AISE.
```

sandboxes.mdx
```diff
@@ -0,0 +1,58 @@
...
+</Warning>
+
+W&B Sandboxes gives you on-demand, isolated compute environments that you can create, use, and discard using Python. No cluster management, no Kubernetes YAML.
```
mdlinville: Nit: Last sentence is a fragment.

sandboxes.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+W&B Sandboxes is built on top of the CoreWeave Sandbox library. For the underlying API reference and library docs, see the [CoreWeave Sandbox documentation](https://docs.coreweave.com/products/coreweave-sandbox).
```
mdlinville:
```suggestion
W&B Sandboxes is built using the CoreWeave Sandbox library. For the underlying API reference and library docs, see the [CoreWeave Sandbox documentation](https://docs.coreweave.com/products/coreweave-sandbox).
```

sandboxes.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+## How it works
+
+A *sandbox* is a single isolated compute environment. You [create](/sandboxes/create-sandbox) it, [run commands inside it](/sandboxes/run-commands), and stop it when it is done. Each sandbox runs in its own container with its own filesystem, network, and process space.
+
```
mdlinville, line 17: Nit: Extra vertical whitespace sprinkled around in these pages

A diagram or two could help to make it clear how this works. For example, I found myself wondering about what "its own network" would look like for a networked workload.

sandboxes.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+W&B authenticates your identity when you create and manage sandboxes. To use W&B or Weave inside a sandbox, pass your API key using the [W&B Secrets Manager](/platform/secrets#manage-access-to-secrets) or environment variables. See [Secrets](/sandboxes/secrets) for more information about using secrets in sandboxes.
+
+
+A sandbox goes through [several states in its lifecycle](/sandboxes/lifecycle). When a container is running, you can [execute commands](/sandboxes/run-commands) inside it.
+
+[Read, write, and mount read-only](/sandboxes/file-access) files to and from the sandbox. Common examples include reading in a Python script to execute, writing out logs or results, or mounting a directory of data read-only for the sandbox to access.
```
mdlinville, line 22: This seems like maybe too much detail for the overview. I got snagged on "read" vs "read-only" for a minute.

sandboxes.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+Use a *session* to manage multiple sandboxes that share configuration. When a session closes, all of its sandboxes are stopped automatically. See [Manage multiple sandboxes](/sandboxes/create-sandbox#create-multiple-sandboxes-with-a-session) for more details.
+
+## Basic usage
+
+{/* The following code snippet shows how to create a sandbox, run a command inside it, and print the output. */}
```
mdlinville: Stray comments?

sandboxes.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+Copy and paste the following code snippet into a Python file and run it. You should see the output `Hello from W&B Sandboxes!` printed to the console.
+
+<Tip>
+Ensure you are logged in to W&B before running the code snippet below. Use the `wandb login` CLI command and follow the prompts to log in to your W&B account. For more information on how W&B searches for credentials, see the [`wandb login`](/models/ref/cli/wandb-login) reference documentation.
```
mdlinville: This feels out of order. Would it be better to put this above line 32? Or show logging in as part of the code snippet itself?

sandboxes/create-sandbox.mdx
```diff
@@ -0,0 +1,100 @@
+---
+title: Create sandboxes
+description: Learn how to create and manage sandboxes.
```
mdlinville: I'm not seeing much about managing here.

sandboxes/create-sandbox.mdx
```diff
@@ -0,0 +1,100 @@
...
+---
+
+Create sandboxes with W&B. Each sandbox runs in its own container with its own [filesystem](/sandboxes/file-access), network, and process space.
```
mdlinville, line 10:
```suggestion
This page shows how to create a W&B Sandbox, with its own container, [filesystem](/sandboxes/file-access), network, and process space.
```

sandboxes/create-sandbox.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+
+<Info>
```
mdlinville, line 12: This seems like it could be moved further down to an advanced section.

sandboxes/create-sandbox.mdx
```diff
@@ -0,0 +1,100 @@
...
+By default, sandboxes use `python:3.11` as the base image. To use a different image, pass `container_image` to [`Sandbox.run()`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox#run) or [`SandboxDefaults`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox-defaults). W&B supports public container images only.
+
+```python
+from wandb.sandbox import Sandbox
+
+with Sandbox.run(container_image="python:3.15") as sandbox:
```
mdlinville, line 20: It might be better to show 3.11 here. It's the default so that is a bit redundant, but sometimes a code block makes it harder to take in the detail of the paragraphs near it, and this could get copy-pasted into customer code as is, and then they will be pinning the Python version for no real reason.

sandboxes/create-sandbox.mdx
```diff
@@ -0,0 +1,100 @@
...
+    sandbox.exec(["python", "--version"]).result()
+```
+
+</Info>
+
+## Create a single sandbox
```
mdlinville:
```suggestion
## Create a sandbox
```

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
+---
+title: Run commands
```
mdlinville: This might be better as the sidebarTitle and have the title mention sandboxes, like "Run commands in a sandbox".

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+description: Learn how to run commands in a sandbox environment and read output and exit codes.
+---
+
+Run shell commands, scripts, and other processes in sandboxes you create with W&B.
```
mdlinville: What about "operating system commands"? This is another situation where lines 6 and 9 seem to just restate each other.

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+
+This page shows how to run commands, check output and exit codes, and reuse the same sandbox for multiple steps.
+
+## Run commands in a sandbox
+
+Based on your use case, use one of these patterns to run commands in a sandbox:
+
+* [Single-command sandbox](#run-a-single-command): Start the sandbox with a main command if that command is the entire workload. Then call `Sandbox.wait_until_complete()` to wait for the sandbox to reach a terminal state. Use this pattern for batch jobs, tests, or training runs.
```
mdlinville:
```suggestion
* [Single-command sandbox](#run-a-single-command): Start the sandbox with a main command if that command is the entire workload. To wait for the sandbox to reach a terminal state, call `Sandbox.wait_until_complete()`. Use this pattern for batch jobs, tests, or training runs.
```
WDYT? But also, these bullets largely duplicate the sections below, so I wonder if the specific details in the bullets can be minimized.

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+* [Multi-command sandbox](#run-multiple-commands-sequentially): Start the sandbox without a main command when you need to run several commands in the same environment. Then use `Sandbox.exec()` for each step and wait on each `Sandbox.exec()` call. Use this pattern for workflows that include installing dependencies, writing files, running a script, and reading results within the same sandbox.
```
mdlinville:
```suggestion
* [Multi-command sandbox](#run-multiple-commands-sequentially): Start the sandbox without a main command when you need to run several commands in the same environment. In sequence, start each command with `Sandbox.exec()` and wait for the call to complete. Use this pattern to install dependencies, write files, run scripts, and read results in the sandbox.
```
WDYT?

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+
+### Run a single command
+
+To run a single command as the sandbox's main process, pass the command to `Sandbox.run()`. Then call `Sandbox.wait_until_complete()` to wait for the command to finish and for the sandbox to reach a terminal state.
```
mdlinville: The "Then" sentence is a fragment. This might be better as 2 numbered steps.

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+from wandb.sandbox import Sandbox
+
+sandbox = Sandbox.run("python", "train.py")
+result = sandbox.wait_until_complete(timeout=3600.0).result()
+print(f"Exit code: {result.returncode}")
+```
+
+### Run multiple commands sequentially
+
+Use the same sandbox for workflows that require multiple steps, such as installing dependencies, writing files, running a script, and reading results.
```
mdlinville:
```suggestion
Run multiple commands in a given sandbox for workflows with multiple steps. For example, install dependencies, write files, run scripts, and read results.
```

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+To run multiple commands in the same sandbox, start the sandbox without a main command. Then use `Sandbox.exec()` to run each command in the running sandbox.
```
mdlinville:
```suggestion
To run multiple commands in a sandbox, start the sandbox without a main command. Use a series of `Sandbox.exec()` to run each command and wait for it to complete.
```

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+with Sandbox.run() as sandbox:
+    result = sandbox.exec(["pip", "install", "torch"]).result()
+    print(result.stdout)
+
+    result = sandbox.exec(["python", "train.py"]).result()
+    print(result.stdout)
+```
+
+`Sandbox.exec()` runs a command in a running sandbox and returns a `Process` object. You can use the returned object to wait for the command to finish, read output, and inspect the exit code.
```
mdlinville:
```suggestion
`Sandbox.exec()` runs a command in the sandbox and returns a `Process` object. You can use the returned object to wait for the command to finish, read output, and inspect the exit code.
```

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+For more information about `Sandbox.exec()` and its parameters, see the [`Sandbox.exec()`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox#exec) reference documentation.
```
mdlinville:
```suggestion
For more information about `Sandbox.exec()` and its parameters, see [`Sandbox.exec()`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox#exec) in the CoreWeave Sandboxes reference documentation.
```

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+#### Keep the sandbox running for additional commands
```
mdlinville, line 60: This might be better promoted to an H3 and titled something like "Keep a sandbox running for a period of time". or even "indefinitely". The example is indefinite but also makes it pretty obvious that you could sleep for a given period instead.

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+with Sandbox.run("sleep", "infinity") as sandbox:
+    result = sandbox.exec(["python", "train.py"]).result()
+    print(result.stdout)
+```
+
+## Read output and exit codes
+
+`Sandbox.exec()` returns a [`Process`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/types/process) object. Call `Process.result()` to wait for the command to finish and get its result. `Process.result()` includes the standard output (`stdout`), standard error (`stderr`), and exit code (`returncode`).
+
+Use the exit code to determine whether the command succeeded. By convention, an exit code of 0 indicates success.
```
mdlinville:
```suggestion
Use the exit code to determine whether the command succeeded. By convention, an exit code of 0 indicates success. The possible exit codes and their meanings depend on the command.
```

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+For example, the following code snippet runs a Python command that prints the standard output (`stdout`) and the exit code (`returncode`). After the command finishes, it prints the standard output, standard error, and exit code to the console.
+
+```python
+from wandb.sandbox import Sandbox
+
+with Sandbox.run() as sandbox:
+    result = sandbox.exec(["python", "-c", "print('hello')"]).result()
+    print(result.stdout)      # "hello\\n"
```
mdlinville, line 87: Are we missing `stderr` here?

sandboxes/run-commands.mdx
```diff
@@ -0,0 +1,101 @@
...
+    print(result.returncode)  # 0
+```
+
+## Check for sandbox execution errors
+
+Specify `check=True` (`Sandbox.exec(check=True)`) to raise an error if the command exits with a non-zero exit code. A [`SandboxExecutionError`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/exceptions/exceptions#sandboxexecutionerror) is raised if the command exits with a non-zero code:
```
mdlinville:
```suggestion
Specify `check=True` (`Sandbox.exec(check=True)`) to raise an error if the command exits with a non-zero exit code. A [`SandboxExecutionError`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/exceptions/exceptions#sandboxexecutionerror) is raised if the command exits with a non-zero code. This example runs a script `might_fail.py` and checks the result.
```

sandboxes/secrets.mdx
```diff
@@ -0,0 +1,104 @@
...
+W&B injects each requested secret into the sandbox as an environment variable. By default, the environment variable name matches the secret name. You can customize the environment variable name with the `env_var` parameter. 
+
+
+<Note>
+See [Manage secrets](/platform/secrets) for information about creating and managing secrets in your W&B account.
+</Note>
+
+## Access secrets in a sandbox
+
+Use `wandb.sandbox.Secret` to request access to secrets stored in your team's Secrets Manager. Pass one or more `Secret` objects to the `secrets` parameter of `Sandbox.run()`.
```
mdlinville: This is a bit confusing to me and it's written a bit different from how other examples are described. I think the first sentence is about the import statement.  And I think what you are passing to the `Sandbox.run` looks like a series of secret names.

sandboxes/secrets.mdx
```diff
@@ -0,0 +1,104 @@
...
+
+Each `Secret` object identifies a secret by name. W&B injects the secret value into the sandbox as an environment variable.
+
+<Tip>
+Pass the secret name, not the secret value. Within the sandbox, read the secret from the corresponding environment variable.
+</Tip>
+
+The following example makes two existing secrets available in the sandbox: `HF_TOKEN` and `OPENAI_API_KEY`.
```
mdlinville, line 29: What about "adds two secrets to the sandbox" or "inserts two secrets into the sandbox's environment"?

sandboxes/secrets.mdx
```diff
@@ -0,0 +1,104 @@
...
+        Secret(name="HF_TOKEN"),
+        Secret(name="OPENAI_API_KEY"),
+    ],
+) as sandbox:
+    result = sandbox.exec([
+        "python",
+        "-c",
+        """
+import os
+print(f"HF token: {os.environ['HF_TOKEN'][:8]}...")
```
mdlinville: Does this print out the secret materials? That seems potentially dangerous, but maybe I'm being overly cautious.

sandboxes/secrets.mdx
```diff
@@ -0,0 +1,104 @@
...
+print(f"OpenAI key: {os.environ['OPENAI_API_KEY'][:8]}...")
+""",
+    ]).result()
+    print(result.stdout)
+```
+
+Inside the sandbox, access each secret through its environment variable. In this example, `HF_TOKEN` is available as `os.environ["HF_TOKEN"]`, and `OPENAI_API_KEY` is available as `os.environ["OPENAI_API_KEY"]`.
```
mdlinville, line 54: Is it worth stating that shell scripts can access the secrets using `${HF_TOKEN}` etc?

sandboxes/secrets.mdx
```diff
@@ -0,0 +1,104 @@
...
+
+To use a different environment variable name, see [Use a custom environment variable name](#use-a-custom-environment-variable-name).
```
mdlinville:
```suggestion
To customize the environment variable name for a secret, see [Use a custom environment variable name](#use-a-custom-environment-variable-name).
```

sandboxes/secrets.mdx
```diff
@@ -0,0 +1,104 @@
...
+
+## Use a custom environment variable name
+
+By default, a secret's environment variable name matches the secret name. To use a different name in the sandbox, set the `env_var` parameter on `Secret` (`Secret(env_var="CUSTOM_NAME")`).
```
mdlinville:
```suggestion
By default, a secret's environment variable name matches the secret name. To customize a sandbox's environment variable for a given secret, set the `env_var` parameter on `Secret` (`Secret(env_var="CUSTOM_NAME")`).
```

sandboxes/file-access.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+# Path of the local file you want to write to the sandbox
+text_file = Path("hello.txt")
+
+with Sandbox.run() as sandbox:
+    # Write a file to the sandbox
+    sandbox.write_file("hello.txt", text_file.read_bytes()).result()
+```
+
+See the [`Sandbox`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox) class reference documentation for a full list of parameters and options for [`Sandbox.write_file()`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox#write_file).
```
mdlinville: Is the first link needed when the second link takes you straight to the correct section?

sandboxes/file-access.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+## Read a file from the sandbox
+
+{/* accessing, viewing, or loading data from a storage medium */}
+{/* transfers a file out of the sandbox. Use it to download outputs your sandbox code generated (models, results, logs). */}
+
+Save a file from the sandbox to your local environment using the [`Sandbox.read_file()`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox#read_file) method.
```
mdlinville, line 47: This is conflating two different concepts that are each interesting on their own: Reading a file, and saving a file. I wonder if it is worth separating them.

sandboxes/file-access.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+    # Save the file on local machine (not sandbox)
+    content = sandbox.read_file("hello.txt").result()  # b"Hello, world.\n"
+    Path("hello.txt").write_bytes(content)
+```
+
+See the `Sandbox` class reference documentation for a full list of parameters and options for [`Sandbox.read_file()`](https://docs.coreweave.com/products/coreweave-sandbox/client/ref/core/sandbox#read_file).
+
+
+## Mount files
```
mdlinville:
```suggestion
## Mount a file or directory
```
Aligns with the previous section heading. But this may be even better as "Read a file or directory"

sandboxes/file-access.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+Mount files or directories to share large files or datasets between your local environment and the sandbox without copying them. Mounted files are available in the sandbox at the path you specify.
```
mdlinville:
```suggestion
To share a large file or directory, such as a dataset, between your local environment and the sandbox without copying it, you can mount the file or directory in the sandbox in read-only mode, and specify the path for the mount point in the sandbox's filesystem.
```
I don't know, maybe this isn't any better. I was surprised about the `<Info>` so my instinct is to mention the read-only thing a bit earlier in the piece.

sandboxes/file-access.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+<Info>
+Mounted files are read-only in the sandbox. If you need to modify files in the sandbox, use `Sandbox.write_file()` instead.
+</Info>
+
+In the following code snippet, local files `train.py` and `requirements.txt` are mounted to the sandbox root directory. After you mount the files, you can use them like any other files in the sandbox. The sandbox installs dependencies from `requirements.txt` and then runs `train.py`.
```
mdlinville: "use them like any other files in the sandbox" seems a little broad, especially after getting the news about it being read-only. Maybe like any other read-only files in the sandbox? I believe that an `Info` admonition is considered optional info that the reader could skip, but that seems like a really important distinction.

sandboxes/invoke-agent-sandbox-tutorial.mdx
mdlinville: Inference is available only on Multi-tenant Cloud, FYI

sandboxes/invoke-agent-sandbox-tutorial.mdx
```diff
@@ -0,0 +1,246 @@
...
+
+Run the [`wandb login`](/models/ref/cli/wandb-login) CLI command and follow the prompts to log in to your W&B account:
+
+```bash
+wandb login
+```
+
+### Store API keys in Secret Manager
+
+Store your OpenAI API key in the [W&B Secret Manager](/platform/secrets) as `OPENAI_API_KEY`. See [Manage secrets](/platform/secrets) for information about creating and managing secrets in your W&B account.
```
mdlinville: What about linking to `sandboxes/secrets.mdx` for the second link? Otherwise it is redundant.

Nit: Extra vertical whitespace in some spots in this page.

mdlinville (approved):
Pre-approving after taking a first pass through all of the pages in this PR. Let me know if you'd like a second look at anything before you merge.

trane293 (approved):

PR #2439, opened 2026-04-09
Add workflow to auto-update docs-skills submodule every day

johndmulhausen:
## Summary

- Adds `.github/workflows/update-submodule-docs-skills.yml` that runs daily (and on manual dispatch), checks if the `.claude` submodule is behind `coreweave/docs-skills`, and opens a draft PR if so.

## Setup required

A repo secret **`DOCS_SKILLS_PAT`** must be created with a PAT (or fine-grained token) that has read access to `coreweave/docs-skills`. The workflow uses this token at checkout to clone the private submodule.

## Test plan

- [ ] Create the `DOCS_SKILLS_PAT` secret on the repo.
- [ ] After merge, trigger the workflow manually via the Actions tab to verify it runs.
- [ ] Confirm a submodule-bump PR is opened when upstream has new commits.


johndmulhausen:
Sadly this simply will not work while there is a hard org-level wall between W&B and CW that prevents us from pulling from the upstream skills repo

PR #2446, opened 2026-04-09
Allow Ellipsis in card previews

dfinster:
Allows the "⋯" character in Knowledgebase Card previews.


mdlinville:
I think we moved on from this approach so maybe you can close this PR out. :)

PR #2464, opened 2026-04-15
WEAVE: Accumulator documentation

dbrian57:
## Description
Resolves [WBDOCS-409](https://coreweave.atlassian.net/browse/WBDOCS-409). Adds a new section to the generator function docs that teaches users how to use the `accumulator` parameter with `weave.op()`.

[WBDOCS-409]: https://wandb.atlassian.net/browse/WBDOCS-409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/guides/tracking/trace-generator-func.mdx
+++ b/weave/guides/tracking/trace-generator-func.mdx
@@ -72,7 +72,7 @@ Weave captures generator outputs only after you fully consume the generator. Con
 
 For more on decorating functions and methods with `@weave.op`, see [Create calls](/weave/guides/tracking/create-call).
 
-## Use the accumulator parameter
+## Accumulate yielded values into a single trace
 
 You can use the `weave.op`'s `accumulator` parameter to customize how yielded values are combined from generator functions, for example, to join streamed text tokens into a single string. The accumulator is a two-argument function that Weave calls once per yielded value, building up a result incrementally.
 
--- a/weave/guides/tracking/trace-generator-func.mdx
+++ b/weave/guides/tracking/trace-generator-func.mdx
@@ -71,3 +71,33 @@ The following screenshot shows the **Traces** page with a selected trace of the 
 Weave captures generator outputs only after you fully consume the generator. Consume the generator by iterating over it (for example, with `list()`, a `for` loop, or `next()` until exhaustion). The same applies to async generators when you use `async for` or equivalent consumption.
 
 For more on decorating functions and methods with `@weave.op`, see [Create calls](/weave/guides/tracking/create-call).
+
+## Use the accumulator parameter
+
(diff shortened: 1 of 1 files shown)


weave/guides/tracking/trace-generator-func.mdx
```diff
@@ -71,3 +71,37 @@ The following screenshot shows the **Traces** page with a selected trace of the
...
+The following example demonstrates a custom accumulator that appends each yielded value to a list, so Weave records that list as the call output after the generator is fully consumed.
+
+```python
+from typing import Generator
+import weave
+
+weave.init("your-team-name/your-project-name")
+
+# Weave calls this after each yield; acc is None on the first call.
+# The last value you return becomes the traced op output.
```
anastasiaguspan:
```suggestion
# The last value you return becomes the traced Op output.
```

weave/guides/tracking/trace-generator-func.mdx
anastasiaguspan:
```suggestion
    # Tracing here produces a hierarchical Call tree:
```

anastasiaguspan:
```suggestion
    # Each yielded value comes from a separate traced Call to `inner`.
```

anastasiaguspan:
```suggestion
    # Weave will trace the Call and its input (`x`),
```

weave/guides/tracking/trace-generator-func.mdx
```diff
@@ -71,3 +71,37 @@ The following screenshot shows the **Traces** page with a selected trace of the
 Weave captures generator outputs only after you fully consume the generator. Consume the generator by iterating over it (for example, with `list()`, a `for` loop, or `next()` until exhaustion). The same applies to async generators when you use `async for` or equivalent consumption.
 
 For more on decorating functions and methods with `@weave.op`, see [Create calls](/weave/guides/tracking/create-call).
+
+## Use the accumulator parameter
```
anastasiaguspan: To make this header something people would be looking for, perhaps describing the function would be better? like:
Accumulate yielded values into a single trace

anastasiaguspan (approved):
capitalized Ops/Calls throughout - including the older content :) I must have missed this file.  lots of 'calls' on this page, that aren't Calls - do double check me. but i think this is one of those pages where the caps Calls helps clarify over function calls.

weave/guides/tracking/trace-generator-func.mdx
dbrian57: I totally meant to change this, but that is a very good header and I'm gonna use it.

PR #2480, opened 2026-04-17
Sweeps updates

ngrayluna:
## Description

Realized that Sweeps flattens YAML objects to make accessing values easier, however, it made debugging my script a pain. Updated docs to make this more clear


models/sweeps/add-w-and-b-to-your-code.mdx
```diff
@@ -64,13 +64,17 @@ To use the W&B Python SDK to start, stop, and manage sweeps, follow the instruct
 
 <Tabs>
 <Tab title="CLI">
-Create a YAML configuration file with your sweep configuration. The
-configuration file contains the hyperparameters you want the sweep to explore. In
-the following example, the batch size (`batch_size`), epochs (`epochs`), and
-the learning rate (`lr`) hyperparameters are varied during each sweep.
+Create a YAML file that defines the hyperparameters to optimize and the metric to optimize. W&B uses this file to determine which hyperparameters to vary during the sweep and which metric to optimize.
```
kelu-wandb, line 67: nit:
>  the hyperparameters to optimize and the metric to optimize.
Feels a bit redundant. Maybe:
> the hyperparameters and metric to optimize.
?

models/sweeps/add-w-and-b-to-your-code.mdx
```diff
@@ -89,38 +93,45 @@ parameters:
...
+Within your training script, add the following code snippets to integrate W&B:
 
-1. Import the W&B Python SDK (`wandb`) and PyYAML (`yaml`). PyYAML is used to read in our YAML configuration file.
-2. Read in the configuration file.
-3. Use [`wandb.init()`](/models/ref/python/functions/init) to start a background process to sync and log data as a [W&B Run](/models/ref/python/experiments/run). Pass the config object to the config parameter.
-4. Define hyperparameter values from `wandb.Run.config` instead of using hard coded values.
-5. Log the metric you want to optimize with [`wandb.Run.log()`](/models/ref/python/experiments/run.md/#method-runlog). You must log the metric defined in your configuration. Within the configuration dictionary (`sweep_configuration` in this example) you define the sweep to maximize the `val_acc` value.
+1. Import the W&B Python SDK (`wandb`).
+2. Initialize a [run](/models/runs) with `wandb.init()`.
+3. Read the YAML configuration file with a Python package such as yaml, and pass the configuration to `wandb.init()`.
```
kelu-wandb, line 102: Steps 3 and 4 are not accurate. The training script for sweeps training just needs to call wandb.init() with no arguments, as the sweeps agents will automatically pass in the appropriate arguments.

models/sweeps/add-w-and-b-to-your-code.mdx
```diff
@@ -89,38 +93,45 @@ parameters:
...
+4. Pass the configuration object to the config parameter of `wandb.init()`.
+5. Retrieve the hyperparameter values from `wandb.Run.config` so that your script uses the values defined in the YAML file instead of hard-coded values. W&B flattens configuration values, so you can access nested values with dot notation or bracket notation as though they were top-level keys.
```
kelu-wandb, line 104: Probably more idiomatic to just use `wandb.config`.

> so that your script uses the values defined in the YAML file 

I would say:

> so that your script uses the suggested arguments for each run

models/sweeps/add-w-and-b-to-your-code.mdx
```diff
@@ -89,38 +93,45 @@ parameters:
...
+6. Log the metric that you want to optimize with `wandb.Run.log()`.
```
kelu-wandb, line 105: Similarly, `wandb.log()` is more idiomatic now

models/sweeps/add-w-and-b-to-your-code.mdx
```diff
@@ -89,38 +93,45 @@ parameters:
...
 def evaluate_one_epoch(epoch):
+    """Simulates evaluation for one epoch and returns the validation accuracy and loss."""
     acc = 0.1 + ((epoch / 20) + (random.random() / 10))
     loss = 0.25 + (1 - ((epoch - 1) / 10 + random.random() / 6))
     return acc, loss
 
-
 def main():
-    # Set up your default hyperparameters
+    # Read in the configuration file
```
kelu-wandb, line 134: This is a bit out of date. The most idiomatic way to do this would be to just go:

```
with wandb init() as run:
```

with no arguments. No need to read any config files. It'll pick up the arguments automatically.

You can note that if you DO set `config` in `wandb.init` for a single run case, it will actually be overridden by the sweep arguments. (That's why this code as written still does work; it's just confusing because the config it loads gets ignored.)

models/sweeps/add-w-and-b-to-your-code.mdx
```diff
@@ -142,9 +153,62 @@ def main():
...
+```python
+import yaml
+
+with open("sample.yaml") as file:
+    yaml_sample = yaml.load(file, Loader=yaml.FullLoader)
+```
+
+You can then access `nested_value1` with `yaml_sample["key2"]["nested_key1"]` or `yaml_sample.key2.nested_key1`.
+
+When you pass a configuration to `wandb.init(config=)`, W&B flattens the values. This means that you access nested values as though they were top-level keys.
```
kelu-wandb, line 179: I think the previous documentation was confusing and caused you to conflate RUN configs with SWEEPS configs.

```
with open("config.yaml") as file:
    config = yaml.load(file, Loader=yaml.FullLoader)
with wandb.init(config=config) as run:
```

This is intended to open a RUN config file, not a SWEEPS config file. You don't actually need to specify any run config file when using Sweeps.

That said, that run config file, as mentioned above, is promptly IGNORED by the sweep args generated from the sweep config that was provided to `wandb sweep`.

models/sweeps/add-w-and-b-to-your-code.mdx
```diff
@@ -142,9 +153,62 @@ def main():
...
+    name: val_acc
+parameters:
+epochs:
+    values: [10, 20, 30]
+learning_rate:
+    min: 0.001
+    max: 0.1
+```
+
+After you read in the file and pass the configuration to `wandb.init(config=)`, access the `goal` value with `run.config["goal"]` instead of `run.config["metric"]["goal"]` or `run.config.metric.goal`.
```
kelu-wandb, line 198: Yeah this is mixing up the files. The issue here is that you've provided a sweep config to use as a run config... and then the run config is being ignored in favor of the args generated by the sweep and silently passed in...  (The only thing that confuses me is that I don't think there should even exist a `goal` field at all??)

In any case please rewrite the code to just use `wandb.init()` with no config loading at all, and see what happens then.

models/sweeps/define-sweep-configuration.mdx
```diff
@@ -46,13 +47,14 @@ parameters:
   optimizer:
     values: ["adam", "sgd"]
 ```
+
+Within the top level `parameters` key (line 7), the following keys are nested: `learning_rate` (line 8), `batch_size` (line 11), `epochs` (line 14), and `optimizer` (line 17). For each of the nested keys you specify, you can provide one or more values, a distribution, a probability, and more.
```
kelu-wandb, line 51: I wouldn't really frame the first layer of keys under `parameters` as nested. Those are actually the top-level parameters. It's only if you nest another level deep that I'd call it nested.

kelu-wandb (changes requested):

models/sweeps/add-w-and-b-to-your-code.mdx
ngrayluna: The SDK Team generally advises us to use `wandb.Run.log()` for a worry-free run management (i.e. no active, hanging runs)  🤔

mdlinville (approved):

PR #2483, opened 2026-04-17
chore: share Cursor MCP plugin settings repo-wide

johndmulhausen:
## Summary
Track `.cursor/settings.json` so contributors using Cursor get the same set of enabled MCP plugins by default, without each person having to enable them individually. Per-developer Cursor state under `.cursor/` stays ignored.

## Changes
- **.gitignore**: Replace blanket `.cursor` rule with `.cursor/*` plus `!.cursor/settings.json` so only the shared settings file is tracked.
- **.cursor/settings.json** (new): Enables the Atlassian, Slack, Notion-workspace, and Glean MCP plugins. These are the plugins the docs team uses day-to-day for ticket lookup, cross-team context, doc research, and internal search.

## Verification
- `git check-ignore -v .cursor/settings.json` resolves to the `!.cursor/settings.json` negation (tracked).
- `git check-ignore -v .cursor/_testfile.tmp` resolves to `.cursor/*` (ignored), confirming other per-developer state still does not land in the repo.

## Notes
- Cursor reads `.cursor/settings.json` as the repo-level plugin config; users can still override or add personal plugins in their user settings.
- No content or docs changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,9 +15,6 @@
 go.mod
 go.sum
 
-# Cursor editor local state (per-developer), but share repo-wide settings
-.cursor/*
-!.cursor/settings.json
 
 .DS_Store
 .env.local
--- a/.gitignore
+++ b/.gitignore
@@ -15,9 +15,9 @@
 go.mod
 go.sum
 
-# Misc
-# Cursor editor local rules and config (per-developer)
-.cursor
+# Cursor editor local state (per-developer), but share repo-wide settings
+.cursor/*
+!.cursor/settings.json
 
 .DS_Store
 .env.local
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -15,9 +15,9 @@
 go.mod
 go.sum
 
-# Misc
-# Cursor editor local rules and config (per-developer)
-.cursor
+# Cursor editor local state (per-developer), but share repo-wide settings
+.cursor/*
+!.cursor/settings.json
```
johndmulhausen:
```suggestion
```

anastasiaguspan (approved):

PR #2530, opened 2026-04-23
docs(on-prem/airgapped): Update outdated images versions

kkaushik-wandb:
## Description
Updated outdated image versions for W&B Server, Console and Controller.

- Updated Server version from 0.59.2 to 0.78.2 (older version is archived, and 0.78.2 is latest-1)
- Updated Console version from 2.12.2 to 2.18.12 (which was over 1 year old)
- Updated Controller version from 1.13.3 to 1.21.3 (which was over 1 year old)


mdlinville:
Apologies but this has gone stale and can't go in as is. Closing here but please resurrect in coreweave/docs if you'd like.

PR #2532, opened 2026-04-23
Inference - Finer grained reasoning information

jamie-rasmussen:
## Description

Update Inference's reasoning page to provide more information about how to enable/disable it. (To date, we've forced it on for all models that support it, but we expect to change that soon.) Add a dynamically generated table of models that support reasoning along with a column about configurability.

Changed the example model to Gemma 4 so users would only have to change one thing in the code example if they want to try out reasoning vs. not.

image

## Testing
- [x] PR tests succeed

### DIFF
--- a/inference/response-settings/reasoning.mdx
+++ b/inference/response-settings/reasoning.mdx
@@ -3,16 +3,15 @@ title: "View reasoning information"
 description: How to return and view reasoning in your W&B Inference responses
 ---
 
-Reasoning models, like [Google's Gemma 4](https://huggingface.co/google/gemma-4-31B-it), include information about their reasoning steps as part of the output returned in addition to the final answer.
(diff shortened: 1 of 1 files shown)


inference/response-settings/reasoning.mdx
```diff
@@ -3,45 +3,78 @@ title: "View reasoning information"
...
 
-You can find reasoning information in the `reasoning` field of responses. The value of this field is `null` in the responses of non-reasoning models.
+Reasoning information appears in the `reasoning` field of responses. The value of this field is `null` in the responses of non-reasoning models.
+
+## Supported models with reasoning
+
+The following table lists the models on W&B Inference that return reasoning output. Each supported model might always include reasoning, or might disable or enable reasoning by default:
+
+{/* takeru reasoning-models - This table is automatically generated, do not edit manually. */}
+| Model ID (for API usage)                       | Support            |
```
anastasiaguspan: Last suggestion i'd make is that this column header "support" is vague if an LLM slurped this out of context.  I'd recommend "Reasoning" as better, or since we have the room "Reasoning support".  this would be in your takeru script tho.

anastasiaguspan (approved):

PR #2539, opened 2026-04-24
W&B CLI: update for v0.26.1

ngrayluna:
## Description

Updates CLI Docs for v0.26.1. Biggest change is to the _snippets/cli/leet* doc. 

Future self: We should consider moving this content elsewhere. It's a bit odd to import a snippet to docs derived from source. i.e. there should be a 1:1 between the docstring and the public docs.


dmitryduev (approved):
Exellente!

PR #2597, opened 2026-05-04
Updates MCP menu and doc to prioritize Claude over Cursor

dbrian57:
## Description
Updates the MCP menu and MCP doc to prioritize Claude Code because it is "industry standard" now.


mdlinville (approved):
I'll approve to unblock, but it would be good to have PM approval too.

johndmulhausen (dismissed):
a) It's Claude Code not Claude CLI
b) I think we should just provide the generic instructions here instead of picking a winner

PR #2650, opened 2026-05-20
Remove ENABLE_REGISTRY_UI

willgoldby:
## Description

Remove `ENABLE_REGISTRY_UI` from docs.

## Related issues

Fixes: https://coreweave.atlassian.net/browse/DOCS-2567


platform/hosting/env-vars.mdx
```diff
@@ -29,8 +29,7 @@ In addition to configuring instance-level settings through the System Settings a
...
-| `ENABLE_REGISTRY_UI`               | When set to true, enables the new W&B Registry UI.            |
+| `GRAPHQL_REJECT_UNAUTHED_REQUESTS` | When set to `true` on the **API** service, rejects GraphQL requests that do not have an authenticated user. Unauthenticated requests receive HTTP 401. **Self-Managed** and **Dedicated Cloud** v0.80.0+ only; not available on Multi-tenant Cloud. This feature is opt-in: if the environment variable is unset or not `true`, behavior is unchanged. Set on the API component only (for example, `api.env` in Helm values). Before activating, confirm that workflows that rely on anonymous GraphQL access (such as viewing shared reports without signing in, or open projects) still meet your requirements. On **Self-Managed**, setting the Gorilla configuration field `graphql-reject-unauthed-requests` to `true` has the same effect. |          |
```
johndmulhausen:
```suggestion
| `GRAPHQL_REJECT_UNAUTHED_REQUESTS` | When set to `true` on the **API** service, rejects GraphQL requests that do not have an authenticated user. Unauthenticated requests receive HTTP 401. **Self-Managed** and **Dedicated Cloud** v0.80.0+ only; not available on Multi-tenant Cloud. This feature is opt-in: if the environment variable is unset or not `true`, behavior is unchanged. Set on the API component only (for example, `api.env` in Helm values). Before activating, confirm that workflows that rely on anonymous GraphQL access (such as viewing shared reports without signing in, or open projects) still meet your requirements. On **Self-Managed**, setting the Gorilla configuration field `graphql-reject-unauthed-requests` to `true` has the same effect. |
```

johndmulhausen (approved):

PR #2710, opened 2026-06-02
Fix bold/emphasis text rendering at normal weight site-wide (DOCS-2705)

johndmulhausen:
## Summary

`**bold**` text (and literal `<b>`) rendered at the normal body weight everywhere on the site — release notes and regular docs pages alike — so emphasized text was not visually heavier than the surrounding copy. Reported in [DOCS-2705](https://coreweave.atlassian.net/browse/DOCS-2705).

## Root cause

The custom CSS layer overrides the Mintlify **Aspen** theme but **never defined a rule for `<strong>`/`<b>`**, so inline emphasis fell through to the inherited body weight (`400`). The design token `--font-weight-bold: 650` already exists in `typography.css` and its comment even earmarks it for *"group headers, active nav, emphasis"* — but no rule ever applied it to emphasis.

Two things compounded this:
1. **No `strong`/`b` rule** in any partial (`scripts/css-minify/*.css`).
2. **`docs.json` pins both fonts to `weight: 400`**, so Mintlify requests only the 400 weight from Google Fonts (`...Source+Sans+3:ital,wght@0,400;1,400`) — no dedicated bold face is ever downloaded.

## Fix

Add a single rule to the `typography.css` partial:

```css
strong, b {
  font-weight: var(--font-weight-bold) !important;   /* 650 */
  font-synthesis-weight: auto;
}
```

- Uses the existing `--font-weight-bold` token, so emphasis now matches the active-nav (`topbar.css`) and card-heading (`cards.css`) rules that already use it.
- `!important` plus our stylesheet's load-order advantage wins over Aspen's base rule — the same mechanism the existing `h1–h6` override relies on.
- `font-synthesis-weight: auto` is the browser default, set explicitly so an upstream `font-synthesis: none` can't suppress the synthesized heavier weight given only the 400 face is loaded. This is the same path the existing 650-weight nav/card rules already take.

`css/styles.css` is the generated bundle and was rebuilt with `npm run build` (`scripts/css-minify`). Per the build setup, the **Build CSS** workflow would also regenerate it on this PR; the committed bundle matches that output.

## How to verify

On the Mintlify preview for this PR, confirm bold renders heavier than body text on:
- A release-notes page (e.g. `/release-notes/...`)
- `/platform/app/settings-page/user-settings#edit-your-intro` — "Edit" should now be bold

Inspecting a `<strong>` should show computed `font-weight: 650` instead of `400`.

## Scope note

This intentionally does **not** change the font loading in `docs.json` (which can only pin one weight). The site already standardizes on the synthesized 650 weight for its other bold UI; this change makes inline emphasis consistent with that. Loading true bold font faces site-wide would be a larger, separate enhancement.

[DOCS-2705]: https://coreweave.atlassian.net/browse/DOCS-2705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mdlinville:
Weird, I wonder why it didn't build a Mintlify preview. Built locally to verify the fix, and now the deploy seems to be running after all.

mdlinville (approved):

PR #2745, opened 2026-06-09
Workspaces:  Compare runs across projects feature

ngrayluna:
## Description

Adds section on comparing runs across projects feature (via pinned columns).

~## Open questions:~
~Is there guidance why run(s) may not show on plots, run comparer? E.g. Here I pinned runs from another project and then created a Run Comparer.~


julia-g-rose:
Noted offline, cross-project pinned runs will only show up for full fidelity line plots, not other panel types

models/runs/compare-runs.mdx
```diff
@@ -28,7 +30,20 @@ See [Limitations](#limitations).
 
 To unpin a run, click the pin icon, or follow the procedure to pin the run, but select **Unpin run** instead.
 
-## Manage the baseline run
+## Compare runs across projects
+
+Compare runs from different projects by selecting runs from another project and pinning them to your current workspace.
+
+1. Navigate to your project's workspace.
+2. Click on the **Select runs from another project** button (small square box with rounded corners and a diagonal arrow).
```
julia-g-rose: Maybe we should mention the location of the button, which is above the runs side-bar in the workspace, or runs table on the tables page

models/runs/compare-runs.mdx
```diff
@@ -28,7 +30,20 @@ See [Limitations](#limitations).
...
+3. In the modal, select a project from the **Source project** dropdown.
+4. Search the source project’s runs to find the run you want to compare. You can search by run name or unique ID.
+5. Select the checkbox next to each run you want to compare, then click **Pin runs**.
+
+W&B shows pinned runs in your project's run selector and runs table. Hover your mouse over the pinned run's name to see the project the run is from. Runs from other projects have an open circle icon next to their name.
```
julia-g-rose: I'm not sure the best place to mention this, but currently we have an upper limit of 20 pinned runs per workspace. This is true regardless of whether it is a pinned run within the project or a pinned run from another project, since its a workspace-level cap.

We could also mention that pinning cross project runs is a workspace-level action, affecting a specific view like your personal workspace, rather than effecting everyone

Another note is that the run logs and overview page and other details link back to the original run in a new tab. That info is not imported over to the current workspace/project where the run was imported.

models/runs/compare-runs.mdx
```diff
@@ -2,33 +2,62 @@
 description: Learn how to use pinned and baseline runs to keep track of important runs and efficiently evaluate model experiments.
 title: Pin and compare runs
 ---
-import PinRunsCondensed from '/snippets/_includes/pinned-and-baseline-runs/pin-runs-condensed.mdx';
+It can be difficult to keep track of runs that are your best performers, production models, failed experiments, or important reference points. You can use the W&B App to help organize and compare runs with:
```
dbrian57: I'm not the biggest fan of this sort of marketing intro to things. IMO, instead of stating a plight and then a solution, we should just state what the feature does and the problem it solves. This is a knit, because I know we use this other places. Up to you!

models/runs/compare-runs.mdx
```diff
@@ -2,33 +2,62 @@
...
+You can pin up to 20 runs in a workspace. If you have a baseline run, you can pin up to 19 runs because the baseline is implicitly pinned.
 
 <Frame>
 ![Runs table with pinned runs](/images/models/pinned-and-baseline-runs/runs-table-with-pinned-runs.png)
 </Frame>
 
 To unpin a run, click the pin icon, or follow the procedure to pin the run, but select **Unpin run** instead.
 
-## Manage the baseline run
+Runs that you pin to a project only impact your [personal workspace view](/models/track/workspaces#saved-workspace-views), including [runs pinned from another project](/models/runs/compare-runs#compare-runs-across-projects).
```
julia-g-rose: nit: it could be pinned in a saved view, not just personal workspace view

models/runs/compare-runs.mdx
```diff
@@ -2,33 +2,62 @@
...
+  </Frame>
+3. In the modal, select a project from the **Source project** dropdown.
+4. Search the source project’s runs to find the run you want to compare. You can search by run name or unique ID.
+5. Select the checkbox next to each run you want to compare, then click **Pin runs**.
+
+View runs that you pin from another project in the run selector or runs table. Pinned runs from other projects have an open circle icon next to their name. To see which project a pinned run comes from, hover over the run name.
+
+Artifacts, logs, and other details for pinned runs link back to the original run. W&B does not import this data into the current project or workspace. To view the details for a run pinned from another project, click the run. W&B opens the original run details in a new browser tab.
+
+W&B does not currently support [Run Comparer](/models/app/features/panels/run-comparer#add-a-run-comparer-panel) panels for runs pinned from another project. To compare runs from different projects, pin the runs you want to compare, then use line plots to visually compare them.
```
julia-g-rose: it's not limited to run comparer panels, we don't support any panels besides full fidelity line plots. So things like media panels from the run will also not show up, or custom charts, and other panel types.

models/runs/compare-runs.mdx
```diff
@@ -2,33 +2,62 @@
 description: Learn how to use pinned and baseline runs to keep track of important runs and efficiently evaluate model experiments.
 title: Pin and compare runs
 ---
-import PinRunsCondensed from '/snippets/_includes/pinned-and-baseline-runs/pin-runs-condensed.mdx';
+It can be difficult to keep track of runs that are your best performers, production models, failed experiments, or important reference points. You can use the W&B App to help organize and compare runs with:
 
-In a workspace with many runs, it can be difficult to keep track of your best performers, production models, failed experiments, or important reference points. The W&B App provides features to help organize and compare runs:
-
-- **Pinned runs**: Pin up to 6 runs to keep them visible in the workspace and at the top of the runs list. If you have a baseline run, you can pin up to 5 runs because the baseline is implicitly pinned.
+- **Pinned runs**: Pin runs from the current or different projects in the same workspace to keep them visible in the workspace and at the top of the runs list. If you have a baseline run, you can pin up to 5 runs because the baseline is implicitly pinned.
```
dbrian57:
```suggestion
- **Pinned runs**: Pin runs from any project in your workspace to keep them visible at the top of the runs list. If you have a baseline run, you can pin up to five runs because the baseline is implicitly pinned.
```
Tried to make this a little more concise, but does it still make sense to the product? The second line sounds a little awkward, but I'm not sure how to fix that because I think my familiarity with pinning is too limited here. Are you pinning runs against that baseline?

dbrian57 (approved):
Minor suggestions inline. Looks good otherwise 👍

models/runs/compare-runs.mdx
ngrayluna: Same. 😅 

I replaced it with:

Use the W&B App to organize, identify, and compare important runs, such as top performers, production models, failed experiments, and reference runs. To accomplish this, you can organize and compare runs with:

julia-g-rose (approved):

PR #2753, opened 2026-06-11
Registry: Add info about SA access to different types of registries

ngrayluna:
## Description
Add info about SA access to different types of registries.

## Related issues

https://coreweave.atlassian.net/browse/DOCS-2549


models/registry/configure_registry.mdx
```diff
@@ -80,11 +80,22 @@ W&B automatically assigns a default **registry role** to a user or team when the
...
 
 A registry admin can assign or modify roles for users and teams in the registry.
 See [Configure user roles in a registry](/models/registry/configure_registry/#configure-registry-roles) for more information.
 
+### Service account access
+
+Service accounts cannot have **Viewer** or **Restricted Viewer** roles. W&B determines a service account's access to a registry with the following rules:
+
+- **Registries with Organization visibility**: a service account automatically has **Member** access.
+- **Registries with Restricted visibility**: service acount access depends on the role assigned to the service account's team:
```
ibindlish: This currently reads as though service accounts can only inherit permissions through their teams. It's only in the following point that it becomes clear service accounts can also be added directly and assigned roles independently. Would it make sense to clarify that here to avoid that initial impression?

models/registry/configure_registry.mdx
```diff
@@ -80,11 +80,22 @@ W&B automatically assigns a default **registry role** to a user or team when the
...
+<a id="service_account_footnote">1</a>: Service accounts cannot have **Viewer** or **Restricted Viewer** roles. See [Service account access](#service-account-access) for how a service account's access is determined.
 
 A registry admin can assign or modify roles for users and teams in the registry.
 See [Configure user roles in a registry](/models/registry/configure_registry/#configure-registry-roles) for more information.
 
+### Service account access
+
+Service accounts cannot have **Viewer** or **Restricted Viewer** roles. W&B determines a service account's access to a registry with the following rules:
+
+- **Registries with Organization visibility**: a service account automatically has **Member** access.
```
ibindlish: should we mention that they can be promoted to Admin too by adding them separately and setting their role as Admin?

models/registry/configure_registry.mdx
```diff
@@ -80,11 +80,22 @@ W&B automatically assigns a default **registry role** to a user or team when the
...
+- **Registries with Restricted visibility**: service acount access depends on the role assigned to the service account's team:
+  - If the team is added to the registry as a **Member** or **Admin**, the service account automatically receives **Member** access.
```
ibindlish: worth adding that here if we want the SA to have admin access, they'd have to be added separately.

dfinster (approved):

models/registry/configure_registry.mdx
```diff
@@ -80,11 +80,28 @@ W&B automatically assigns a default **registry role** to a user or team when the
...
 See [Configure user roles in a registry](/models/registry/configure_registry/#configure-registry-roles) for more information.
 
+### Service account access
+
+W&B assigns a default access level to a service account based on the registry's visibility and the service account's team role. A registry admin can modify a service account's access by [adding the service account to the registry with a Member or Admin role](/models/registry/configure_registry/#add-a-user-or-a-team-to-a-registry).
+
+The following describes the default access level for a service account:
+
+- **Registries with Organization visibility**: a service account automatically has **Member** access.
+- **Registries with Restricted visibility**: service acount access depends on the role assigned to the service account's team:
```
ibindlish:
```suggestion
- **Registries with Restricted visibility**: service acount access depends on the role assigned to the service account's team or the service account itself:
```

ibindlish: this still reads to me as though one _needs_ to assign a team role for the service account to receive a role in a restricted registry. i think we should instead focus on how service accounts dont inherit _higher_ permissions just because a team they're in gets assigned "admin" in a registry. (and this is true for both restricted and non-restricted registries)

ngrayluna:
@ibindlish Hopefully this does the trick?

models/registry/configure_registry.mdx
```diff
@@ -63,23 +63,18 @@ This section shows how to configure roles for Registry members. For more informa
 The following sections give more information about Registry roles.
 
 <Note>
-Your [role in a team](/platform/app/settings-page/teams/#team-roles-and-permissions) has no impact or relationship to your role in any registry.
+Your [role in a team](/platform/app/settings-page/teams/#team-roles-and-permissions) does not automatically assign the same role in every registry. For a registry owned by a team, however, W&B considers your role in that team when it calculates your effective registry role.
```
ibindlish: couple of things to note here:
- what does it mean for a team to _own_ a registry?
- we don't consider the user's role in the team while resolving registry access. we just consider the role of the team in the registry + the user's role

models/registry/configure_registry.mdx
```diff
@@ -115,19 +110,62 @@ The following table lists each Registry role, along with the permissions provide
...
-A user's effective role in a particular registry matches their _highest_ role among their role in the organization, the registry, and the team that owns the registry, whether inherited or explicitly assigned. For example:
+W&B calculates the effective role by selecting the user's highest applicable role from these sources:
+
+- Their organization role.
+- Their assigned role in the registry.
+- Their role in the team that owns the registry.
+
+If a higher role comes from the organization or owning team instead of the registry role dropdown, the user inherits that higher role. For example:
 
 - A team **Admin** or organization **Admin** with the **Viewer** role in a particular registry owned by the team is effectively an **Admin** of the registry.
```
ibindlish: i acknowledge this isn't a part of your changes, but not sure what it means for a team to own a registry

models/registry/configure_registry.mdx
```diff
@@ -115,19 +110,62 @@ The following table lists each Registry role, along with the permissions provide
...
 <Frame>
     <img src="/images/registry/role_conflict.png" alt="Registry membership list showing the user's effective registry role"  />
 </Frame>
 
-A user's effective role in a particular registry matches their _highest_ role among their role in the organization, the registry, and the team that owns the registry, whether inherited or explicitly assigned. For example:
+W&B calculates the effective role by selecting the user's highest applicable role from these sources:
+
+- Their organization role.
+- Their assigned role in the registry.
+- Their role in the team that owns the registry.
```
ibindlish:
```suggestion
- The default registry role for the organization, if it is not a restricted registry.
- Their assigned role in the registry.
- The role(s) assigned in the registry to any team(s) they are a part of.
```

ibindlish: I don't think we need to mention what individual role gets assigned to a user when they're added to a registry when explaining what their effective registry role is. wdyt?

ibindlish (commented):
almost there i think! thanks for your patience

models/registry/configure_registry.mdx
```diff
@@ -63,23 +63,18 @@ This section shows how to configure roles for Registry members. For more informa
...
-| User or service account (non admin)    | Restricted Viewer (Server v0.75.0+) <br />Viewer (Server v0.74.x and below) | Restricted Viewer                               |
-| Service account (non admin)            | Member<sup><a href="#service_account_footnote">1</a></sup>                | Member<sup><a href="#service_account_footnote">1</a></sup> |
-| Org admin                              | Admin                                                       | Admin                                           |
+### Role types
 
-<a id="service_account_footnote">1</a>: Service accounts cannot have **Viewer** or **Restricted Viewer** roles. See [Service account access](#service-account-access) for how a service account's access is determined.
+W&B Registry has the following roles:
+- **Restricted Viewer**: Provides read-only access to registry artifacts without the ability to create, update, or delete collections, automations, or other registry resources. [This role is available for Dedicated Cloud and Self-Managed Server v0.75.0 or newer](/models/registry/configure_registry#restricted-viewer-role-details).
```
ibindlish: I think we could add a note that the read-only access in this case means that the user will only be able to access/read the artifact's metadata, and not its files' contents.

ibindlish (approved):
lgtm - i just think it would be good to add the file-read note for the restricted viewer role.

appreciate you going back and forth with me on this one!

PR #2754, opened 2026-06-11
Add a section on config values

ngrayluna:
## Description

Adds a section on how to view config values logged to W&B.

## Related issues

- Fixes https://coreweave.atlassian.net/browse/DOCS-2538

### DIFF
--- a/models/track/config.mdx
+++ b/models/track/config.mdx
@@ -171,7 +171,7 @@ You can update multiple values at a time:
 run.config.update({"lr": 0.1, "channels": 16})
 ```
 
-### Set the configuration after your finishes
+### Set the configuration after your run finishes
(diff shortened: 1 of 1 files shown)


models/track/config.mdx
```diff
@@ -274,6 +274,18 @@ with wandb.init(config=config_dictionary) as run:
     ...
 ```
 
+## View config values in the W&B App
+
+View your config values in the **Config** section of a run's Overview tab. The config values are also available in the **References** section if you pin them.
+
+1. Navigate to your project in the W&B App.
+2. Click the run whose config values you want to view.
```
skullivan: "whose" sounds kind of weird when talking about a run but I think it's technically correct so I'm just gonna leave it alone

skullivan (approved):
👍

PR #2771, opened 2026-06-15
fix(weave): restore broken otel docs page + correct OTLP endpoint guidance

gtarpenning:
## Summary
Two fixes to `weave/guides/tracking/otel.mdx`:

- **Restore the broken page (urgent).** #2757 enabled `twoslash` here, and otel.mdx is the only weave page with twoslash TS blocks. The build now throws a Shiki error, `Language console not found` (raised while twoslash renders hover JSDoc), so the live page serves Mintlify's "🚧 A parsing error occured" boundary instead of content. Reverts the `twoslash` keyword and the `// @noErrors` lines #2757 added, restoring the pre-#2757 content. Verified rendering with `mint dev`.
- **Correct OTLP endpoint guidance.** Renames the example variable `OTEL_EXPORTER_OTLP_ENDPOINT` to `WEAVE_TRACES_ENDPOINT`: it was named identically to the real env var, so copying the value into the env var (or using `opentelemetry-instrument`) makes the OTel SDK append `/v1/traces` -> `.../otel/v1/traces/v1/traces` 404 (~273/day in prod). Adds a callout pointing env-var setups at signal-specific `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`.

## Testing
`mint dev` renders the page (it serves the parse-error boundary on `main`); `mint validate` passes. cc #2757 author re: the twoslash revert.

### DIFF
--- a/weave/guides/tracking/otel.mdx
+++ b/weave/guides/tracking/otel.mdx
@@ -67,8 +67,7 @@ tracer_provider = trace_sdk.TracerProvider(resource=Resource({
     "wandb.project": PROJECT,
 }))
 ```
-```typescript twoslash TypeScript lines
-// @noErrors
+```typescript TypeScript lines
 import { NodeTracerProvider } from "@opentelemetry/sdk-trace-node";
 import { OTLPTraceExporter } from "@opentelemetry/exporter-trace-otlp-proto";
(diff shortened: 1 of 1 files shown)


mdlinville:
Griffin confirms this has already been fixed outside of this PR so I'll close it.

PR #2806, opened 2026-06-24
ci(knowledgebase-nav): skip job for Dependabot PRs

dfinster:
## Problem

The **Knowledgebase Nav** workflow fails on Dependabot PRs (e.g. [run #28116171194](https://github.com/wandb/docs/actions/runs/28116171194/job/83256096718) on #2805, a `docengine` submodule bump).

Dependabot-triggered `pull_request` events run with a restricted token, and the regular `secrets` / `vars` contexts are **empty** (Dependabot secrets live in a separate store). So in the `Require app credentials` step both `vars.DOCS_PR_WRITER_CLIENT_ID` and `secrets.DOCS_PR_WRITER_PRIVATE_KEY` resolve to empty, the guard hits `exit 1`, and the job fails.

## Fix

Add `github.actor != 'dependabot[bot]'` to the **job-level** `if`, matching the convention already used in `docengine-build.yml`. Dependabot PRs now **skip** the job (reported as skipped, not failed) — the same handling already applied to fork PRs, which also lack access to the private `coreweave/docengine` submodule and the App credentials. A maintainer regenerates KB nav after merge.

## Notes

- The job is skipped because it genuinely cannot run on Dependabot PRs (no App token to push the commit-back, no private-submodule access) — this isn't suppressing a meaningful check.
- If "Knowledgebase Nav" is a required status check, confirm a skipped run satisfies branch protection (skipped jobs normally report neutral and pass).


dfinster:
Corrects the issue noted here: 
https://github.com/wandb/docs/pull/2805#issuecomment-4791888827

dfinster:
The root cause is that DependaBot does not have access to the normal action variables, and thus cannot run this workflow with its secrets. DependaBot PRs will never modify Knowledgebase, so it's okay to skip this check when it's a Dependabot PR.

cc: @mdlinville

mdlinville (approved):

PR #2836, opened 2026-06-30
Release notes: note ClickHouse 26.2+ requirement for Server v0.81.0+

amarkov-cw:
## What
Adds an upgrade-prerequisite `<Note>` to the **v0.81.0** server release notes: deployments using the ClickHouse/OLAP features (Registry Search, Run Store Accelerator, or History) need **ClickHouse 26.2+** before upgrading to v0.81.0+.

## Why
W&B Server 0.81.0 introduced a runs-accelerator schema migration (`services/gorilla/clickhouse/migrations/runs/001_init.up.sql`) that uses a **ClickHouse-26.2-only** full-text index — `INDEX … TYPE text(tokenizer = array)`. On older ClickHouse (e.g. 25.x) the migration **fails on boot and the `api`/`glue` pods don't start**. 0.80 and earlier don't carry this migration. Dedicated/managed instances are already on 26.2; this is an on-prem / self-managed concern.

This came up with a self-managed customer and isn't currently documented. It mirrors the existing v0.78 Weave/ClickHouse upgrade note already on this page.

## Notes for reviewer
- Placed once on the **v0.81.0** block with "v0.81.0+" wording so it covers the 0.81.x line. Happy to also echo it onto the 0.81.1/.2/.3 patch blocks (as the v0.78 note was repeated) if you prefer that pattern.
- Wording/placement is yours to adjust — flagging the requirement is the goal.


mdlinville:
We handled this separately and it's already published so I will close this for now, but please reopen it if that's a mistake. Thanks.

PR #2898, opened 2026-07-13
Weave workflow differentiation reorg (LiteTouch ™ strategy)

dbrian57:
## Description
(WIP) Reorganizes the Weave docs to better differentiate between the new Agents workflow and the old `weave.op` workflow.

Updates:
* Weave landing page with new copy and tabs for each workflow
* TOC/nav to appropriately nest features under each workflow
* "What is Weave?" page to better explain the bifurcated workflow story

This also adds new generic OTel Agents workflow and semi-deprecates old OTel workflow


weave/concepts/what-is-weave.mdx
```diff
@@ -19,43 +19,64 @@ Weave provides the following core functionality:
...
 
 If you're building an agent, start with the [agent integration quickstart](/weave/agent-integration-quickstart) to pick a built-in integration, or follow the [custom agents quickstart](/weave/custom-agents-quickstart) to manually instrument your own. To learn how the Weave SDK models agents, see [Trace your agents](/weave/guides/tracking/trace-agents).
 
 If you're using a supported third-party agent harness, such as Claude Code or OpenAI Agent SDK, Weave instruments it automatically with no additional code. See [Integrations](/weave/guides/integrations) for all supported frameworks.
 
-### Application tracing
+### Instrument and trace an application
```
anastasiaguspan: I love the "choose a workflow" addition - it really delivers the message that these are separate things that don't... interweave (lol).
However, i thought we were going to be more explicitly blunt about the old way, by adding "using Weave Ops and Calls" everywhere so they realize that is the fundamental point of the divergence?
so for length maybe this would be something like "Trace functions using Weave Ops"

weave/concepts/what-is-weave.mdx
```diff
@@ -19,43 +19,64 @@ Weave provides the following core functionality:
...
 - Drill down into specific prompts and how answers are produced.
 
 To trace individual functions, follow the Weave [Op tracing quickstart](/weave/quickstart) or learn to use the Weave [Ops and Calls](/weave/guides/tracking/tracing).
 
-If you're using a supported third-party agent framework, such as Claude Code, Weave instruments it automatically with no additional code. See [Integrations](/weave/guides/integrations) for all supported frameworks.
+The Weave SDK also autopatches with many different frameworks and LLM providers to automatically captures traces from LLM calls in your code. See [Integrations](/weave/guides/integrations) for all supported providers and frameworks.
+
+### When to use agent tracing and when to use application instrumentation
+
+Both approaches are fully supported. Choose which workflow to use based on what you want to observe:
```
anastasiaguspan: is this really true though?, that the agent path is "fully supported"?

docs.json
```diff
@@ -550,6 +562,63 @@
                       {
                         "group": "Trace your applications",
                         "pages": [
+                          {
+                            "group": "Integrate with your LLM provider and frameworks",
```
anastasiaguspan, line 577: yes, moving the 'old' integrations under applications I think will help alot

docs.json
```diff
@@ -550,6 +562,63 @@
                       {
                         "group": "Trace your applications",
```
anastasiaguspan: same comment, what aobut 'Trace functions using Ops and Calls' or 'Trace using Ops and Calls'

anastasiaguspan: that way it ties understanding to a literal (Weave ops and calls) that they can match into all the pages and code examples, instead of trying to map conceptually 'okay, what makes an application different than an agent, what if my agent is in an application'

docs.json
anastasiaguspan: Is this accurate saying "Evaluate your agents and applications" ? doesn't this imply this section is supported by the agentic workflows...which they aren't yet?

docs.json
```diff
@@ -548,8 +560,65 @@
                         ]
                       },
                       {
-                        "group": "Trace your applications",
+                        "group": "Trace functions with ops and calls",
```
anastasiaguspan:
```suggestion
                        "group": "Trace functions with Ops and Calls",
```

weave.mdx
```diff
@@ -40,6 +42,36 @@ The following docs guide you through the basics of how to use Weave's suite of t
   </Card>
 
 </CardGroup>
+</Tab>
+<Tab title="Trace functions with ops and calls">
```
anastasiaguspan:
```suggestion
<Tab title="Trace functions with Ops and Calls">
```

weave.mdx
```diff
@@ -40,6 +42,36 @@ The following docs guide you through the basics of how to use Weave's suite of t
...
+<CardGroup cols={3}>
+  <Card
+    title="Quickstart: Track LLM inputs & outputs"
```
anastasiaguspan: I believe you renamed this quickstart above?

weave/concepts/what-is-weave.mdx
anastasiaguspan: I personally prefer the way you renamed this elsewhere in this PR - to "Instrument and trace functions". More consistent that way as well i think.  (I personally don't like including the word 'application' because it doesn't seem like a helpful disambiguator - a customer has an 'application' regardless of their workflow.  (of course, lol, they have functions too, but this still seems better since a function is so closely tied to weave Op.))
approving either way. but if you do change it here you'll have to update it for consistency in the section that follows "When to use agent tracing and when to use application instrumentation" also.

anastasiaguspan (approved):
very valuable improvements

PR #2909, opened 2026-07-15
fix(weave): rename shadowing OTLP endpoint var in otel tracing docs

gtarpenning:
## Summary
- Rename the local `OTEL_EXPORTER_OTLP_ENDPOINT` variable (which shadows the real OTel SDK env var of the same name) to `WEAVE_OTLP_ENDPOINT` in otel.mdx, agno.mdx, pydantic_ai.mdx.
- Add a callout in otel.mdx distinguishing base-URL vs full-path vs `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` usage, since exporting the old name verbatim and using a zero-arg exporter doubles the path and silently drops spans.
- vercel_ai_sdk.mdx uses an inline string literal, not the shadowing var, so no change needed there.

## Testing
docs-only text change, no test suite.

### DIFF
--- a/weave/guides/tracking/otel.mdx
+++ b/weave/guides/tracking/otel.mdx
@@ -29,6 +29,10 @@ Weave uses the `wandb-api-key` header to authenticate requests and resource attr
 
 The following example shows how to configure authentication and project routing:
 
+<Warning>
+`WEAVE_OTLP_ENDPOINT` below is a local variable, not the OTel SDK's `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable, and it's set to the full traces path (`.../otel/v1/traces`). It's passed explicitly to `endpoint=`/`url:` on the exporter, so the SDK sends requests to that URL verbatim. If you instead `export` a value as the real `OTEL_EXPORTER_OTLP_ENDPOINT` and construct the exporter with no arguments (for example, `OTLPSpanExporter()`), the SDK appends `/v1/traces` itself, so that variable must hold the base URL (`.../otel`) instead, or you get a doubled path (`.../otel/v1/traces/v1/traces`), a `404`, and silently dropped spans. The signal-specific `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` is also used verbatim, like the local variable here.
+</Warning>
+
 <CodeGroup>
 ```python Python lines {7,8}
 import os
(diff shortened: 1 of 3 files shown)


weave/guides/tracking/otel.mdx
```diff
@@ -29,6 +29,10 @@ Weave uses the `wandb-api-key` header to authenticate requests and resource attr
 
 The following example shows how to configure authentication and project routing:
 
+<Warning>
+`WEAVE_OTLP_ENDPOINT` below is a local variable, not the OTel SDK's `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable, and it's set to the full traces path (`.../otel/v1/traces`). It's passed explicitly to `endpoint=`/`url:` on the exporter, so the SDK sends requests to that URL verbatim. If you instead `export` a value as the real `OTEL_EXPORTER_OTLP_ENDPOINT` and construct the exporter with no arguments (for example, `OTLPSpanExporter()`), the SDK appends `/v1/traces` itself, so that variable must hold the base URL (`.../otel`) instead, or you get a doubled path (`.../otel/v1/traces/v1/traces`), a `404`, and silently dropped spans. The signal-specific `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` is also used verbatim, like the local variable here.
+</Warning>
+
 <CodeGroup>
 ```python Python lines {7,8}
 import os
```
anastasiaguspan, line 38:
```suggestion
<Warning>
`WEAVE_OTLP_ENDPOINT` below is a local variable, not the OTel SDK's `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable, and it's set to the full traces path (`.../otel/v1/traces`). It's passed explicitly to `endpoint=`/`url:` on the exporter, so the SDK sends requests to that URL verbatim. If you instead `export` a value as the real `OTEL_EXPORTER_OTLP_ENDPOINT` and construct the exporter with no arguments (for example, `OTLPSpanExporter()`), the SDK appends `/v1/traces` itself, so that variable must hold the base URL (`.../otel`) instead, or you get a doubled path (`.../otel/v1/traces/v1/traces`), a `404`, and silently dropped spans. The same is true of `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`: it also needs the full traces path, since the SDK won't append /v1/traces to it either.
</Warning>

<CodeGroup>
```python Python lines {7,8}
import os
```

anastasiaguspan (approved):

PR #2910, opened 2026-07-15
ci: parallelize Mintlify checks

johndmulhausen:
## Summary

- Run `mint validate` and `mint broken-links` concurrently on separate matrix runners.
- Preserve the branch-protection-required `validate-mdx` check as a terminal gate.
- Use merge-base-aware three-dot diffs for PR scoping and fail closed if the diff cannot be computed.
- Treat deleted, renamed, or validator-code changes as validation triggers.
- Keep the helper script's existing no-argument behavior while allowing CI to run either check independently.
- Restrict the workflow token to read-only repository contents.
- Force base-10 day-of-year arithmetic when generating the Mint CLI cache key.

## Why

Actions run 29436949448 took 4 minutes 24 seconds, including approximately 1 minute 40 seconds for `mint validate` followed by 1 minute 56 seconds for `mint broken-links`.

[The first parallel run](https://github.com/wandb/docs/actions/runs/29438981575) completed in 2 minutes 38 seconds: 1 minute 46 seconds, or 40%, faster. Total runner time increased by about 9% because checkout and Mint setup run once per matrix job.

The original run also exposed a scoping bug. PR #2908 changed only `kapa-widget.js`, but the two-dot base/head comparison included 37 upstream MDX changes because the branch was behind `main`. The three-dot comparison correctly considers only the PR's changes, so non-relevant PRs can skip Mint setup and validation.

This PR also adopts applicable caller safeguards from [coreweave/docs#3370](https://github.com/coreweave/docs/pull/3370). W&B keeps separate matrix runners instead of CoreWeave's background-process approach because they provide independent Mintlify state, non-interleaved logs, and native process cleanup for these longer-running checks.

## Validation

- `actionlint`
- `shellcheck`
- `bash -n scripts/mdx-validation/validate-mdx-mintlify.sh`
- YAML parsing and `git diff --check`
- PR #2908 base/head regression fixture
- mocked per-command dispatch, failure propagation, and legacy `all` behavior
- mocked push and `workflow_dispatch` command paths
- invalid Git diff fails closed
- zero-padded day-of-year cache-key fixtures
- completed GitHub Actions parallel run

### DIFF
--- a/.github/workflows/validate-mdx.yml
+++ b/.github/workflows/validate-mdx.yml
@@ -6,6 +6,9 @@ on:
       - main
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 concurrency:
   group: validate-mdx-${{ github.ref }}
   cancel-in-progress: true
@@ -32,6 +35,8 @@ jobs:
       - name: Determine if Mintlify validation is needed
         id: scope
         run: |
+          set -euo pipefail
+
           if [ "${{ github.event_name }}" != "pull_request" ]; then
             echo "needs_validation=true" >> "$GITHUB_OUTPUT"
             echo "Runs full validation for push/workflow_dispatch (not a PR diff)."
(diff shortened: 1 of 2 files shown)


johndmulhausen:
DF has some thoughts on a better approach here

PR #2912, opened 2026-07-15
docs(weave): add BYOB media reference guides (Weave Op + agent spans)

gtarpenning:
## Summary

Reference media stored in your own cloud bucket by logging the object URI — no bytes are ingested. Weave presigns the URI and renders the image/video inline (by file extension).

This branch splits the guide into **two pages that share one setup snippet**, since the Op-tracing and agent workflows are separate in the TOC:

- **`byob-references`** (Trace your applications → Advanced Ops) — *Reference media in your own bucket (BYOB) using Weave Op*: logging a URI from a `weave.op` function.
- **`agents-byob-references`** (Trace your agents) — *Reference media in your own bucket (BYOB) using agent spans*: returning a URI as an agent tool result so it renders in the Agents view.
- Shared setup extracted to `snippets/_includes/byob-reference-setup.mdx` (intro, how it works, prerequisites, bucket config, registration). Both pages use `sidebarTitle: "Reference media in your own bucket"` with distinct full titles for search.

## Scope / accuracy corrections

- **Multi-tenant Cloud only** — the feature is gated to SaaS today; both pages state this in prerequisites.
- **AWS S3 and GCS only; Azure removed.** The backend `ValidateReferenceBucket` rejects everything but AWS/GCP, so `az://` and the Azure prerequisite were dropped (supported schemes are `s3://`, `gs://`, `gcs://`).
- Bucket-config tabs hardcode the read-only integration principals: `arn:aws:iam::725579432336:role/WandbIntegration` (S3) and `wandb-integration@wandb-production.iam.gserviceaccount.com` with `roles/storage.objectViewer` (GCS), with CLI apply commands for each (`aws s3api put-bucket-policy`/`put-bucket-cors`, `gcloud storage buckets ...`).

## Screenshots

Added real in-app screenshots for both View sections: `imgs/traces-byob.png` (Traces view) and `imgs/agents-byob.png` (Agents view).

## Testing

- **Agent-span code examples verified live** against `wandb/openai-agents-esm`: Python `weave==0.53.2` and TypeScript `weave@0.16.3` both run and emit conversation/turn/tool spans with the bucket URI as the tool result. The agent-span API requires a recent SDK (TS `0.12.x` lacks it entirely).
- Rendering path confirmed in `core` (`ToolCallPayload` → `MaybeReferenceMedia`, `ValueViewReferenceUri`). End-to-end image render not exercised here (needs a registered bucket); without one the URI shows the plain-text fallback.

## Preview

- [Op page](https://wb-21fd5541-griffin-weave-byob-references.mintlify.site/weave/guides/tracking/byob-references)
- [Agents page](https://wb-21fd5541-griffin-weave-byob-references.mintlify.site/weave/guides/tracking/agents-byob-references)


weave/guides/tracking/byob-references.mdx
```diff
@@ -0,0 +1,133 @@
...
+This is a Weave feature and is separate from [platform BYOB (secure storage connector)](/platform/hosting/data-security/secure-storage-connector), which backs W&B artifact and run storage. Reference media buckets are configured per team and only need read access.
+</Note>
+
+## How it works
+
+1. You store a media object in your bucket, for example `s3://your-bucket/frames/000123.png`.
+2. You log that URI as a string value anywhere in a trace: an op input or output, or an agent tool result. The field name does not matter.
+3. When you open the trace, Weave resolves the URI to a short-lived [presigned URL](/platform/hosting/data-security/presigned-urls) and renders it inline. The file extension decides how it renders: images (`.png`, `.jpg`, `.gif`, ...) and videos (`.mp4`, `.mov`, `.webm`, ...) render inline; anything else renders as a download link.
+
+Weave only resolves URIs that point at a bucket registered for your team. Supported schemes are `s3://`, `gs://`, `gcs://`, and `az://`.
```
pingleiwandb: Right now we only support `s3://` and `gs://` object url format. `gcs` and `az` are not supported. We don't support Azure right now and Azure also does not have standard scheme for its object url like AWS/GCP does.

weave/guides/tracking/byob-references.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+## Prerequisites
+
+- A W&B account and [API key](https://wandb.ai/authorize).
+- A cloud storage bucket you control (S3, GCS, or Azure), containing at least one object.
+- Permission to edit your team settings.
```
pingleiwandb: I think now it requires org admin, basically you can only add a reference bucket if you can create a new team in this org, same permission requirement as the Models' BYOB, which is bundled with creating team.

weave/guides/tracking/byob-references.mdx
```diff
@@ -0,0 +1,133 @@
...
+   }
+   ```
+
+2. Apply a CORS configuration so the app can fetch objects in the browser:
+
+   ```json
+   [{
+     "AllowedHeaders": ["*"],
+     "AllowedMethods": ["GET", "HEAD"],
+     "AllowedOrigins": ["*"],
```
pingleiwandb: We can also set this to `"https://wandb.ai"`, though `*` works for other deployment types directly...

weave/guides/tracking/byob-references.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+<Note>
+If an object is encrypted with a customer-managed KMS key, its key policy must also grant the same principal `kms:Decrypt`. The bucket policy alone is not sufficient.
+</Note>
+
+## Register the reference bucket
+
+1. Go to your team settings page (`https://wandb.ai/<your-team>/settings`).
+2. In the **Reference buckets** section, select **Add bucket**.
+3. Enter your bucket URI, for example `s3://your-bucket`, and save.
```
pingleiwandb: Our UI is a bit e ... not intuitive (said the one who wrote/copied the UI),  user cannot just paste objectUrl directly. They need to enter a bucket, select AWS, set bucket name without s3 prefix, set existing object url. Might worth having a screenshot like https://docs.wandb.ai/platform/hosting/monitoring-usage/slack-alerts

anastasiaguspan: @pingleiwandb  Noted, that definitely sounds like it could use more details.
BUT, I don't seem to be an admin on any of my teams, so I can't see this UI - so either i need to be granted admin somewhere OR can someone send me the screenshot of this so I can include it?

Based on your details, sounds like we want the procedure to read like (confirm):
▎ 1. Go to your team settings page (https://wandb.ai/[YOUR-TEAM]/settings).
▎ 2. In the Reference buckets section, select Add bucket.
▎ 3. Select your cloud provider (AWS or Google Cloud).
▎ 4. For the bucket name, enter [YOUR-BUCKET] only — without the s3:// or gs:// prefix.
▎ 5. Enter the path to an existing object in the bucket. 
▎ 6. Save.

Questions:
A. Is access confirmed at point of save; or is nothing really confirmed on Save.?
B. You are setting an existing *object* url even though this is setting up the bucket?  so any object unlocks access to all objects in that bucket?  Having the exact field name in the UI entry would also help to add.

weave/guides/tracking/byob-references.mdx
anastasiaguspan: @pingleiwandb  ....'other deployment types' ... tho i thought we said this is multi-tenant cloud only (at least currently)
thus, almost sounds like I should replace the * directly in the code snip and use the wandb.ai?

anastasiaguspan:
@gtarpenning  and @pingleiwandb 

Updates for your review ->  Most changes were wording/style updates, but these are worth your review:
1. Addressed the ‘mtsaas’-only by adding a prerequisite  that this is mutti-tenant cloud
2. Removed azure, gcs
3. Added a tab to handle gcloud’s counterpart. Please confirm details.
4. Griffin: Biggest change - I had to split the pages to handle the 'agents' workflow vs the older 'weave.op' workflow because these workflows are segregated in the navigation.  For the Agents workflow, that link out to the quickstart i didn't feel clarified the 'how to' for customers, so i created dedicated Python/TS code examples for this workflow.
-All the top bucket setup stuff is SHARED (one source) between the two docs pages so you only need to QA one, the unique content starts at 'Log a media reference...' Weave code.
-I didn't test the bucket config stuff since that came from you I presumed it was accurate. i did though test the *mechanics* of the new Weave agents code I added this round (not the render tho w a real bucket, just the code/span entry).

weave/guides/tracking/byob-references.mdx
pingleiwandb: > A. Is access confirmed at point of save; or is nothing really confirmed on Save.?

Yeah, it is confirmed on save. There is a `Test connection` button to test if the bucket work without saving it. The `Add bucket` button in dialog persist the config.

> B. You are setting an existing object url even though this is setting up the bucket?

Yes, only one existing object key (not url) is required. We use it to test user has granted the right permission and set right CORS setting. User can limit the permission to a subset of objects in their bucket by setting bucket permission with more granularity, but there is no good way for us to inspect that (unless we require more permissions). So we basically assume user configured things properly instead of granting access to a single object.

Here is a screenshot, user need to be admin of org (not just team) to have permission to add a bucket.

Screenshot 2026-07-17 at 09 32 29

weave/guides/tracking/byob-references.mdx
pingleiwandb: Yeah, use `https://wandb.ai` is better. For other deployment types e.g. dedicated, they need to use their own domain.

anastasiaguspan (approved):

pingleiwandb (approved):
LGTM

PR #2918, opened 2026-07-16
Registry: Link prompts to registry 

ngrayluna:
## Description
Adds doc on how to link a prompt to a collection/registry.

## Related issues

- Fixes https://coreweave.atlassian.net/browse/DOCS-2797
- StringPrompt docstring: https://github.com/wandb/weave/pull/7583


models/registry/link_prompt.mdx
```diff
@@ -0,0 +1,110 @@
+---
+title: Link a prompt to a collection
```
ruhiparvatam: **suggestion:** I think it would be helpful to make it clear that "prompt" is a Weave asset type. Maybe we can update this title to be something like: `Link a W&B Weave prompt to a collection`. 

And that way we can also add a link to the [weave prompts page](https://docs.wandb.ai/weave/guides/core-types/prompts). Wdyt?

models/registry/link_prompt.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+## View linked prompts in a registry
+
+1. Navigate to the W&B Registry.
+2. Select the name of the registry that you linked the prompt to.
+3. Select the name of the collection.
+4. From the list of prompt versions, select the version you want to access. Version numbers are incrementally assigned to each linked artifact version starting with `v0`.
+5. To view details about a prompt version, click the version. From the tabs in this page, you can view that version's overview, prompt, and use.
+
+## Use a prompt linked to a registry
```
ruhiparvatam, line 64: Out of scope for this PR but I see that we have a section for `Use Weave with W&B Models` - could be nice to add this as an example in https://docs.wandb.ai/weave/cookbooks/Models_and_Weave_Integration_Demo#save-the-new-rag-model-to-the-registry ?

ruhiparvatam:
Another suggestion of a place to mention this integration: 
- Add a link to this page: https://docs.wandb.ai/weave/guides/core-types/prompts-version that takes you to the page created by this PR?

models/registry/link_prompt.mdx
ngrayluna: I left the title the same (for now), but I did add (throughout this doc that you need to create a weave prompt objects. I also added this tip, link to the Weave docs:

Image

Reasoning for keeping the title the same: as a user I am interested in tracking my prompt. The user may not know what a weave prompt object is, so the title might be confusing.

Let me know what ya think. I am not wedded to it so I can change the title if I'm outvoted :)

ruhiparvatam: yeah I see your perspective! but I'd argue that to create a prompt you'll already have some knowledge of the weave ecosystem, because you'll have a weave project and call the weave API. Maybe @aninahpets can be our tiebreaker :)

Thanks for making these changes! The page looks great!

ruhiparvatam (approved):

models/registry/link_prompt.mdx
ngrayluna: Nina has spoken!

models/registry/link_prompt.mdx
```diff
@@ -0,0 +1,113 @@
+---
+title: Link a Weave prompt to a collection
+---
+
+Track, version, and manage your AI prompts in the W&B Registry by linking Weave prompt objects to a collection. *Linking* a prompt object to a [collection](/models/registry/create_collection) moves the prompt version from a [private, project-level scope to a shared, organization-level scope](/models/registry/create_registry#visibility-types).
+
+<Note>
```
thedriftofwords, line 7: nit: "Before you start" or "Prerequisites" H2 heading rather than a note callout

thedriftofwords (approved):
approving with a non-blocking style nit, up to your discretion

PR #2923, opened 2026-07-17
Send generic OTel spans to Weave Agents view

dbrian57:
## Description

This guide demonstrates how to send OTel spans to Weave's Agent view without installing the Weave SDK.

PR #2930, opened 2026-07-17
docs(weave): document linked agent eval traces

bcsherma:
## Summary

- Documents that Python agent calls must run inside the `EvaluationLogger.log_prediction()` context to link their spans to evaluation results.
- Adds a focused OpenAI Agents SDK example showing the context, output, score, and example ID.
- Documents every raw `weave.eval.*` OpenTelemetry attribute, including the two call IDs required for evaluation and result linkage.
- Adds Python and TypeScript examples for manually stamped spans and explains how to retrieve the linkage IDs from the evaluation results API.
- Adds a distributed Python example that creates a `predict_and_score` call, propagates both call IDs to an external agent service, and stamps the receiving OTel span.
- Clarifies that automatic evaluation-to-agent span linking is Python-only and that TypeScript requires manual OTel attributes with both call IDs.
- Shows how to open evaluation-linked spans in the Weave UI.

## Validation

- `npx mint@latest validate`
- The new OpenAI Agents SDK, Agents OTel endpoint, and evaluation results API links resolve.


weave/guides/evaluation/evaluation_logger.mdx
```diff
@@ -370,7 +370,106 @@ While TypeScript doesn't have automatic cleanup with context managers, `logSumma
...
+#### Set evaluation attributes on raw spans
+
+You don't need to set evaluation attributes on spans created inside the `log_prediction()` context. `EvaluationLogger` sets them automatically. If you send spans with your own OpenTelemetry instrumentation, or create spans in another process, set the attributes directly on each span that you want to link. OpenTelemetry span attributes do not propagate from parent spans to child spans.
+
+The span must be sent to the same Weave project as the evaluation. For information about configuring the destination project, see [Send OpenTelemetry traces to Weave](/weave/guides/tracking/otel).
+
+| Attribute | Type | Description |
+| --- | --- | --- |
+| `weave.eval.run_id` | string | The call ID of the evaluation run (`Evaluation.evaluate`). Required to include the span in the evaluation-level **View spans** result. |
+| `weave.eval.predict_and_score_call_id` | string | The call ID of the `Evaluation.predict_and_score` operation for a specific result and trial. Set this together with `weave.eval.run_id` to link the span to that result. |
```
NiWaRe: Maybe clearer if we include a quick example of how to create and propagate both IDs in above example? 

Should be very similar but just to make sure the creation and propagation of the `predict_and_score` is also clear - from my understanding it is auto-created when a new prediction is logged and you would pass it along the external agent invocation?

NiWaRe (commented):
Minor feedback to make sure everyone understands - great otherwise thanks for the fast turn around!!

NiWaRe (approved):
Great example now!

weave/guides/evaluation/evaluation_logger.mdx
```diff
@@ -370,7 +370,187 @@ While TypeScript doesn't have automatic cleanup with context managers, `logSumma
...
+eval_logger.log_summary()
+```
+
+If the agent runs before the prediction context starts or after it finishes, Weave records the trace but does not link it to the evaluation result.
+
+#### Set evaluation attributes on raw spans
+
+You don't need to set evaluation attributes on spans created inside the `log_prediction()` context. `EvaluationLogger` sets them automatically. If you send spans with your own OpenTelemetry (OTel) instrumentation, or create spans in another process, set the attributes directly on each span that you want to link. OTel span attributes do not propagate from parent spans to child spans.
+
+The span must be sent to the same Weave project as the evaluation. For information about configuring the destination project, see [Send OpenTelemetry traces to Weave](/weave/guides/tracking/otel).
```
anastasiaguspan: we want to point to agents endpoint, not the old v1 stuff.  so we'll need to point to Dan's new agents otel page after it gets merged.

weave/guides/evaluation/evaluation_logger.mdx
```diff
@@ -370,7 +370,187 @@ While TypeScript doesn't have automatic cleanup with context managers, `logSumma
...
+        return {"output": str(result.final_output or "")}
+```
+
+The wrapper span in this example is linked to the evaluation result. If the agent framework creates additional spans, copy `eval_context` onto those spans as well; OTel does not inherit span attributes from the wrapper.
+
+To inspect linked traces in the UI:
+
+1. Open the evaluation run in **Evals**.
+2. In **Results**, click **View spans** to open **Agents > Spans** filtered to that evaluation.
+3. Open an example preview and click **View spans** to inspect the agent spans linked to that example.
```
anastasiaguspan: @bcsherma I don't understand where you're trying to go with Step 3?  step 2 already opens the Spans for the evaluation....
what is "an example preview" ?   When step 2 finishes i have spans on the spans table to select, and even if i open the span detail panel for one of them.... i can find nothing that says 'view spans' again

weave/guides/evaluation/evaluation_logger.mdx
bcsherma: Updated this to the canonical /agents/otel/v1/traces API reference so it no longer points at legacy raw OTel ingestion. Dan's guide is #2923 and is still open, so I used the stable endpoint reference for this PR and can swap to the guide after it merges.

weave/guides/evaluation/evaluation_logger.mdx
bcsherma: Good catch — step 3 described an incorrect extra flow. I removed it; step 2 is the complete evaluation-level path.

anastasiaguspan (approved):

PR #2948, opened 2026-07-21
Embed live W&B Reports (regular reports, not Fully Connected articles) in docs pages

johndmulhausen:
## Description

Adds a reusable way to embed live, interactive W&B Reports in docs pages instead of static screenshots, plus the guardrails to keep those embeds from silently rotting.

> [!IMPORTANT]
> **Works with regular W&B reports only, not Fully Connected blog articles.** A regular report shared publicly (or via a view-only link) renders in a slim embed view. A Fully Connected article renders its full blog chrome — the black nav bar, the "W&B Fully Connected › Articles" breadcrumb, author, stars — regardless of the embed parameter, so it looks broken in a frame. Embed regular reports; don't embed FC articles.

**Component**
- `snippets/WandbReport.jsx` — stateless component that renders a report in a lazy-loaded iframe inside a themed card. It requests W&B's slim embed view (`?jupyter=true`) for the frame while keeping the caption link on the normal report URL. Colors come from the existing `--color-*` CSS variables, so the card chrome tracks dark mode with **no JS and no changes to the generated `css/styles.css`**. It validates `src`, degrades to a link-only card when framing is unavailable, and always renders a caption link.

**Registry + validation (`scripts/report-embeds/`)**
- `registry.yaml` — one entry per embedded report (id, URL, owner, purpose, pages, last-reviewed).
- `check_embeds.py` — enforces that every `<WandbReport>` is registered, that the page also carries a **plain prose link** to the report (agents, the llms.txt export, and the translation pipeline read MDX source, where the iframe is opaque), and — in liveness mode — that each URL still returns an anonymous `200` that isn't a login redirect. `wandb.ai` is excluded from the site-wide lychee check and lychee accepts `403/429`, so embeds need their own check. Includes unit tests.
- `.github/workflows/report-embeds.yml` — static checks on PRs; weekly liveness that files an issue on failure. Network results never block PRs.

**Author guidance**
- `AGENTS.md` — house rule covering the prose-link requirement, registration, anonymous-viewable/public-forever, and the 1–2-per-page cap.
- `models/reports/embed-reports.mdx` — the "public reports only" note now also covers view-only (magic) links.

**Pilot embeds** — two existing public regular reports on topically matching pages:
- `models/reports.mdx` (Reports overview) — SafeLife benchmark report, beside the static screenshot it already linked.
- `models/track/log/media.mdx` (`## 3D visualizations`) — LIDAR point clouds report.

(A third pilot on `run-comparer.mdx` was dropped: that report is a Fully Connected article and rendered the full blog chrome — see the limitation above.)

## Testing
- [x] `mint validate` succeeds (compiles `WandbReport.jsx`, imported by the embed pages)
- [x] `mint broken-links` succeeds
- [x] `python3 scripts/report-embeds/check_embeds.py --mode all` passes (static + anonymous liveness on both URLs)
- [x] `python3 -m unittest discover -s scripts/report-embeds/tests` (20 tests)
- [x] Slim embed view confirmed rendering in the Mintlify preview for both pages

## Reviewer notes
- The pilot embeds are `stacey/*` public regular reports. They're article-length but render in the slim embed view; the `AGENTS.md` rule still says to *prefer* skinny purpose-built reports, and swapping in trimmed versions later is just a URL change in the registry.
- This touches two content pages plus the mechanism. Happy to split into a mechanism PR and a content-embeds PR if preferred.
- Out of scope (follow-ups): a `?theme=`/postMessage-height feature ask to the app team (and getting `?jupyter=true` honored on more report surfaces), and authoring embeds programmatically via the Reports API.


models/app/features/panels/run-comparer.mdx
```diff
@@ -28,3 +30,13 @@ By default, Run Comparer doesn't differentiate runs with different values for [`
 
 When you search, filter, group, or sort the list of runs in the Runs Table, the Run Comparer automatically updates to compare the first 10 runs. Filter or search within the Runs Table to compare similar runs, such as by filtering or sorting the list by `job_type`. Learn more about [filtering runs](/models/runs/filter-runs/).
 </Note>
+
+## Example
+
+Reports capture a run comparison so you can share the conclusion alongside the supporting charts. The [When Inception-ResNet-V2 Is Too Slow report](https://wandb.ai/stacey/estuary/reports/When-Inception-ResNet-V2-is-too-slow--Vmlldzo3MDcxMA) embedded below compares model variants to show which architectures train faster; scroll within it to read the analysis and inspect the panels.
```
ngrayluna: Semicolons reduce readability for ESL folks.

models/app/features/panels/run-comparer.mdx
```diff
@@ -28,3 +30,13 @@ By default, Run Comparer doesn't differentiate runs with different values for [`
...
+
+<WandbReport
```
ngrayluna: Not sure if this is a good example. Someone skimming this might be off put by this UI that looks like an error:

Image

models/track/log/media.mdx
```diff
@@ -320,6 +321,14 @@ If histograms are in your summary they will appear on the Overview tab of the [R
 
 Log 3D point clouds and Lidar scenes with bounding boxes. Pass in a NumPy array containing coordinates and colors for the points to render. 
 
+Logged point clouds are fully interactive in the W&B app. The [LIDAR Point Clouds of Driving Scenes report](https://wandb.ai/stacey/lyft/reports/LIDAR-Point-Clouds-of-Driving-Scenes--Vmlldzo2MzA5Mg) embedded below shows scenes from the Lyft dataset with 3D bounding boxes — drag to rotate and zoom within the viewer.
```
ngrayluna: Em dashes also reduce readability for ESL folks (applies to other doc as well).

models/app/features/panels/run-comparer.mdx
johndmulhausen: Good catch. Removed the semicolon in c0a8708b6 by splitting it into two sentences.

scripts/report-embeds/check_embeds.py
```diff
@@ -0,0 +1,377 @@
...
+    findings: list[Finding] = []
+    try:
+        raw = yaml.safe_load(path.read_text(encoding="utf-8"))
+    except FileNotFoundError:
+        return [], [Finding("error", "REGISTRY_MISSING", f"Registry not found: {path}")]
+    except yaml.YAMLError as exc:
+        return [], [Finding("error", "REGISTRY_UNREADABLE", f"Invalid YAML in {path}: {exc}")]
+
+    if raw is None or raw.get("reports") is None:
+        return [], [Finding("error", "REGISTRY_SHAPE", "Registry must have a top-level `reports:` list")]
```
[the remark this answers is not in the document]
johndmulhausen: Fixed in c0a8708b6. load_registry now checks isinstance(raw, dict) first, so a top-level list or string returns a REGISTRY_SHAPE finding instead of raising AttributeError. Added a test covering the list and string cases.

models/track/log/media.mdx
johndmulhausen: Fixed in c0a8708b6. Removed the em dashes from the added prose here and on the other two embed pages (reports.mdx and run-comparer.mdx).

models/app/features/panels/run-comparer.mdx
johndmulhausen: Agreed that it reads poorly. What shows here is the full W&B report page: the app nav, the Fully Connected article header, and the cookie banner. That chrome comes with any embedded report, and because the iframe is cross-origin the docs site cannot style or trim it. Two things are in flight: the component now requests W&B's notebook embed view (jupyter=true), which the app team would need to honor for it to strip the chrome, and the caption link under the frame is the always-visible fallback. For this page in particular we are leaning toward dropping this example or swapping in a purpose-built single-panel report rather than a full article. Will follow up.

johndmulhausen: Removed in 950dbe08d. Root cause: this report is published to Fully Connected, and FC articles render the full blog chrome regardless of the jupyter=true embed param (unlike the other two, which are regular reports and now render slim). Rather than swap in another report here, we dropped the example, since the Run Comparer page is about the panel itself. The two remaining embeds (Reports overview and Log media) render cleanly.

johndmulhausen: And, removed. Turns out this really only works as intended with regular W&B Reports, not FC articles.

models/reports/embed-reports.mdx
```diff
@@ -13,7 +13,7 @@ Use an HTML `iframe` element to embed a report on any web page that accepts cust
 In the upper right of a report, select the **Share** button to open a dialog, then select **Copy embed code**. The copied snippet is an `iframe` HTML element that you can paste into any page that accepts custom HTML.
 
 <Note>
-Only *public* reports are viewable when embedded.
+Only reports that anonymous viewers can open render when embedded: reports in *public* projects, or reports shared with a [view-only link](/models/reports/cross-project-reports#view-only-report-links) (the link's access token is part of the URL).
```
dbrian57:
```suggestion
You can only embed and render reports from public projects or reports that you've shared using [view-only links](/models/reports/cross-project-reports#view-only-report-links) (the link's access token is part of the URL).
```
More active and direct

models/track/log/media.mdx
```diff
@@ -320,6 +321,14 @@ If histograms are in your summary they will appear on the Overview tab of the [R
 
 Log 3D point clouds and Lidar scenes with bounding boxes. Pass in a NumPy array containing coordinates and colors for the points to render. 
 
+Logged point clouds are fully interactive in the W&B app. The [LIDAR Point Clouds of Driving Scenes report](https://wandb.ai/stacey/lyft/reports/LIDAR-Point-Clouds-of-Driving-Scenes--Vmlldzo2MzA5Mg) embedded below shows scenes from the Lyft dataset with 3D bounding boxes. Drag to rotate and zoom within the viewer.
```
dbrian57:
```suggestion
Logged point clouds are fully interactive in the W&B app. For example, the [LIDAR Point Clouds of Driving Scenes report](https://wandb.ai/stacey/lyft/reports/LIDAR-Point-Clouds-of-Driving-Scenes--Vmlldzo2MzA5Mg) embedded below shows scenes from the Lyft dataset with 3D bounding boxes. Drag to rotate and zoom within the viewer.
```

This is a cool report, but it's also from 2021... Do we have anything more current?

models/reports.mdx
```diff
@@ -24,7 +25,13 @@ The following image shows a section of a report created from metrics that were l
     <img src="/images/reports/safe-lite-benchmark-with-comments.png" alt="W&B report with benchmark results" max-width="90%"  />
 </Frame>
 
-View the report where the above image was taken from [here](https://wandb.ai/stacey/saferlife/reports/SafeLife-Benchmark-Experiments--Vmlldzo0NjE4MzM).
+View the report where the above image was taken from [here](https://wandb.ai/stacey/saferlife/reports/SafeLife-Benchmark-Experiments--Vmlldzo0NjE4MzM). The same report is embedded live below. Scroll within it to explore the benchmark runs, panels, and commentary as readers see them in the W&B app.
```
dbrian57: Do we need the image above if the report is embedded? I think we can reword lines 22-28 to introduce the report and explain what readers can do with it.

dbrian57 (approved):
Some suggestions inline.

models/reports/embed-reports.mdx
johndmulhausen: Applied in ef9d360bc. Thanks, more direct.

models/track/log/media.mdx
johndmulhausen: Applied the "For example," lead-in in ef9d360bc. On the age: agreed it's from 2021. I picked it because it's a public, non-Fully-Connected report that renders slim in the frame and matches this section's topic (3D point clouds and Lidar with bounding boxes). I don't have a newer public one that fits all three. If you know a more current 3D/point-cloud report that's public or view-only shared, swapping it is a one-line change to the URL in scripts/report-embeds/registry.yaml plus the two references on this page. Happy to swap.

models/reports.mdx
johndmulhausen: Good call, done in ef9d360bc. Removed the static screenshot (the live embed shows the same report) and reworded the intro to introduce the report and what readers can do with it, keeping the plain link the CI check requires.

PR #2985, opened 2026-07-29
Updates Weave's landing page with new auto-instrumentation skill

dbrian57:
## Description
Resolves DOCS-2949. This updates Weave's landing page with the product's new onboarding workflow that uses a prompt to instruct an agent on how to install Weave into a codebase.

### DIFF
--- a/weave.mdx
+++ b/weave.mdx
@@ -73,7 +73,7 @@ Select your workload type and then use the following docs to guide you through t
 </Tab>
 </Tabs>
 
-## Automatically set up Weave
+## Automatically instrument Weave
 
 To automatically instrument your agents to use Weave, update the following prompt with your project name and [W&B API key](https://wandb.ai/settings) environment variables, then pass the prompt to your agent:
 
--- a/weave.mdx
+++ b/weave.mdx
@@ -73,21 +73,27 @@ Select your workload type and then use the following docs to guide you through t
 </Tab>
 </Tabs>
 
-## Install Weave
+## Automatically instrument agents with Weave
 
-W&B Weave provides Python and TypeScript libraries. To install the Weave library, run the following command:
+To automatically instrument your agents to use Weave, update the following prompt with your project name and [W&B API key](https://wandb.ai/settings) environment variables, then pass the prompt to your agent:
 
-<Tabs>
-  <Tab title="Python">
-    ```bash
-    pip install weave
-    ```
-  </Tab>
-  <Tab title="TypeScript">
-    ```bash
-    npm install weave
-    ```
-  </Tab>
-</Tabs>
+```text
+Fetch and then use the following skill to instrument all agents with Weave:
 
-To start using the Weave library, create a [Weights & Biases (W&B) account](https://wandb.ai) and an [API key at User Settings](https://wandb.ai/settings). The API key allows you to authenticate to your W&B account and start sending data to it.
+https://raw.githubusercontent.com/wandb/weave/refs/heads/master/skills/weave-instrument/SKILL.md
+
+Project configuration:
+WEAVE_PROJECT=$[YOUR-PROJECT-NAME]
+WEAVE_API_KEY=$[WANDB-API-KEY]`;
+```
+
+Your agent installs Weave's auto-instrumentation skill and uses it to:
+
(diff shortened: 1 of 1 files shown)


weave.mdx
```diff
@@ -73,21 +73,27 @@ Select your workload type and then use the following docs to guide you through t
...
-  </Tab>
-</Tabs>
+```text
+Fetch and then use the following skill to instrument all agents with Weave:
+
+https://raw.githubusercontent.com/wandb/weave/refs/heads/master/skills/weave-instrument/SKILL.md
+
+Project configuration:
+WEAVE_PROJECT=$[YOUR-PROJECT-NAME]
+WEAVE_API_KEY=$[WANDB-API-KEY]`;
```
anastasiaguspan, line 87: I wouldn't think it'd be a good idea to pass an API key visibly through a prompt to an LLM?
What about if we told people to set these in their zshrc file instead, and then the prompt could extract it from there?

dbrian57: I think that's what the `$` symbol is kind of denoting here, right? Like `echo $WANDB_API_KEY`? And we mention that you should set these as env variables in the opening paragraph. Is there a better way here?

anastasiaguspan: oh, i thought that was *setting* an environment variable to the value, and that we were asking them to replace [WANDB-API-KEY] (since that's our replaceable format style) with the text of the key?

would we want the brackets on there if it is supposed to be *reading* from an en var that exists by that name?

also, confused why we're setting WEAVE_API_KEY? i checked the SKILL and it doesn't have this token

anastasiaguspan: like, the skill has an auth-related section, but says nothing about WEAVE_API_KEY:
- **Authentication is the user's job, not yours, so never handle their key.** A W&B API key is a
  secret, and it must not enter your context or the repo. Tell the user to authenticate in their own
  terminal: set `WANDB_API_KEY` (from https://wandb.ai/authorize), or run `wandb login` (Python) or
  `await weave.login(key)` once (Node). You may check *that* a key is reachable (the env var is set,
  or `~/.netrc` has an `api.wandb.ai` entry), but never read or print its value, and never write it
  into source code or into a committed `.env`.

anastasiaguspan: i interpret that as these are the instructions we'd give the user up top:: "authenticate in your own
terminal: set WANDB_API_KEY (from https://wandb.ai/authorize), or run wandb login (Python) or
await weave.login(key) once (Node). "
and then that line gets removed from the prompt altogether?

anastasiaguspan: WEAVE_PROJECT doesn't seem to be used by the skill either?  (tho sounds like the skill will prompt for it itself)
maybe i'm just confused ;)

weave.mdx
```diff
@@ -73,21 +73,27 @@ Select your workload type and then use the following docs to guide you through t
...
-  <Tab title="TypeScript">
-    ```bash
-    npm install weave
-    ```
-  </Tab>
-</Tabs>
+```text
+Fetch and then use the following skill to instrument all agents with Weave:
+
+https://raw.githubusercontent.com/wandb/weave/refs/heads/master/skills/weave-instrument/SKILL.md
```
anastasiaguspan, line 83: also, curious why we're providing this 'raw' link instead of the actual repo https://github.com/wandb/weave/blob/master/skills/weave-instrument/SKILL.md?

weave.mdx
```diff
@@ -73,21 +73,27 @@ Select your workload type and then use the following docs to guide you through t
 </Tab>
 </Tabs>
 
-## Install Weave
+## Automatically set up Weave
```
anastasiaguspan:
```suggestion
## Automatically instrument Weave
```
seems like this skill does far more than 'set up' - i think of setup as being library install, auth, maybe so far as init.
Actually, i'd go as far as maybe 'Automatically instrument Weave into your existing application'

weave.mdx
dbrian57: I think the idea is to make it as easy as possible for the agent to ingest. This is Andrew's prompt that he designed and put in the Weave UI, so I was just copying the workflow he designed.

weave.mdx
dbrian57: this was the original title 😆

PR #3006, opened 2026-08-04
add compact option

ngrayluna:


ngrayluna:
Added in: https://github.com/wandb/docs/pull/3164

PR #3012, opened 2026-08-04
Runs: add support article on env var for ephemeral env

ngrayluna:
## Description

Adds support article on how to set env var to persist, capture `/wandb` files in an ephemeral environment. 

## Related issues

- Fixes https://coreweave.atlassian.net/browse/DOCS-2702


mdlinville (approved):
LGTM, I'd prefer a nicer title before merge. :)

PR #3017, opened 2026-08-04
docs: Add how-to for deleting artifacts in bulk

anastasiaguspan:
## What

Adds a new how-to guide, **Delete artifacts in bulk** (`models/artifacts/bulk-delete-artifacts.mdx`), for readers who need to remove a large volume of artifacts, typically to get back under a storage limit.

The core guidance is scope selection: delete at the coarsest scope that is safe for your data (project, then collection, then run + artifacts, then individual version), because one server-side deletion replaces potentially millions of API calls.

Sections:

- Find what uses your storage (Manage storage page, Usage & Alerts, and a per-collection inventory script)
- Choose the largest safe scope (scope/tool/when-to-use table)
- Delete a project
- Delete artifact collections
- Delete artifact versions selectively (two-pass select-then-delete with a resumable log)
- Expire artifacts in bulk with a TTL policy
- Pace deletions to stay within limits
- Troubleshoot deletion errors
- Confirm storage was reclaimed
- Prevent storage from growing again

## Why

`models/artifacts/delete-artifacts.mdx` explains per-version deletion and garbage collection, but nothing in the docs addresses the scale case. Support has handled this repeatedly, including a user with 5.2 TB and roughly two million artifacts who spent weeks deleting one at a time through the API and the UI before support suggested deleting the project instead. WB-37221 ("Inefficiencies querying artifacts", Done) asked for exactly these documented patterns: bulk listing, filtering by type/age/size, deletion patterns, system-managed artifact limitations, and storage accounting.

## Also changed

- `docs.json`: adds the page to **Models > Artifacts > Manage data**, after Delete an artifact.
- `models/artifacts/delete-artifacts.mdx`: tip pointing to the new page from "Delete multiple artifact versions".
- `platform/app/settings-page/storage.mdx`: adds the new page to the "Delete data" options.
- `platform/hosting/managing-bucket-storage.mdx`: adds the new page to "Reduce bucket usage".

English only. Localized copies under `ja/`, `ko/`, and `fr/` are untouched, per `AGENTS.md`.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -346,6 +346,7 @@
                             "group": "Manage data",
                             "pages": [
                               "models/artifacts/delete-artifacts",
+                              "models/artifacts/bulk-delete-artifacts",
                               "models/artifacts/ttl",
                               "models/artifacts/storage"
                             ]
--- a/models/artifacts/bulk-delete-artifacts.mdx
+++ b/models/artifacts/bulk-delete-artifacts.mdx
@@ -1,0 +1,261 @@
+---
+title: Delete artifacts in bulk
+description: Delete many artifacts at once to reclaim storage and get back under your storage limit.
+keywords: ["bulk delete artifacts", "storage limit", "reduce storage", "delete many artifacts", "reclaim storage"]
+---
+
+Delete many artifacts at once when you need to reclaim storage, for example to get back under your storage limit. This page describes how to find what consumes your storage, how to choose a deletion method that scales to thousands or millions of artifacts, and how to confirm that W&B reclaimed the space.
+
+Deleting one artifact version at a time works for a handful of artifacts, but it doesn't scale. Each call to [`wandb.Artifact.delete()`](/models/ref/python/experiments/artifact#method-artifact-delete) is a separate API request, so clearing a million versions means a million requests. Instead, delete at the largest scope that is safe for your data: a single project or collection deletion removes all of the artifacts it contains in one server-side operation.
+
+<Warning>
+Deleted artifacts can't be recovered after garbage collection completes. Download or back up anything you might need before you delete it. See [Download and use an artifact](/models/artifacts/download-and-use-an-artifact).
+</Warning>
+
+## Before you begin
+
+Before you delete artifacts in bulk, confirm the following:
+
+- You use a current release of the W&B Python SDK. If deletions don't appear in the W&B App as expected, upgrade the SDK and retry.
+- You know which artifacts you must keep. Bulk deletion isn't reversible.
+- For team-wide time-to-live (TTL) policies, you are a team admin. Only team admins can set a team's default TTL policy.
+- For artifacts with [protected aliases](/models/registry/aliases#protected-aliases), you are a registry admin. Other users can't delete these artifacts.
+
+## Find what uses your storage
+
+Target the largest consumers first. A single project that logs model checkpoints or rendered media every run usually accounts for most of the total.
+
+To review usage in the W&B App, use either of the following pages:
+
+- The **Manage storage** page at `https://wandb.ai/account-settings/[ENTITY]/usage/manage/`, where `[ENTITY]` is your team or username. This page shows which projects and runs consume the most space.
+- The **Usage & Alerts** page in organization settings, where organization admins can select the **Storage** category to see usage over time. See [Billing settings](/platform/app/settings-page/billing-settings#view-usage-over-time).
+
+To inventory a project programmatically, list its artifact collections and report the size and age of each version. Replace `[ENTITY]` and `[PROJECT]` with your own values:
+
+```python
+import wandb
+
+api = wandb.Api()
+project = api.project(name="[PROJECT]", entity="[ENTITY]")
+
+for collection in project.collections():
+    total_bytes = 0
+    count = 0
+    newest = None
+    for version in collection.artifacts():
+        total_bytes += version.size
+        count += 1
+        if newest is None or version.created_at > newest:
+            newest = version.created_at
+    print(
+        f"{collection.type}/{collection.name}: "
+        f"{count} versions, {total_bytes / 1e9:.2f} GB, last logged {newest}"
+    )
+```
+
+You should see output similar to the following:
+
+```text
+model/checkpoints: 812 versions, 431.60 GB, last logged 2025-11-04T18:22:07
+run_table/predictions: 5104 versions, 88.19 GB, last logged 2025-11-04T18:24:51
+dataset/train-split: 3 versions, 12.04 GB, last logged 2025-06-12T09:03:44
+```
+
+W&B deduplicates files across versions of the same artifact, so the space you reclaim is smaller than the total size the versions report. Deleting a version frees only the files that no other version references. For an example, see [How much storage does each artifact version use?](/support/models/articles/how-much-storage-does-each-artifact-vers).
+
+## Choose the largest safe scope
+
+Pick the coarsest scope that contains only data you are willing to lose. Coarser scopes delete far more artifacts per API request, which makes them faster and less likely to fail partway through.
+
+| Scope | How to delete | When to use it |
+| ----- | ------------- | -------------- |
+| Project | [Delete the project](#delete-a-project) in the W&B App | An experiment or scratch project is finished and you no longer need any of its runs, artifacts, or history. |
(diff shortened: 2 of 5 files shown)


anastasiaguspan:
## Sources and decision log

### Where the facts came from

**SDK reference (authoritative, generated from the SDK)**

- `Artifact.delete(delete_aliases=False)` raising when aliases exist, and deleting only the link for a linked artifact: `models/ref/python/experiments/artifact.mdx`, confirmed against [`wandb/sdk/artifacts/artifact.py`](https://github.com/wandb/wandb/blob/main/wandb/sdk/artifacts/artifact.py).
- `Artifact.created_at` returning an ISO 8601 **string** (not a `datetime`): verified in the SDK source, which is why the example parses it with `datetime.fromisoformat()`.
- `ArtifactCollection.delete()` deleting the entire collection: `models/ref/python/public-api/artifactcollection.mdx`.
- `Project.collections()`, `Api.project()`, `Api.artifacts(type_name, name, tags, per_page, start)` and the `cursor`/`start` resume pattern: `models/ref/python/public-api/{project,api}.mdx`.
- `Run.delete(delete_artifacts=False)`: `models/ref/python/public-api/run.mdx`.
- `artifact.size` in bytes and `qualified_name`: `models/ref/python/experiments/artifact.mdx`.

**Existing docs (reused rather than restated)**

- Soft delete, garbage collection, protected aliases, and GC prerequisites per deployment: `models/artifacts/delete-artifacts.mdx`.
- Team default TTL applying to existing artifacts by creation date, and the auto-generated types (`run_table`, `code`, `job`, `wandb-*`) that can't take a TTL: `models/artifacts/ttl.mdx`.
- Project deletion UI steps and the sweeps-keep-running caveat: `models/track/project-page.mdx`.
- Storage meter lag, the `account-settings/[ENTITY]/usage/manage/` URL, and automatic lifting of access restrictions: `support/models/articles/why-does-the-storage-meter-not-update-af.mdx`.
- Deduplication accounting (deleting a version frees only files unique to it): `support/models/articles/how-much-storage-does-each-artifact-vers.mdx`.
- `time.sleep()` between calls to avoid `429`: `support/models/articles/how-do-i-paginate-through-large-api-results-in-wandb.mdx`.
- Usage & Alerts / Storage category: `platform/app/settings-page/billing-settings.mdx`.

**Internal context (shaped the guidance, not quoted)**

- Slack `#billing-support-wandb`, Dec 2025 to Jan 2026: user with 5.2 TB and roughly 2M artifacts, deleting one at a time through the API and portal, hitting "cannot delete artifact because alias exists" and a portal error. Support's resolution after the user backed up their data was **delete the project**. This is why the page leads with scope selection instead of a deletion loop.
- WB-37221 "Inefficiencies querying artifacts" (Done): asks for documented patterns for bulk listing, filtering by type/age/size, deletion patterns, system-managed artifact limitations, and storage accounting. The outline maps to this list.
- WB-32858 (bulk run deletion with `deleteArtifacts: true` returning HTTP 500 because the request was too large to run synchronously) and a July 2026 SaaS incident where two customer bulk-deletion jobs saturated the metadata database. **Neither is referenced in the page.** They are the basis for the user-facing guidance to delete sequentially, keep batches to a few hundred, and expect oversized requests to time out.

### Decisions

- **How-to, not concept.** Deletion mechanics and GC already have a home in `delete-artifacts.mdx`. The new page links there rather than duplicating it.
- **Scope selection is the spine of the page.** The strongest evidence in the sources is that per-version deletion is the wrong tool at this volume, so the page ranks scopes before showing any code.
- **Two-pass select-then-delete** in the version section, rather than deleting inside the pagination loop. It gives the reader a reviewable list, a resumable log, and avoids mutating a result set while paging it.
- **Bracket placeholders** (`[ENTITY]`, `[PROJECT]`) per `AGENTS.md`, even though neighboring artifact pages still use `<entity>`.
- **Left out deliberately:** GraphQL mutations (not public API surface), direct bucket key deletion (already covered in `managing-bucket-storage.mdx` with the right cautions for operators), and `wandb artifact cache cleanup` (local cache, not server storage — covered in `models/artifacts/storage.mdx`).
- **No incident or bug references.** Internal failure modes appear only as neutral pacing advice.

### Needs SME verification

- [ ] **Team default TTL at scale.** The page states that setting a team default TTL expires an existing backlog by creation date (from `ttl.mdx`). Please confirm this reliably processes a large backlog, and whether there's practical latency worth documenting.
- [ ] **Batch size for run deletion.** "A few hundred at a time" is inferred from the HTTP 500 report on oversized bulk run deletion, not from a documented limit. Is there a recommended maximum?
- [ ] **`time.sleep(0.1)` as the pacing default.** Carried over from the existing pagination support article. Is that the right starting delay for deletion mutations specifically, which are heavier than reads?
- [ ] **Manage storage page coverage.** The page tells readers to find their largest consumers there. A 2025 ticket (WB-23942) reported artifacts not appearing on that page. Is that fixed? If artifacts can still be missing, the page should say so.
- [ ] **UI failure wording.** "The W&B App returns an error when you select many versions at once" is generalized from a single user report of a portal error at very high selection counts. Please confirm the symptom and, if there's a known practical selection ceiling, we can state it.
- [ ] **`ArtifactCollection.delete()` and protected aliases.** The page implies collection deletion is unrestricted. `delete-artifacts.mdx` says registry admins can delete collections containing protected aliases — does a non-admin's `collection.delete()` fail if any version carries a protected alias?
- [ ] **Inventory script cost.** Iterating every version of every collection to sum `size` may be slow or rate-limit-prone on very large projects. Should the page warn about this, or recommend a lighter approach?
- [ ] **Example output realism.** The illustrative `text` block uses invented collection names, counts, and sizes. Please confirm the shape matches what the script actually prints.

anastasiaguspan:
## Resume prompt

Use this to pick up the work in a new session:

> Continue work on wandb/docs PR https://github.com/wandb/docs/pull/3017, branch `aguspan/docs-bulk-delete-artifacts`, which adds `models/artifacts/bulk-delete-artifacts.mdx` ("Delete artifacts in bulk") plus nav and cross-reference updates in `docs.json`, `models/artifacts/delete-artifacts.mdx`, `platform/app/settings-page/storage.mdx`, and `platform/hosting/managing-bucket-storage.mdx`.
>
> Context: the page was authored with the `author-docs` skill in autonomous mode (no plan-mode approval step) and style-reviewed with the `style-guide` skill applied inline (passes run in-context rather than via sub-agents, because the user asked for no Agent tool usage in that session).
>
> Key decisions, do not silently reverse them:
> - The page's spine is scope selection: delete a project, then a collection, then runs with artifacts, then individual versions. This came from a support case (Slack `#billing-support-wandb`, Dec 2025 to Jan 2026) where a user with 5.2 TB and roughly 2M artifacts was resolved by deleting the project after backing up, not by a per-version loop.
> - Deletion mechanics and garbage collection stay in `models/artifacts/delete-artifacts.mdx`. The new page links there instead of restating them.
> - The version-deletion example is intentionally two-pass (select to a file, then delete from that file with a resumable log).
> - Internal failure modes (WB-32858 HTTP 500 on oversized bulk run deletion, and a July 2026 metadata-DB saturation incident) inform the pacing advice but are never referenced in the page.
> - Placeholders use the `[ENTITY]` bracket style per `AGENTS.md`, not the `<entity>` style used by older artifact pages.
> - English only. Do not edit `ja/`, `ko/`, or `fr/`.
>
> Remaining work: resolve the "Needs SME verification" checklist in the PR's sources-and-decision-log comment. The two most likely to change the text are (1) whether a team default TTL reliably expires a large existing backlog and (2) whether a recommended maximum batch size exists for run deletion with artifacts. Also confirm the `Manage storage` page reliably lists artifact-heavy projects (WB-23942 reported artifacts missing from it in 2025); if not, add a caveat.
>
> No JIRA ticket was filed for this work. The closest related tickets are WB-37221 (Inefficiencies querying artifacts, Done) and DOCS-2932 (Artifact soft delete, hard delete, restore, Icebox).

anastasiaguspan:
## Skill metadata

```json
{
  "skill": "author-docs",
  "version": "1.1.0",
  "depends_on": {
    "style-guide": "1.0.0"
  },
  "skills_invoked": ["style-guide"],
  "model": "claude-opus-5[1m]",
  "mode": "autonomous",
  "note": "style-guide passes 1-8 were applied inline in the main context rather than via sub-agents, because the session prohibited Agent tool usage. Notion MCP was unavailable; Notion was searched through Glean's app filter instead, and Glean read_document/chat permissions were not granted, so JIRA and Slack evidence came from search snippets only."
}
```

PR #3027, opened 2026-08-05
test: Pin guard smoke test (do not merge)

dfinster:
Throwaway. Confirms the DocEngine pin guard blocks a non-release pin. Close without merging.


dfinster:
@mdlinville Smoke test works! 🚀 
https://github.com/wandb/docs/actions/runs/31043515005/job/92433344791?pr=3027

PR #3029, opened 2026-08-05
Generate Weave Service API reference pages as stubs

mdlinville:
Completes the migration begun in #3019, applying the same mechanism to the Weave Service API: **115 stub pages across 23 nav groups**, with the `openapi` block removed from the Service API nav group. Also wires the generator into the Service API sync workflow — deliberately held back from #3019 so the monthly sync couldn't perform this cutover unprompted.

## This is the change that actually fixes the broken links

All eight internal links to OpenAPI operation pages target **this** spec. None targeted serverless-training, which is why the pilot fixed none:

```
/weave/reference/service-api/agents/export-genai-trace          (EN)
/weave/reference/service-api/eval-results/eval-results-query     (EN)
/{fr,ja,ko}/weave/reference/service-api/agents/export-genai-trace
/{fr,ja,ko}/weave/reference/service-api/eval-results/eval-results-query
```

They break because Locadex correctly locale-prefixes them while **the destination doesn't exist at the English slug** — the French page for `POST /agents/otel/v1/traces` lives at `.../agents/exporter-la-trace-genai`.

**Fixing the links by hand does not hold.** `8f09c18a2` stripped the prefixes on #2972 and CI went green; Locadex then regenerated those files in `610a68143`, and the identical 6 links broke again. That's now demonstrated, not predicted.

Stubs fix it at the **destination** rather than the link. Once Locadex mirrors these stubs into `fr/ja/ko` under identical filenames, `/fr/weave/reference/service-api/agents/export-genai-trace` resolves — so Locadex's prefixing becomes *correct* instead of broken, with no manual reapplication after each refresh.

## What the pilot already proved

Verified on the current Locadex branch after #3019 merged, so this is measured rather than assumed:

- **12 locale stubs per locale**, filenames preserved (`create-sft-training-job.mdx` in all four).
- **Titles translated** — `"Créer une tâche d'entraînement SFT"`, `"SFT トレーニングジョブを作成"`.
- **`openapi:` rewritten per locale** — `/fr/serverless-training/api-reference/openapi.json`. This was the open question for General Translation; `experimentalLocalizeStaticUrls` handles it.
- **Nav mirrored** — `openapi` block removed from all three locale blocks, 4 groups × 12 pages each, group labels translated (`chat-complétion`, `チャット補完`, `채팅-완성`), paths locale-prefixed.
- Locadex emits the `openapi:` value **line-folded** across two lines. Mintlify accepts it — `mint validate` reported "No errors or warnings detected."

## Verification

- **No English URL moves.** Slug parity reproduces all **115** published operation URLs exactly.
- The **5 `x-exclude`** operations keep their stub files (they're published) but stay out of nav. The `Inference` group correctly disappears, since all 5 of its operations are excluded — matching current behavior.
- Generator is **idempotent** — second run reports `create 0, rebind 0, prune 0`.
- `mint validate` and `mint broken-links` both pass at CI version **4.2.771**.
- `docs.json` diff is confined to the Service API group; the round-trip is byte-identical elsewhere.

## After merge

The next Locadex run should mirror all 115 stubs into `fr/ja/ko` (~345 files, one-time) and the 6 broken links should resolve on their own. Worth confirming on that run rather than pre-emptively editing the links again.

### DIFF
--- a/.github/workflows/update-service-api.yml
+++ b/.github/workflows/update-service-api.yml
@@ -58,7 +58,19 @@ jobs:
       run: |
         echo "Converting Service API links to fully qualified URLs..."
         python scripts/reference-generation/weave/fix_service_api_links.py
-        
+
+    # Runs unconditionally, not just when the spec changed: the generator is idempotent,
+    # so it is a no-op in the common case, and running it always lets it heal drift (for
+    # example a stub deleted by hand) rather than letting nav and pages diverge silently.
+    - name: Generate API reference pages and navigation
+      id: stubs
+      run: |
+        echo "Generating stub pages from the Weave Service API OpenAPI spec..."
+        OUTPUT=$(python scripts/reference-generation/common/generate_openapi_stubs.py --spec weave 2>&1)
+        echo "$OUTPUT"
+        SUMMARY=$(echo "$OUTPUT" | grep '^STUBS_SUMMARY ' | head -1 | cut -d' ' -f2-)
+        echo "summary=${SUMMARY:-unavailable}" >> $GITHUB_OUTPUT
+
     - name: Check for changes
       id: check-changes
       run: |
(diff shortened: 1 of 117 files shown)


mdlinville:
Added @dbrian57 as FYI, but this is low-risk, using the same approach we used for the smaller reference and all of the tests pass, so I'll merge myself.

PR #3051, opened 2026-08-10
ci: Migrate the DocEngine build workflow to the published action (rc harness)

dfinster:
Test harness for the docengine-build action per the docs CI plan's rc-tag loop: the pin moves to an rc tag, so **PinGuard is red by design and this PR can never merge**. It exists to run the new thin shell against the real repo, including the issue-report failure path. The mergeable version of the shell change follows the v3.7.0 release on a clean branch; this draft is then closed unmerged.


dfinster:
Superseded by the mergeable PR; rc harness closed unmerged per the rc-tag loop.

PR #3059, opened 2026-08-11
Add TypeScript tabs to OpenAI Realtime API integration guide

dbrian57:
## Description

Resolves DOCS-2551. Adds TypeScript coverage to the Weave OpenAI Realtime API doc.

### DIFF
--- a/weave/guides/integrations/openai-realtime-audio.mdx
+++ b/weave/guides/integrations/openai-realtime-audio.mdx
@@ -320,6 +320,7 @@ To run the example, complete the following steps:
     npm install weave @openai/agents zod
     npm install --save-dev typescript @types/node
     npx tsc --init
+    npm pkg set type=module
     ```
 
 1. Create a file titled `weave_realtime_agent.ts` and add the following code to it.
@@ -409,7 +410,7 @@ To run the example, complete the following steps:
 
     ```bash
     npx tsc
-    node weave_realtime_agent.js
+node --import=weave/instrument weave_realtime_agent.js
     ```
 
 The agent connects to the Realtime API, asks about the weather in San Francisco, calls the weather tool, and prints the assistant's response before disconnecting. In the Weave UI, you can explore the resulting trace, including the session, the generation, the tool call, and the assistant's audio output.
--- a/weave/guides/integrations/openai-realtime-audio.mdx
+++ b/weave/guides/integrations/openai-realtime-audio.mdx
@@ -1,32 +1,60 @@
 ---
 title: OpenAI Realtime API
 description: Use Weave to automatically trace your calls to the OpenAI Realtime API.
-keywords: ["speech-to-speech", "voice assistant", "WebSockets", "Agents SDK", "audio tracing"]
+keywords: ["speech-to-speech", "voice assistant", "WebSockets", "Agents SDK", "audio tracing", "TypeScript"]
 ---
 
 Weave's integration with [OpenAI's Realtime API](https://developers.openai.com/api/docs/guides/realtime/) lets you automatically trace your application's speech-to-speech interactions as they occur. You can use this to capture conversations between your agents and users to review and evaluate agent performance.
 
-This guide shows you how to enable Realtime API tracing in your application and walks through two end-to-end voice assistant examples that you can run locally.
+This guide shows you how to enable Realtime API tracing in your application and walks through end-to-end examples in Python and TypeScript that you can run locally.
 
 ## Integrate realtime traces
 
 This section shows the minimum code needed to enable Weave tracing for the OpenAI Realtime API in any application.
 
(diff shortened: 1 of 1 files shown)


weave/guides/integrations/openai-realtime-audio.mdx
```diff
@@ -282,26 +306,133 @@ Once the assistant is running, press `T` on your keyboard to mute or unmute the
...
+
+To run the example, complete the following steps:
+
+1. In a new directory, set up a Node.js project with TypeScript and install the following libraries:
+
+    ```bash
+    npm init -y
+    npm install weave @openai/agents zod
+    npm install --save-dev typescript @types/node
+    npx tsc --init
```
dbrian57, line 322: ❌ **Refuted** — reproduced locally: these steps fail to compile with current TypeScript. `npx tsc --init` (TypeScript 5.9+/7.x) now generates `"module": "nodenext"` + `"verbatimModuleSyntax": true`, and `npm init -y` creates a CommonJS package, so `npx tsc` fails with **TS1295** ("ECMAScript imports and exports cannot be written in a CommonJS file") on the example's first line.

```suggestion
    npx tsc --init
    npm pkg set type=module
```

**Source says:** verified locally with TypeScript 7.0.2 / Node 22.23.1 — as written, `npx tsc` exits 1 with TS1295; after `npm pkg set type=module`, it compiles clean. Note this makes the app ESM, which also requires the run-command change flagged at line 412. · **Source:** local reproduction of the doc's exact steps; TS config semantics per [tsconfig verbatimModuleSyntax](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax)

_🤖 CoreWeave Docs Team's Validator — refuted_

weave/guides/integrations/openai-realtime-audio.mdx
```diff
@@ -282,26 +306,133 @@ Once the assistant is running, press `T` on your keyboard to mute or unmute the
...
+    ```
+    </Accordion>
+
+1. Update the `WEAVE_PROJECT` value with your team and project names.
+1. Set your `OPENAI_API_KEY` and `WANDB_API_KEY` environment variables.
+1. Compile and run the agent:
+
+    ```bash
+    npx tsc
+    node weave_realtime_agent.js
```
dbrian57: ❌ **Refuted** — with the ESM fix from line 322 (`"type": "module"`), this command runs but **produces no traces**: in a plain ESM app, importing Weave does not register the module loader hooks in time, so `RealtimeSession` is never instrumented (verified with a runtime probe: no Weave adapter attached to the session). Weave's own ESM test fixture launches with `node --import=weave/instrument` — "same invocation pattern real ESM consumers use."

```suggestion
node --import=weave/instrument weave_realtime_agent.js
```

**Source says:** probe with `--import=weave/instrument` → adapter attached ✅; plain `node` → not attached ❌. The `weave` package exports `./instrument` for exactly this. Also note the commented fallback `weave.patchRealtimeSession()` (line 366) throws `TypeError: Cannot redefine property: RealtimeSession` in plain-ESM apps — worth raising with Chance An. · **Source:** [weave:esm-basic/main.mjs#L3-L5](https://github.com/wandb/weave/blob/c0f6aeef0af8ae14ad45b7f4772f61511a1deb3f/sdks/node/src/__tests__/hostApps/fixtures/esm-basic/main.mjs#L3-L5)

_🤖 CoreWeave Docs Team's Validator — refuted_

weave/guides/integrations/openai-realtime-audio.mdx
```diff
@@ -282,26 +306,133 @@ Once the assistant is running, press `T` on your keyboard to mute or unmute the
...
+    ```
+
+The agent connects to the Realtime API, asks about the weather in San Francisco, calls the weather tool, and prints the assistant's response before disconnecting. In the Weave UI, you can explore the resulting trace, including the session, the generation, the tool call, and the assistant's audio output.
```
dbrian57: 🟡 **Partially confirmed** — the code and this claim are verbatim from the SDK team's own example ([weave:openaiAgentsRealtime.ts](https://github.com/wandb/weave/blob/c0f6aeef0af8ae14ad45b7f4772f61511a1deb3f/sdks/node/examples/openaiAgentsRealtime.ts#L4-L5)), and `output_text` is a real content type in `@openai/agents-realtime`; **whether the assistant's reply actually arrives as `output_text` (and prints) in a default voice session is unverified** — it may arrive as `output_audio` only, in which case nothing prints.

**Recommended:** run the example once with real `OPENAI_API_KEY`/`WANDB_API_KEY` and confirm text is printed, or soften to "the response appears in the Weave trace."

_🤖 CoreWeave Docs Team's Validator — partially confirmed_

thedriftofwords (approved):

PR #3062, opened 2026-08-12
ci: Install the dormant DocEngine import shell (rc harness)

dfinster:
Test harness for the dormant import shell per the docs CI plan's rc-tag loop: the pin moves to an rc tag, so **PinGuard is red by design and this PR can never merge**. It exists to prove the no-op dispatch against empty importers. The mergeable version follows the v3.8.0 release on a clean branch; this draft is then closed unmerged.


dfinster:
Closing release candidate test harness. This will be deployed on https://github.com/wandb/docs/pull/3063

PR #3066, opened 2026-08-13
ci: Install the dormant DocEngine overlay shell (rc harness)

dfinster:
Test harness for the dormant overlay shell per the docs CI plan's rc-tag loop: the pin moves to an rc tag, so **PinGuard is red by design and this PR can never merge**. It exists to prove the no-op dispatch against empty overlays. The mergeable version follows the v3.9.0 release on a clean branch; this draft is then closed unmerged.


dfinster:
Testing complete, closing harness

PR #2673, opened 2026-05-27
style: apply /style-guide pass to models/integrations

johndmulhausen:
## Summary

This PR applies the `/style-guide` skill (Google Developer Style Guide + CoreWeave conventions) to 52 files under `models/integrations`. The run was automated; only style, terminology, voice, and formatting were changed — no technical content was added or modified.

## Files edited

- `models/integrations/accelerate.mdx`
- `models/integrations/add-wandb-to-any-library.mdx`
- `models/integrations/autotrain.mdx`
- `models/integrations/azure-openai-fine-tuning.mdx`
- `models/integrations/catalyst.mdx`
- `models/integrations/cohere-fine-tuning.mdx`
- `models/integrations/composer.mdx`
- `models/integrations/dagster.mdx`
- `models/integrations/databricks.mdx`
- `models/integrations/deepchecks.mdx`
- `models/integrations/deepchem.mdx`
- `models/integrations/diffusers.mdx`
- `models/integrations/docker.mdx`
- `models/integrations/dspy.mdx`
- `models/integrations/farama-gymnasium.mdx`
- `models/integrations/fastai.mdx`
- `models/integrations/fastai/v1.mdx`
- `models/integrations/huggingface.mdx`
- `models/integrations/huggingface_transformers.mdx`
- `models/integrations/hydra.mdx`
- `models/integrations/ignite.mdx`
- `models/integrations/keras.mdx`
- `models/integrations/kubeflow-pipelines-kfp.mdx`
- `models/integrations/lightgbm.mdx`
- `models/integrations/lightning.mdx`
- `models/integrations/metaflow.mdx`
- `models/integrations/mmengine.mdx`
- `models/integrations/mmf.mdx`
- `models/integrations/nim.mdx`
- `models/integrations/openai-api.mdx`
- `models/integrations/openai-fine-tuning.mdx`
- `models/integrations/openai-gym.mdx`
- `models/integrations/paddledetection.mdx`
- `models/integrations/paddleocr.mdx`
- `models/integrations/prodigy.mdx`
- `models/integrations/pytorch-geometric.mdx`
- `models/integrations/pytorch.mdx`
- `models/integrations/ray-tune.mdx`
- `models/integrations/sagemaker.mdx`
- `models/integrations/scikit.mdx`
- `models/integrations/simpletransformers.mdx`
- `models/integrations/skorch.mdx`
- `models/integrations/spacy.mdx`
- `models/integrations/stable-baselines-3.mdx`
[...]

PR #3072, opened 2026-08-15
fix: use a stable seed in the PyTorch integration

marcus-campbell:
## Description

The PyTorch integration page says this setup ensures deterministic behavior, but it doesn't appear to actually do that. This is because it derives its seeds from `hash("...")`. By default, Python randomizes string hashes _between_ interpreter processes, so the example can initialize its RNGs differently across runs.

This fix replaces the hash-derived values with one named integer seed.

Before:

```python
torch.backends.cudnn.deterministic = True
random.seed(hash("setting random seeds") % 2**32 - 1)
np.random.seed(hash("improves reproducibility") % 2**32 - 1)
torch.manual_seed(hash("by removing stochasticity") % 2**32 - 1)
torch.cuda.manual_seed_all(hash("so runs are repeatable") % 2**32 - 1)
```
After:

```python
seed = 42
torch.backends.cudnn.deterministic = True
random.seed(seed)
np.random.seed(seed)
torch.manual_seed(seed)
torch.cuda.manual_seed_all(seed)
```

### Why this matters

A small exploratory search found the same recognizable recipe in at least 24 other repositories. This suggests that the pattern has propagated beyond W&B's own docs, so correcting the bug here can help prevent it from spreading further.

The problem was also noted in the technical-review notes for [#2673](https://github.com/wandb/docs/pull/2673), but it wasn't fixed as it was out-of-scope for that PR (it was a style guide pass).

The corresponding notebooks are updated in [wandb/examples#644](https://github.com/wandb/examples/pull/644).

## Testing

- [x] `git diff --check` succeeds.
- [x] Confirmed that the diff contains only the intended seed changes.
- [x] Confirmed that the old hash-derived seeds vary with `PYTHONHASHSEED`.
- [x] Confirmed that the replacement has no interpreter hash-secret dependency.
- [ ] PR tests succeed.

### DIFF
--- a/models/integrations/pytorch.mdx
+++ b/models/integrations/pytorch.mdx
@@ -70,11 +70,12 @@ import torchvision.transforms as transforms
 from tqdm.auto import tqdm
 
 # Ensure deterministic behavior
+seed = 42
 torch.backends.cudnn.deterministic = True
-random.seed(hash("setting random seeds") % 2**32 - 1)
-np.random.seed(hash("improves reproducibility") % 2**32 - 1)
-torch.manual_seed(hash("by removing stochasticity") % 2**32 - 1)
-torch.cuda.manual_seed_all(hash("so runs are repeatable") % 2**32 - 1)
+random.seed(seed)
+np.random.seed(seed)
+torch.manual_seed(seed)
+torch.cuda.manual_seed_all(seed)
 
 # Device configuration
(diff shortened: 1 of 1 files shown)


mdlinville (approved):
Thank you!

mdlinville (approved):
Thank you

PR #3094, opened 2026-08-21
Workspaces: Clarify media panel image options

ngrayluna:
## Description

There was internal confusion about media panel options. While I was here I updated the layout of the doc to both update the latest settings and organize it based on the UI layout.

## Related issues

- Fixes https://coreweave.atlassian.net/browse/DOCS-3104

### DIFF
--- a/models/app/features/panels/media.mdx
+++ b/models/app/features/panels/media.mdx
@@ -30,9 +30,120 @@ To add a media panel manually so that you can configure it during creation:
 
 ## Configure a media panel
 
-Panels for all media types have the same options, so the following steps apply regardless of which media type you're working with.
+When you add a media panel manually, its configuration page opens after you select the media type.
 
-When you add a media panel manually, its configuration page opens after you select the type of media. To update the configuration for an existing panel, hover over the panel, then click the gear icon that appears at the top right. The following sections describe the settings available in each tab.
+To configure an existing panel, hover over the panel and select the gear icon (<Icon icon="gear" iconType="solid"/>) in the upper-right corner. The following sections describe the settings available on each tab.
+
+### Display
+
+Configure the panel's appearance and behavior.
+
+#### Viewer Settings
+
+- **Panel title**: Specify the panel title.
+- **Panel mode**: Specify the layout of the media panel. Choose from the following modes:
+    - **Gallery** (default): Specify the number of columns and whether the grid displays one item per run, step, or index. By default, the grid displays one item per run.
+    - **Grid**: Specify the number of columns, the x-axis, and the y-axis. By default, the x-axis is **Log Step** and the y-axis is **Run**.
+    - **Compare**: Compare up to four media items side by side. Optionally, fan out items by step or index. For more information, see [Compare mode](#compare-mode).
+- **Media key(s)**: Logged W&B fields that contain the images, videos, or other media the panel should display.
+- **Slider key**: Specify the key to use for the panel's step slider. The default is **Step**.
+
+#### Appearance
+
+Configure how the panel displays media:
+
+- **Smooth image**: Smooth images to improve their appearance.
+
+#### Navigation & Playback
+
+Control how you navigate through media logged at different points in a run:
+
+- **Snap to step**: Toggle on or off. If on, the stepper advances to the next existing step after *Stride length*. Otherwise, **Snap to step** advances by **Stride length** even if it doesn't align with an existing step.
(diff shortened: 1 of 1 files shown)


models/app/features/panels/media.mdx
```diff
@@ -30,68 +30,129 @@ To add a media panel manually so that you can configure it during creation:
...
 
-- **Sync slider by key**: Configure whether the step sliders for videos in the section move in sync.
-- **Autoplay videos**: Configure whether videos start playing when the page loads.
-- **Loop videos**: Configure whether videos in the section restart automatically and play continuously until stopped. Not customizable at the section level. Appears only if the workspace has video media panels.
+Configure how the panel displays media:
 
-### Display
+- **Smooth image**: Smooth images to improve their appearance.
+
+#### Navigation & Playback
```
ngrayluna, line 59: For this and following section titles (that have an ampersand), I wasn't sure if I should:

a) Keep the ampersand
b) Use sentence casing

Re: Those are the actual section names in the UI.

anastasiaguspan: valid question - it certainly isn't our style to use the ampersands yet it is the explicit UI : hmmm, maybe change the sections to be "Navigation and Playback" and augment the introductory line of the list to include the explicit capture: "Control how you navigate through media logged at different points in a run through `Navigation & Playback`:

models/app/features/panels/media.mdx
```diff
@@ -30,68 +30,129 @@ To add a media panel manually so that you can configure it during creation:
...
+
+- **Autoplay videos**: Start playing videos automatically when you open the workspace or panel.
+- **Loop videos**: Restart videos automatically after playback ends.
 
 ### Layout
 
-Customize the display of the panel's individual items:
+Configure how the panel arranges and displays individual media items.
+
+#### Grid & Organization
```
anastasiaguspan, line 83: Could do the same thing here.  title "Grid and Organization" and then add a lead-in to the bullets of "Control X using `Grid & Organization`:"
(since we're supposed to have intro sentences for lists anyway i think)

models/app/features/panels/media.mdx
anastasiaguspan, line 26: took me a minute to find this location (vs all the other buttons controls). Maybe:
**Global**:  In the **Charts** search bar, click **Add panels**.

anastasiaguspan, line 24: This is under Workspace -> Charts right? nothing here says 'where' these instructions start

models/app/features/panels/media.mdx
```diff
@@ -30,68 +30,129 @@ To add a media panel manually so that you can configure it during creation:
...
-        - **Original size**: Load full-resolution images in their original dimensions.
+- **Thumbnails** (default): Load smaller previews for faster loading. W&B scales thumbnails to fit the panel.
+- **Original image**: Load the full-resolution image file. W&B scales the image to fit the panel.
 
-        On Multi-tenant Cloud, W&B always loads full-resolution images in the full-screen view, regardless of this setting.
-    - **Videos**: In **Gallery** mode, turn on **Use original size** to display videos at their original dimensions. If off (default), W&B scales videos to fill as much panel space as possible while maintaining the image's aspect ratio. In **Grid** mode, each video fills the grid width.
-- **Right-handed system**: If on, point clouds use the right-handed system for plotting points, rather than the default left-handed system.
+For W&B Dedicated Cloud and W&B Self-Managed:
+
+- **Default** (default): Use the default image dimensions for the panel mode.
```
anastasiaguspan: i get that this is consistent, but do we really need to (should we really) label a named default as a (default), or is this implied enough?  could prb drop the (default), but i also see the argument to leave it in.

anastasiaguspan (approved):
since you said your goal was clarifying confusion, i left some comments on the intro too (tho outside your edited scope).

anyway, much improved!

models/app/features/panels/media.mdx
ngrayluna: Interesting. Thanks for calling this out. Admittedly, I am not as clear on this myself (didn't write this chunk). I created a second ticket to address this feedback (so as to not prevent merging this PR)

models/app/features/panels/media.mdx
ngrayluna: Good question. Unsure. That's we historically have done. But it could be updated. For consistency I'll leave as is for now but it might be worth adding to the list of style guide questions?

models/app/features/panels/media.mdx
ngrayluna: And here is said ticket: https://coreweave.atlassian.net/browse/DOCS-3111

PR #3107, opened 2026-08-25
Workspaces: Updates UI

ngrayluna:
## Description
The UI was changed from “Quick add” was changed to “Quick panel builder”. Also, the workflow changed, too.

## Related issues

- Fixes https://coreweave.atlassian.net/browse/DOCS-3114


anastasiaguspan (approved):
:)

PR #3118, opened 2026-08-25
Comment out ARIA project memories, which are not yet available

mdlinville:
## Description
Comment out ARIA project memories, which are not yet available

## Testing
- [ ] PR tests succeed

### DIFF
--- a/aria/overview.mdx
+++ b/aria/overview.mdx
@@ -52,7 +52,7 @@ ARIA remembers what it learns. As you work, ARIA writes _memories_ scoped to you
 
 Each memory has a name and description so ARIA knows when to recall it, plus the memory text itself.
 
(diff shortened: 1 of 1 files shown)


aria/overview.mdx
clocasto-wandb: As implemented, the restriction is "any member of the team that owns the project". The idea is that external users shouldn't be able to read or share memories generated on public projects, for example (though the teams that own the public projects can still benefit).

clocasto-wandb, line 58: I'm unfamiliar with the doc rendering tool being use, so maybe this is fine, but should these numbers continue (e.g. 2, 3, 4 ...)?

mdlinville:
```suggestion
Anyone in the team that owns the project can view, edit, or delete memories:
```

PR #3169, opened 2026-09-01
Artifacts: digest algorithm + info on verify considerations 

ngrayluna:
## Description

- Mentions new digest algorithm.
- Adds info about when and why to use `wanb.Artifact.verify()`


models/artifacts/construct-an-artifact.mdx
```diff
@@ -22,41 +22,55 @@ Initialize the [`wandb.Artifact()`](/models/ref/python/experiments/artifact) cla
...
 import wandb
-
 # Create an artifact object
-artifact = wandb.Artifact(name="<name>", type="<type>")
+artifact = wandb.Artifact(
+    name="my-dataset",
+    type="dataset",
+    description="This is my dataset artifact",
+    metadata={"source": "generated", "version": 1},
+    digest_algorithm="XXH128"
```
trane293, line 54: this is the main caveat from the rollout plan, so i think we should put it directly next to the opt-in example. MD5 is still the default, existing MD5 sequences stay on MD5, and SDKs before 0.29.0 can't verify XXH128 entries. something like:

> MD5 remains the default. With W&B Python SDK 0.29.0 or later, set `digest_algorithm="XXH128"` to opt in to faster hashing. Existing artifact sequences that use MD5 continue to use MD5. SDK versions before 0.29.0 cannot verify artifacts containing XXH128 entries with `Artifact.verify()`.

models/artifacts/download-and-use-an-artifact.mdx
```diff
@@ -97,32 +97,58 @@ Alternatively, you can download files from a certain directory. To do so, specif
...
+```python
+import wandb
+
+with wandb.init(project="project-name", job_type="job-type") as run:
+    artifact = run.use_artifact("entity/project/my-dataset:latest")
+    artifact_dir = artifact.download()
+    artifact.verify(artifact_dir)
+```
+
+[`wandb.Artifact.verify()`](/models/ref/python/experiments/artifact#method-artifact-verify) recomputes checksums for local files and compares them with the checksums in the artifact manifest. It raises a `ValueError` if a file is missing, modified, or if the directory contains files that are not included in the artifact.
```
trane293: small factual correction here: a missing file currently raises `FileNotFoundError`, while modified or unexpected files raise `ValueError`. easiest fix is to avoid naming the exception:

> Verification fails if a file is missing or modified, or if the directory contains files that are not included in the artifact.

trane293 (commented):
looks good overall, i left two small factual comments. both should be quick to address and then we're good to ship.

mdlinville (approved):

PR #3223, opened 2026-09-14
docs: Note Dedicated Cloud availability for Serverless Sandboxes (DOCS-3205)

mdlinville:
## Freeze exception

`wandb/docs` is in a soft freeze. This is an agreed exception: the content publishes **Thursday 2026-09-17**, alongside the Dedicated Cloud fleet upgrade to Server 0.85.0, and it has to land here first so the Forge import can carry it down.

DOCS-3205. Blocked-by epic WB-38844 (Sandboxes on Dedicated — identity work) is Done.

## The change

One line in `snippets/_includes/sandboxes-public-preview.mdx`, which is included by **8 pages**:

`sandboxes.mdx` · `sandboxes/create-sandbox` · `sandboxes/file-access` · `sandboxes/invoke-agent-sandbox-tutorial` · `sandboxes/lifecycle` · `sandboxes/mltrain-in-sandbox-tutorial` · `sandboxes/run-commands` · `sandboxes/secrets`

**Before**
> Serverless Sandboxes is in **public preview**. Access is enabled per organization. To request access for your organization, contact support.

**After**
> Serverless Sandboxes is in **public preview**, available in **Multi-tenant Cloud** and in **Dedicated Cloud** v0.85.0+. Access is enabled per organization. To request access for your organization, contact support.

The notice previously named no deployment type at all, so all eight pages silently implied sandboxes work everywhere. A search across `sandboxes/` found no other availability claim to update.

## Why v0.85.0+

Dedicated Cloud reaches sandboxes with the fleet upgrade on Thursday. Orgs pinned to older or custom builds are the known exception.

That floor also matches the code. In `wandb/core`, `services/gorilla/api/handler/sandboxes_billing.go` is new in `server-release-0.85.x` (absent from 0.84.x), and `NewRouteMiddlewareWithSandboxesAuth` gains its `onprem bool` parameter in the same branch. The `RampSandboxes` constant predates both, but the onprem-serving surface does not.

## Why Self-Managed is not listed

Not an omission. `RampSandboxes` in `services/gorilla/pkg/ramping/ramp_settings_statsig_gates.go` is:

```go
RampSandboxes: {
    Description: "Whether to enable sandboxes",
    Type:        RampTypeStatsig,
    IDType:      RampIDTypeOrgID,
},
```

No `Default` (so `false`) and no `Overridable`. Onprem has no Statsig key and falls back to `Default`, and because the gate is not `Overridable`, a Self-Managed operator cannot enable it with an env var — setting one on a non-`Overridable` gate panics at startup.

## Why "public preview" is unchanged

The label is loose — the product is preview-by-request rather than self-serve — but the same looseness exists elsewhere in the docs, and it's being resolved after 9/30 rather than piecemeal here. The second sentence, "Access is enabled per organization," is accurate and carries the real mechanism: the gate is `RampIDTypeOrgID` and is flipped per org.

## Support address

This keeps `support@wandb.com` deliberately. The CoreWeave copy of this snippet uses the Forge address, which is not public knowledge until 9/30. The Forge import 3-way merges that local edit back in downstream.

## Validation

`scripts/mdx-validation/validate-mdx-mintlify.sh` — passed, no errors, no broken links.

## Translations

English only; locadex regenerates `fr/`, `ja/`, `ko/` on merge.

## Unrelated bug found while verifying

Two comments in `core` state the opposite of their implementation. `sandboxes_billing.go:110` says "On-prem always returns false" and `sandboxes.go:765` says "on-prem always fails (402)", but `hasAvailableSandboxesUsage` returns `true, nil` when `s.onprem`. Actual behavior is no usage or billing enforcement on Dedicated sandboxes. Filing separately against `core`; noted here because anyone documenting from those comments would state the inverse.


mdlinville:
Tagged @cassface who may be able to help resolve the billing question.

mdlinville:
This isn't yet ready for the public so I'll close this PR for now.

